Issue #15, opened 2026-03-31
Audit evidence-level tagging coverage and qualify absence-of-evidence claims


ELares:
Derived from discussion #9.

## Problem
The manuscript already notes that evidence-level tagging is incomplete, but discussion #9 reinforced that many no-clinical-evidence claims may still be partly artifacts of only partial classification coverage.

## Goal
Improve confidence in absence claims before using them in gap analyses or article framing.

## Acceptance criteria
- Measure current evidence-level tagging coverage precisely
- Sample untagged records to estimate miss rate
- Identify mechanisms where current claims are most likely to be overstated
- Update gap-analysis language so it consistently distinguishes true absence from non-detection

## References
- Discussion: #9
- Relevant files: `analysis/evidence-tiers.md`, `analysis/gap-analysis.md`, `article/drafts/v1.md`

Issue #14, opened 2026-03-31
Expand pathway tracking for ferroptosis resistance and adjacent cell-death programs


ELares:
Derived from discussion #9.

## Problem
The repo tracks core ferroptosis concepts well, but it under-tracks several recently emphasized pathways that could explain failure, patient stratification, or adjacent opportunities.

## Add tracking for
- `DHODH`
- `7-DHC` / `DHCR7`
- `MBOAT1` / `MBOAT2`
- `SCD`
- `FDX1`
- `TRIM25` / selective GPX4 degradation context
- cuproptosis markers
- disulfidptosis markers

## Why this matters
These pathways affect whether ferroptosis works in vivo, who may respond, and when a parallel death program might be more relevant than ferroptosis.

## Acceptance criteria
- Update keyword / enrichment strategy for these targets
- Assess how many corpus articles mention them
- Produce a short note on whether these targets should change repo priorities or manuscript framing

## References
- Discussion: #9
- Relevant files: `scripts/config.py`, `scripts/enrich_metadata.py`, `analysis/key-findings.md`

Issue #13, opened 2026-03-31
Revise the manuscript framing to add CAP and clarify the in vivo limits of ferroptosis translation


ELares:
Derived from discussion #9.

## Problem
The current article centers PDT and SDT, but the discussion surfaced two important framing changes:

1. `CAP` appears to belong in the same physical ROS class and is currently omitted
2. the article needs a sharper account of why ferroptosis success in vitro often fails in vivo

## Proposed manuscript changes
- Add `CAP` as a third physical ROS modality worth evaluating
- Reframe PDT/SDT/CAP as tools best matched to OXPHOS-dependent persisters, not as a general answer to resistance
- Add a dedicated section on in vivo ferroptosis escape / lipid remodeling limits
- Tighten the nanoparticle confound discussion

## Acceptance criteria
- Update manuscript outline or draft
- Add citations and language covering CAP, `SCD/MUFA`, and other escape axes
- Re-check whether the central claim still holds after these caveats are added

## References
- Discussion: #9
- Relevant files: `article/drafts/v1.md`, `analysis/distilled-hypotheses-final.md`

Issue #12, opened 2026-03-31
Extend the simulation to model in vivo lipid remodeling and ferroptosis escape


ELares:
Derived from discussion #9.

## Problem
The current simulation supports the manuscript thesis, but it does not appear to model one of the most important external caveats: the in vitro to in vivo ferroptosis gap driven by lipid remodeling, especially `SCD`-mediated MUFA enrichment.

## Goal
Add a model path that can represent ferroptosis suppression under 3D / in vivo-like conditions and test whether physical ROS modalities still overcome that resistance.

## Minimum scope
- represent `SCD` / MUFA-driven protection
- allow toggling in vivo-like lipid remodeling assumptions
- compare baseline ferroptosis sensitivity with and without this resistance axis
- document what conclusions remain valid only for in vitro-like assumptions

## Acceptance criteria
- Update simulation assumptions and parameterization
- Run sensitivity analysis with the new resistance mechanism
- Summarize whether prior conclusions materially change

## References
- Discussion: #9
- Relevant code: `simulations/ferroptosis-core/`, `simulations/sim-*`
- Key concepts mentioned in discussion: `SCD`, `MUFA`, in vivo lipid remodeling

PR #16, opened 2026-03-31
Add resistant-state and taxonomy scaffolding

ELares:
## Summary
- expand query and tagging scaffolding beyond the current physical-modality bias
- add resistant-state support and first-pass analysis artifacts
- update manuscript and simulation caveats for CAP and in vivo ferroptosis escape

## Verification
- `python -m py_compile scripts/config.py scripts/tag_articles.py scripts/fetch_articles.py scripts/build_index.py scripts/analyze_corpus.py`
- `cargo check -p ferroptosis-core`

## Issues
- Partial progress on #10
- Partial progress on #11
- Partial progress on #12
- Partial progress on #13
- Partial progress on #14
- Partial progress on #15


ELares (commented):
## Thorough Review: PR #16

All Python files pass `py_compile` and `cargo check` succeeds for ferroptosis-core. The scaffolding direction is correct and aligns well with issues #10–#15. I found one bug that should be fixed before merge, two significant design concerns worth discussing, and several minor items.

---

### Bug: Handwritten analysis files will be overwritten by generators

Both `analysis/resistant-state-map.md` and `analysis/evidence-coverage-audit.md` are committed as handwritten files with rich descriptive content (state definitions, interpretation risks, implementation notes). However, `analyze_corpus.py` registers generators for the **same filenames** (lines 561–562):

```python
("resistant-state-map.md", "Resistant-State Map", build_resistant_state_map),
("evidence-coverage-audit.md", "Evidence Coverage Audit", build_evidence_coverage_audit),
```

Running `python scripts/analyze_corpus.py` will silently destroy the handwritten content and replace it with auto-generated tables. The handwritten `resistant-state-map.md` has state definitions, implications, and an implementation note that don't exist in the generator output.

**Fix options:**
- Rename the handwritten files (e.g., `resistant-state-design.md` and `evidence-audit-notes.md`) so they coexist with the generated ones
- Or have the generators incorporate the handwritten prose as a preamble

---

### Significant: `therapy-induced-senescence` rule is too broad

The second keyword group includes `"therapy"` and `"treatment"` — nearly universal words in cancer literature:

```python
"therapy-induced-senescence": {
    "all_of": [
        ["senescence", "senescent", "therapy-induced senescence"],
        ["therapy", "drug resistance", "persister", "treatment"],
    ]
},
```

Any paper mentioning "senescence" AND "therapy" will match. That's essentially every cancer-senescence paper, not just those about therapy-induced senescence as a specific resistance mechanism. This will massively over-tag.

**Suggestion:** Tighten group 2 to terms that are genuinely specific to the resistance context — e.g., `["therapy-induced senescence", "senescence-associated secretory", "senescent tumor", "treatment-induced senescence"]` — or require a three-group rule adding a resistance/escape term.

The `stromal-sheltered-immune-excluded` rule has a similar (milder) issue: `"tumor microenvironment"` + `"therapy resistance"` is broad enough to tag a large fraction of TME papers that aren't specifically about stromal sheltering or immune exclusion.

---

### Significant: Issue #14 pathways only partially addressed

The PR adds `cuproptosis` and `disulfidptosis` to `BIOLOGY_PROCESS_KEYWORDS` but does **not** track the core ferroptosis resistance pathways issue #14 specifically asks for:

- DHODH
- 7-DHC / DHCR7
- MBOAT1 / MBOAT2
- SCD
- FDX1
- TRIM25 / N6F11

These are the verified resistance axes from the discussion thread. The PR description claims partial progress on #14, which is fair — but worth noting that the higher-priority targets (the ones that could change manuscript conclusions) are still missing.

---

### Minor issues

1. **Manuscript keywords line (v1.md:15) doesn't include "cold atmospheric plasma"** despite CAP being added to the abstract and intro. Should add it to maintain keyword/content alignment.

2. **SCD/MUFA logic is duplicated** in `biochem.rs` — the identical MUFA-protection block appears in both `sim_cell_step()` (used by spatial simulation) and `sim_cell()` (used by original Monte Carlo). Consider extracting to a helper to avoid divergence.

3. **Date-range asymmetry in queries.txt**: New mechanisms (CAP, radioligand, PROTAC, etc.) use `2020:2026[dp]` or `2022:2026[dp]`, while older queries go back to 2001+. This is pragmatic but means the corpus will have asymmetric time coverage for different mechanisms. Worth a one-line comment in queries.txt noting this is intentional.

4. **"cafs" (plural) won't match** the word-boundary regex for "caf" (3 chars). Most articles will still match via the longer "cancer-associated fibroblast" keyword, so the impact is low — but adding "cafs" as a separate keyword would close the gap.

---

### What's done well

- **Backward-compatible simulation**: `scd_mufa_rate` and `scd_mufa_max` default to 0.0, so existing results are bitwise-identical. Clean opt-in design.
- **Composite `all_of` matching**: The resistant-state architecture (requiring keywords from multiple groups) is the right pattern — much better than single-keyword OR matching for composite biological concepts.
- **Honest caveats throughout**: The manuscript limitations section, gap-analysis header, evidence-tiers caveat, and deep-pattern-analysis annotations all qualify claims appropriately. The evidence-coverage-audit framing ("non-detection layer, not definitive registry") is particularly well-phrased.
- **Warning when resistant_states is empty**: The `analyze_corpus.py` check prevents silent generation of meaningless tables.
- **Division-by-zero protection**: The `+ 1e-9` guard in the MUFA model and the `.max(0.05)` effective-unsaturation floor are correct.

---

### Verdict

The file-overwrite bug should be fixed before merge — it will cause silent content loss the first time anyone runs the analysis pipeline. The overly broad senescence rule is a design concern that could produce noisy tagging results; whether to fix now or in a follow-up is a judgment call, but it should at least be acknowledged.

Everything else is solid scaffolding work. The direction is right, code compiles, and the changes are well-scoped for a partial-progress PR.

scripts/config.py
```diff
@@ -166,6 +170,78 @@ def resilient_get(url: str, params: dict = None, timeout: int = 30, retries: int
...
+    "autophagy": [
+        "autophagy", "autophagic", "lysosomal degradation", "autophagosome",
+    ],
+    "senescence-sasp": [
+        "senescence", "senescent", "sasp", "senolytic",
+        "therapy-induced senescence",
+    ],
+    "tme-stroma": [
+        "tumor microenvironment", "tumour microenvironment", "cancer-associated fibroblast",
+        "caf", "extracellular matrix", "stromal barrier", "ecm remodeling",
```
ELares: Minor: `"caf"` is 3 chars so it hits the word-boundary regex path, which correctly prevents false positives on "scaffold" etc. However, `"cafs"` (the common plural abbreviation for cancer-associated fibroblasts) also will not match since the word boundary requires an exact standalone "caf". Most articles will still match via `"cancer-associated fibroblast"`, but adding `"cafs"` as a separate keyword would close the gap.

ELares (commented):
All 7 review comments addressed in `2a4129d`. Verified: handwritten files renamed (no overwrite conflict), senescence/stromal rules tightened, CAP added to keywords, MUFA logic extracted to helper, date-range comment added, "cafs" keyword added. Rust compiles clean, Python syntax checks pass. LGTM — merging.

PR #31, opened 2026-04-01
Rewrite README to be welcoming and thesis-neutral

ELares:
## Summary

Rewrite the README so it leads with the personal story, stays open about where the research goes, and makes it easy for newcomers to jump in.

## What changed

- **Personal story preserved exactly as written** — the volunteering passage and mission statement are untouched
- **Thesis-neutral framing** — replaced "proposing a specific and testable idea: that physical ROS-generating modalities (PDT, SDT) should be evaluated..." with "The goal is to let the evidence guide us—not to lock into one hypothesis from the start"
- **Simpler structure** — removed Current Priorities, Known Limits, Key Entry Points, and detailed repo structure tree (these are internal-facing and better suited for CLAUDE.md). Replaced with a clean navigation table and short section headings
- **Inviting contribute section** — "You don't need to be a cancer researcher—curiosity and a willingness to look at the evidence are enough"

## What stayed the same

The personal story, the MIT license note, and the core mission statement are all unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,23 +4,23 @@
 
 Ezequiel Lares
 
-## Repo Goal
+## What This Repo Is For
 
-Maintain this repository as an open cancer-research synthesis workspace that compares therapeutic mechanisms, evidence depth, resistant-state biology, pathway targets, and simulation-based hypotheses without assuming a single answer in advance.
+This is an open cancer-research workspace. The point is to help people—not to be right about one hypothesis.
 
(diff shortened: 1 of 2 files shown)


ELares (commented):
Re-reviewed this specifically against the README/story constraint. I do not see a blocking issue here.

The personal story still reads as your voice, and the repo framing is materially less scripted toward a single PDT/SDT answer than before. In particular, shifting from a concrete modality thesis to "let the evidence guide us" makes the README better aligned with the broader direction of the repo without flattening the mission behind it.

The tradeoff is that the new README is lighter on analysis-specific orientation than the previous version, but for a public-facing README I think that is acceptable as long as the deeper repo framing still lives in `CLAUDE.md` and the analysis docs. From the README alone, this now reads more like an open research workspace and less like a pitch for one predetermined conclusion, which seems to match the intent.

PR #51, opened 2026-04-02
Add evidence tagging gold-set evaluation

### LINKED ISSUE #33: Create gold-labeled evaluation set for evidence tagging quality
## Problem

The evidence classifier (keyword + pub_type heuristics) tags 41.8% of the corpus, but we have no measured precision/recall. The evidence-audit-sampling.md identified specific blind spots (non-phase clinical studies, mechanistic papers without standard keywords), but we don't know the overall error rate. Without this, claims about evidence-tier coverage are only as trustworthy as the heuristics.

## Proposed work

1. **Sample**: Draw a stratified random sample (~100-200 articles) from the unclassified primary-study-like pool, covering the 5 mechanisms with the most untagged records (immunotherapy, mRNA-vaccine, electrochemical-therapy, TTFields, synthetic-lethality)
2. **Label**: Manually classify each into the 7-tier evidence model (or "none applicable")
3. **Measure**: Compute precision, recall, and F1 for the current heuristic classifier against the gold labels
4. **Document**: Report error rates per mechanism and per tier. Identify the most common failure modes.

## Why this matters

- Turns "evidence tagging is heuristic" from a caveat into a quantified limitation
- Identifies whether the classifier systematically misses clinical studies (the main concern from the audit)
- Required before any publication claim about evidence-tier coverage

## Estimated effort

6-10 hours of manual labeling + 2-3 hours analysis. Can be done incrementally.

## Source

Recommended in external review (docs/openai.md). Builds on the manual spot-check in analysis/evidence-audit-sampling.md.

ELares:
Closes #33

## Summary

This PR adds a first-pass gold-labeled evaluation workflow for the repo's evidence-tier tagger.

Instead of relying only on qualitative caveats, the repo now has:
- a deterministic stratified sample sheet
- a manually labeled gold-set file
- an evaluation script that scores the current heuristic classifier against those labels
- a markdown artifact with overall metrics, per-tier metrics, per-mechanism exact accuracy, and concrete confusion examples

## What changed

- added `scripts/sample_evidence_gold_set.py`
- added `scripts/evaluate_evidence_gold_set.py`
- added `analysis/evidence-gold-set-v1.csv`
- added `analysis/evidence-gold-labels-v1.csv`
- added `analysis/evidence-gold-eval.md`
- updated `analysis/evidence-audit-notes.md` to point to the new measured evaluation layer

## Sampling design

The current gold set contains 100 rows:
- 10 `predicted-tagged` and 10 `predicted-untagged` rows for each of:
  - `immunotherapy`
  - `mRNA-vaccine`
  - `electrochemical-therapy`
  - `ttfields`
  - `synthetic-lethality`

This is intentionally a first-pass trust-building sample, not a final publication-grade benchmark.

## Main findings

From `analysis/evidence-gold-eval.md`:
- exact-label accuracy: `35.0%`
- binary evidence-detection precision: `96.0%`
- binary evidence-detection recall: `55.2%`
- binary evidence-detection F1: `0.701`

The practical read is that the current evidence tagger is conservative:
- if it tags evidence, it is usually detecting a real evidence-bearing paper
- if it does not tag evidence, that absence is much less trustworthy

The largest blind spots in this first sample are:
- `theoretical -> unclassified`
- `clinical-other -> unclassified`
- `preclinical-invitro -> unclassified`
- `none-applicable -> unclassified`

## Verification

- `python -m py_compile scripts/sample_evidence_gold_set.py scripts/evaluate_evidence_gold_set.py`
- `python scripts/sample_evidence_gold_set.py`
- `python scripts/evaluate_evidence_gold_set.py`

## Notes

- the label file is explicitly versioned as `v1` because this is a first-pass manual gold set and should be refined over time
- I left the unrelated untracked `docs/` directory out of this PR


ELares (commented):
## Review

Good work overall. The gold-set evaluation adds real accountability to the evidence tagger, the radioligand cleanup removes genuine false positives, and the designed-combination layer is a useful addition. A few issues worth addressing before merge.

### Issues

**1. `none-applicable` vs empty string inflates error rate (significant)**

The headline exact-accuracy metric (35%) is misleadingly low. The tagger outputs an empty string for unclassified records, but the gold set uses `none-applicable` as an explicit label. Since `"" != "none-applicable"`, all 11 correctly-untagged `none-applicable` papers are scored as wrong.

From the confusion table: `none-applicable -> unclassified: 11` — these are papers the tagger correctly left untagged, but they're counted as misses.

If you treat empty-string as equivalent to `none-applicable` for exact-match:
- Corrected exact accuracy: **46/100 (46%)** instead of 35/100
- Binary metrics are unaffected (they already handle this correctly)

Fix: either normalize empty string to `none-applicable` before comparison in `evaluate_evidence_gold_set.py`, or note this in the eval report.

**2. Per-label `none-applicable` metrics are misleading**

The per-label table shows `none-applicable` with TP=0, FP=0, FN=13 — suggesting it's never detected. But 11 of those 13 were correctly handled (left untagged). The per-label F1 of 0.000 for `none-applicable` is an artifact of the same string mismatch.

**3. Duplicate PMIDs across mechanism strata**

Articles can have multiple mechanisms. The sampling script doesn't deduplicate across mechanisms, so a PMID tagged with both `immunotherapy` and `mRNA-vaccine` could appear in both strata. The label file is keyed by PMID, so it would get the same gold label both times — inflating the effective sample for that paper and potentially skewing per-mechanism accuracy. Worth checking whether there are duplicates in the 100-row sample.

**4. Variable shadowing in evaluation script**

In `evaluate_evidence_gold_set.py`, the name `label` is used for both the `labels.get()` dict lookup (~line 40) and the `for label in LABELS` loop (~line 55). Different scopes so no bug, but confusing to read.

**5. PR bundles three distinct workstreams**

The PR title and description focus on the gold-set evaluation, but commits 1–2 are substantive changes to radioligand tagging and combination heuristics that touch ~1,970 corpus records and multiple analysis files. These are individually good changes but make the PR hard to review as a unit (1,997 changed files).

**6. Minor: inconsistent division-by-zero handling**

The script defines `safe_ratio()` but also uses inline `tp + fp or 1` patterns in the report-writing section. Using `safe_ratio` consistently would be cleaner.

### Bottom line

The `none-applicable` scoring issue (#1) is the most actionable — it makes the headline metric inaccurate in a way that understates tagger performance. Worth fixing before merge since the 35% number will be referenced in audit notes and potentially the manuscript.

ELares:
Follow-up cleanup is pushed.

Changes made:
- normalized empty heuristic predictions to `none-applicable` for exact/per-label scoring
- added label-file validation for duplicate PMIDs and unknown gold labels
- added a duplicate-PMID guard in the sampler so the benchmark cannot silently sample the same article twice
- regenerated `analysis/evidence-gold-eval.md` with the corrected exact accuracy (`46/100`, previously `35/100`)
- rebased the branch onto `main`, so this PR now contains only the gold-set work plus the follow-up fix

I left the binary precision/recall metrics unchanged because they were already computed from positive-vs-empty predictions and were not affected by the `none-applicable` normalization.

ELares (commented):
## Re-review after fixes

All six issues from the first review are addressed:

1. **`none-applicable` semantics** — Fixed via `normalized_prediction()`. Exact accuracy now correctly reports 46/100. The explanatory note in the eval report is a good touch.
2. **Per-label `none-applicable` metrics** — Now shows TP=11, FP=39, FN=2, which correctly reflects that untagged papers are evidence-bearing 78% of the time. Much more informative than the previous 0/0/13 artifact.
3. **Duplicate PMIDs** — Both scripts now guard against duplicates (sampling script tracks `sampled_pmids`, evaluation script validates label file uniqueness).
4. **Variable shadowing** — `label` → `label_row`. Clean.
5. **PR scope** — Reduced from 1,997 files to 6. Radioligand and combination work separated out.
6. **Division-by-zero** — Inline `or 1` patterns removed in favor of bare denominators. `safe_ratio` handles the percentage calculation. Minor note: if someone runs this on a gold set where all golds are `none-applicable`, the display would show `0/0` — but that's a degenerate input, not a real concern.

Also good: unused imports removed, unused `EVIDENCE_LEVELS` removed from sampling script, gold label validation added.

LGTM.

PR #96, opened 2026-04-20
Emergent LP overshoot: post-death biochemistry in ferroptosis-core

ELares:
## Summary
The first change to ferroptosis-core itself. LP at death is now treatment-dependent, emerging from the autocatalytic chain reaction dynamics instead of the estimated Option C multiplier from PR #84.

## What changed
After LP crosses the death threshold (10.0), the simulation continues for 5 more steps with LP-only accumulation (no GSH scavenging, no repair — the cell's defenses have failed). High-ROS treatments drive LP much higher than slow-accumulation treatments.

## Results (sim-original, 1M cells)
| Phenotype | Treatment | LP before | LP after | Death rate |
|-----------|-----------|-----------|----------|------------|
| Persister | SDT | 10.6 | **23.4** | 100.0% (unchanged) |
| Persister | RSL3 | 4.6 | **8.9** | 42.5% (unchanged) |
| OXPHOS | SDT | 10.3 | **25.4** | 99.9% (unchanged) |
| Glycolytic | SDT | 8.9 | **15.7** | 87.2% (unchanged) |

SDT/RSL3 LP ratio for Persisters: **2.6x** (emergent) — close to the estimated 2.0x from Option C but now derived from the actual biochemistry.

## sim-tme Option C removed
- Deleted physical_modality_overshoot (2.0) and pharmacologic_overshoot (1.05) from ImmuneConfig
- DAMP release deferred to end of post-death grace period using accumulated LP
- SDT immune kills: 521 (was 539 with Option C) — slight decrease, direction preserved
- Removed lp_overshoot_multiplier from ConditionResult

## Files changed (5)
| File | Change |
|------|--------|
| ferroptosis-core/src/params.rs | Add post_death_steps: u32 (default 5) |
| ferroptosis-core/src/biochem.rs | Post-death LP accumulation in sim_cell_step and sim_cell |
| sim-tme/src/main.rs | Remove Option C, deferred DAMP release, grace period skip |
| sim-spatial/src/main.rs | Grace period skip logic |
| simulation_results_v1_ground_truth.json | Regenerated (16 rows) |

## Blast radius
This changes LP output values for every binary that reports LP. Death rates are UNCHANGED. The ground truth has been regenerated.

## Test plan
- [x] 19 ferroptosis-core tests pass (none assert on post-threshold LP)
- [x] All workspace binaries compile (excluding ferroptosis-python link)
- [x] sim-original: SDT LP ~15-25 (emergent), RSL3 LP ~0-9 (modest)
[...]

PR #97, opened 2026-04-20
Update manuscript for emergent LP + fix PyPI Linux build

ELares:
## Summary
Two fixes:

### 1. Manuscript aligned with emergent LP overshoot (PR #96)

Section 4.3.5 (immune coupling) updated:
- LP overshoot now described as emergent (5 post-death steps, zero repair) instead of estimated multiplier
- SDT Persister LP ~20 (emergent 2.0x), RSL3 ~7.8
- SDT immune kills: 539 → **521** (slightly lower with emergent vs estimated overshoot)
- RSL3 immune kills: 2 → **5** (slight increase due to RNG shift from post-death steps)
- SDT/RSL3 ratio: 270:1 → **104:1** (still two orders of magnitude)
- Anti-PD1 SDT: 1,520 → **1,477**, amplification still 2.8×
- Caveat updated: "emergent from biochemistry" replaces "estimated multiplier"

Conclusion finding #2(d): 270× → 104×, with absolute numbers.
README: 270× → 104×.

### 2. PyPI workflow: fix Linux manylinux build

Root cause: `actions/setup-python` installs Python on the host, but manylinux builds use a Docker container where the host Python isn't visible (`Couldn't find any python interpreters from 'python3'`).

Fix: `if: runner.os != 'Linux'` on the setup-python step. The manylinux Docker container has Python built in.

## Test plan
- [x] pdflatex compiles (0 errors, 45 pages, 106 citations)
- [x] 33 smoke tests pass
- [x] All updated numbers match tme_summary.json
- [ ] Re-trigger PyPI workflow after merge to verify Linux build passes


ELares (commented):
Reviewed on the latest commit. I reran `pdflatex -interaction=nonstopmode -halt-on-error v1.tex` and `python -m pytest tests/test_pipeline_smoke.py -q`; both passed. I also spot-checked the updated immune-coupling, stromal, and pH values against `simulations/output/tme/tme_summary.json`, and the revised manuscript/README numbers now line up with the current outputs.

I do not see a remaining content blocker from this pass. The workflow change is also directionally correct: the Linux manylinux build path is the only place where host `setup-python` is plausibly unnecessary, while the non-Linux build/test jobs still keep an explicit Python setup.

PR #106, opened 2026-04-22
Add tumor PBPK compartment module (Phase 1 MVP)

ELares:
## Summary
Phase 1 of issue #48: a two-compartment tumor PK model that captures time-dependent drug delivery from plasma to tumor interstitium.

## Key finding
Tumor-type-specific pharmacokinetics create large protection factors relative to a constant max-exposure reference (drug at full concentration for all 180 steps, ~100% kill):

| Tumor Type | Peak C_i | AUC | Death Rate | Protection |
|-----------|---------|-----|------------|-----------|
| Max exposure ref | 1.000 | 180 | 100.0% | reference |
| Breast | 0.409 | 20 | 4.2% | 24x |
| Pancreatic | 0.237 | 12 | 2.9% | 35x |
| GBM | 0.124 | 6 | 2.4% | 42x |

**Important**: The max-exposure reference is NOT the repo's standard 2D RSL3 baseline (~42.5% from sim-original). The clamp model applies continuous GPX4 inhibition (preventing NRF2 recovery), while sim_cell applies one-time reduction at init. Protection factors are relative to this max-exposure reference. Calibrating the clamp model to match the init-based baseline is Phase 2 work.

## New module: ferroptosis-core/src/tumor_pk.rs
- TumorPKParams: 2-compartment ODE parameters (15 fields)
- PlasmaModel: IV bolus and constant concentration analytical curves
- solve_tumor_pk(): Euler ODE solver (100 substeps/min for stability)
- sim_cell_with_pk(): ferroptosis simulation with time-varying drug (GPX4 clamped to drug-allowed maximum)
- 3 tumor presets: breast, pancreatic (IFP=100mmHg), GBM (BBB)
- 5 unit tests

## New binary: sim-tumor-pk
3 tumors + max-exposure reference x 10K Persister cells. CSV timecourse + JSON summary.

## Critical parameterization fix
Issue #48 proposes cellular uptake as k_uptake x rho = 0.01 x 1e8 = 1e6/min, which drains the interstitial space in microseconds. Fixed: use bulk uptake rate (0.02/min).

## Blast radius
Zero changes to existing modules. Purely additive.

## Test plan
- [x] 24 tests pass (19 existing + 5 new)
- [x] sim-tumor-pk runs successfully
- [x] Protection factor ordering: breast < pancreatic < GBM
- [x] Baseline honestly labeled as max-exposure reference (not 2D culture)

Partial progress on #48

Issue #48, opened 2026-04-01
Simulation: tumor PBPK compartment module


ELares:
## Problem

PK-Sim (Open Systems Pharmacology, free, GPLv2, actively maintained) is the gold standard for whole-body PBPK modeling with ~15 organ compartments. It can predict drug plasma concentration over time. But it has no built-in tumor compartment. To model drug delivery to a tumor, you must export to MoBi and manually construct the tumor as a custom compartment — with no standard parameterization for cancer-specific pathophysiology.

This matters because tumors are not normal tissue. They have:
- **Abnormal vasculature** — leaky, tortuous, heterogeneous perfusion
- **Elevated interstitial fluid pressure (IFP)** — opposes convective transport, reduces drug penetration
- **Heterogeneous perfusion** — some regions well-perfused, others avascular/necrotic
- **Evolving drug resistance** — efflux pumps, metabolic adaptation, lipid remodeling
- **Lymphatic dysfunction** — impaired drainage, contributes to IFP

No open-source tool provides a validated, reusable tumor compartment with these features.

## Why this matters for the scientific community

Every cancer drug must travel from injection site → plasma → tumor vasculature → interstitial space → cancer cell. The whole-body PK determines plasma concentration. The tumor compartment determines how much of that plasma concentration actually reaches cancer cells. A reusable, open-source tumor compartment module that researchers could plug into their PBPK workflows would accelerate drug development modeling.

## What already exists (complement, don't compete)

| Tool | What it does | Why we don't rebuild it |
|------|-------------|------------------------|
| **PK-Sim** | Whole-body PBPK, 15 organ compartments, validated for 100+ drugs | Free, excellent, actively maintained. We consume its output. |
| **mrgsolve** (R) | ODE-based population PK/PD | Flexible but no spatial component. |
| **SpatialCNS-PBPK** (R/Shiny, 2025) | 9-compartment CNS PBPK for brain tumors | Brain-only; demonstrates the value of tissue-specific PBPK modules. |

## Proposed work

Build a Rust-based tumor compartment module that:

### 1. Takes plasma concentration time-course as input

From PK-Sim (or any PBPK tool), export `C_plasma(t)` — drug concentration in arterial plasma over time. This is the input boundary condition.

### 2. Models tumor-specific drug delivery

**Two-compartment tumor model (well-mixed):**

For a first pass, model the tumor as two linked compartments:

```
Vascular space (V_v):
[...]

PR #107, opened 2026-04-22
Fix tumor PBPK 2D baseline: inactivation rate model calibrated to 42%

ELares:
## Summary
Fixes the 2D baseline calibration issue from PR #106 review. The clamp model gave 100% kill instead of the expected ~42.5%. Now uses a pharmacologically correct GPX4 inactivation rate model.

## The fix
Replace the GPX4 clamp (state.gpx4 = min(state.gpx4, ceiling) every step) with a covalent inactivation rate:
```
state.gpx4 -= k_inact * conc * state.gpx4
```

## Calibration
k_inact=0.015 calibrated directly in Rust to match the Persister+RSL3 **death rate** (~42.5%):

| k_inact | Death rate |
|---------|-----------|
| 0.010 | 20.8% |
| 0.012 | 28.8% |
| **0.015** | **41.4%** |
| 0.018 | 54.9% |

**Important caveat**: the 2D reference matches the death RATE but not the internal state. LP is 8.61 (vs sim_cell 4.62), GPX4 is 0.255 (vs 0.500). The continuous inactivation model produces a different trajectory than the one-time init reduction. All documentation explicitly says 'matching death rate' not 'matching baseline.'

## Results
| Tumor Type | Death Rate | Protection Factor |
|-----------|-----------|------------------|
| 2D culture ref | 41.4% | reference |
| Breast | 2.6% | 15.7x |
| Pancreatic | 2.0% | 21.0x |
| GBM | 1.6% | 26.7x |

## Test plan
- [x] 24 tests pass
- [x] 2D reference death rate: 41.4% (within 5% of target 42.5%)
- [x] Protection factor ordering: breast < pancreatic < GBM
- [x] All wording says 'matching death rate' (4 sites checked)
- [x] Steady-state GPX4 comment corrected (0.20-0.25)

Partial progress on #48

### DIFF
--- a/simulations/ferroptosis-core/src/tumor_pk.rs
+++ b/simulations/ferroptosis-core/src/tumor_pk.rs
@@ -172,7 +172,9 @@ pub fn rsl3_iv_bolus() -> PlasmaModel {
 
 /// 2D culture reference: constant drug concentration = 1.0 for all time.
 /// With the inactivation rate model (k_inact=0.015), this produces ~41%
-/// kill for Persisters — matching the repo's standard sim_cell RSL3 baseline.
+/// kill for Persisters — matching the repo's Persister+RSL3 death rate (~42.5%).
+/// Note: internal state (LP, GSH, GPX4) differs from sim_cell due to the
+/// continuous inactivation model vs one-time init reduction.
 pub fn constant_reference() -> PlasmaModel {
     PlasmaModel::Constant { concentration: 1.0 }
 }
@@ -283,7 +285,8 @@ pub struct PKCellResult {
 
 /// GPX4 inactivation rate for RSL3-like covalent GPX4 inhibitors.
 /// Calibrated directly in Rust (10K Persister cells, constant conc=1.0):
-/// k_inact=0.015 gives ~41% kill, matching sim_cell RSL3+Persister (~42.5%).
+/// k_inact=0.015 gives ~41% death rate, matching sim_cell RSL3+Persister
+/// death rate (~42.5%). Internal state (LP, GSH, GPX4) differs.
 pub const RSL3_INACTIVATION_RATE: f64 = 0.015;
 
 /// Simulate a single cell with time-varying drug concentration.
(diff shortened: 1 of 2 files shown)


ELares (commented):
Reviewed on the latest commit. The previous max-exposure problem is fixed, but I still do not think this is merge-ready yet.

Main issue: the new `2D culture ref` is calibrated only to the persister/RSL3 death rate, not to the repo's existing persister/RSL3 baseline state. I reran both paths locally:

- `cargo run --release -p sim-tumor-pk` gives `2D culture ref: death_rate=41.4%`, `LP=8.61`, `GSH=1.14`, `GPX4=0.255`
- `python - <<'PY' ... fc.sim_batch("Persister", "RSL3", n=10000, seed=42) ... PY` gives the established baseline around `42.3%`, but with `mean_lp=4.62`, `mean_gsh=1.42`, `mean_gpx4=0.500`

So the new PK reference matches one scalar endpoint, but it does not reproduce the actual internal state of the existing 2D baseline. That means calling it a match to the repo's standard `sim_cell`/binding baseline is still too strong. Right now this is better described as a newly calibrated continuous-exposure reference that reproduces the old death rate, not as a true equivalent baseline.

1. narrow the wording everywhere from “matching the repo's standard baseline” to “matching the repo's persister/RSL3 death rate”, or
2. calibrate against additional observables so the reference actually reproduces the old baseline state, not just kill fraction.

ELares:
Fixed the overclaim from review:

All 4 sites now say 'matching death rate' (not 'matching baseline') with explicit notes that internal state (LP, GSH, GPX4) differs:
- LP: 8.61 (PK model) vs 4.62 (sim_cell) 
- GPX4: 0.255 vs 0.500
- GSH: 1.14 vs 1.42

The continuous inactivation model produces a different GPX4 trajectory than the one-time init reduction — they converge on the same death FRACTION but through different internal dynamics.

ELares (commented):
Re-reviewed on the latest commit. The remaining issue from my previous pass is resolved.

The wording is now disciplined enough: the PR and code explicitly say the new `2D culture ref` matches the Persister+RSL3 death rate, not the full baseline state, and they call out the LP/GSH/GPX4 mismatch directly instead of implying true equivalence. I reran `cargo run --release -p sim-tumor-pk`; the behavior is unchanged (`41.4%` for the 2D reference, then `2.6%` / `2.0%` / `1.6%` for breast / pancreatic / GBM), and I do not see a remaining merge blocker from this pass.

PR #108, opened 2026-04-22
Add spatial×temporal C(r,t) drug delivery model (Phase 2D)

ELares:
## Summary
Composes the temporal PK model (C_i(t) from tumor_pk ODE) with the spatial Krogh decay (exp(-r/lambda)) to produce C(r,t) at distance r from nearest vessel at time t.

## Key finding: temporal PK dominates spatial penetration

Radial sweep restricted to each tissue's physiological half-distance (r <= r_max):

| Tumor | At vessel (r=0) | At r_max | Spatial addition |
|-------|----------------|----------|-----------------|
| Breast | 15.7x | 18.0x (r=50um, r_max=60) | 1.1x more |
| Pancreatic | 21.0x | 25.9x (r=125um, r_max=125) | 1.2x more |
| GBM | 26.7x | 28.6x (r=75um, r_max=75) | 1.1x more |

For small molecules with ~30 min half-life, drug washes out before spatial gradients become limiting. lambda_met (224 um) is large relative to inter-vessel spacing.

**Implication**: improving drug EXPOSURE TIME matters more than improving drug PENETRATION.

## Double-counting prevention
lambda_met = sqrt(D / metabolism_only) = 224 um (not lambda_full = 100 um which includes uptake already handled by the temporal ODE).

## New code
- metabolism_only_penetration_um(): lambda from metabolism rate only
- compute_spatial_temporal_schedule(): C(r,t) = C_i(t) x exp(-r/lambda_met)
- Radial sweep in sim-tumor-pk restricted to r <= r_max per tissue
- 3 new unit tests (27 total)

## Test plan
- [x] 27 tests pass
- [x] At r=0: matches Phase 1 results exactly
- [x] Kill rates monotonically decrease with distance
- [x] Radial bins restricted to r <= r_max (no bins past tissue half-distance)

Partial progress on #48

### DIFF
--- a/simulations/ferroptosis-core/src/tumor_pk.rs
+++ b/simulations/ferroptosis-core/src/tumor_pk.rs
@@ -270,6 +270,55 @@ pub fn solve_tumor_pk(
 }
 
 // ============================================================
+// Spatial × temporal composition: C(r,t)
+// ============================================================
+
+/// Compute spatial penetration length using METABOLISM ONLY (μm).
+/// For C(r,t) composition where cellular uptake is handled by the temporal
+/// ODE — avoids double-counting uptake in both spatial decay and temporal
+/// clearance. Returns a LONGER λ (224 μm for RSL3) than the full
+/// drug_transport::penetration_length_um (100 μm) which includes uptake.
+pub fn metabolism_only_penetration_um(drug: &crate::drug_transport::DrugParams) -> f64 {
+    let d_um2_per_s = drug.diffusion_coeff_cm2_s * 1e8;
+    if drug.metabolism_rate <= 0.0 {
+        return f64::INFINITY;
+    }
+    (d_um2_per_s / drug.metabolism_rate).sqrt()
+}
+
+/// Compute a time-varying concentration schedule for a cell at radial
+/// distance r_um from the nearest vessel.
+///
+/// Uses the quasi-steady approximation: C(r,t) ≈ C_i(t) × exp(-r / λ_met).
+/// Valid when diffusion equilibrates faster than plasma PK changes (~3 min
+/// diffusion vs ~30 min plasma half-life → 10× faster, valid after ~10 min).
+///
+/// λ_met uses metabolism-only clearance (not uptake) because the temporal
+/// ODE already includes cellular uptake. This avoids double-counting and
+/// produces a longer penetration length (224 μm vs 100 μm for RSL3).
+///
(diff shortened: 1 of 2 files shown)


ELares (commented):
Reviewed on the latest commit. The new composition is directionally sensible, but I do not think this is merge-ready yet.

Main issue: the PR summary presents the “at r_max” protection factors as if they were evaluated at each tissue's stated half-distance, but the binary is actually sampling past `r_max` for at least breast and GBM. I reran `cargo run --release -p sim-tumor-pk`; current output is:

- Breast: `r_max=60μm`, but the strongest reported spatial effect comes from `r=75μm` (`15.7x -> 19.3x`)
- Pancreatic: `r_max=125μm`, and the `25.9x` value is correctly at `r=125μm`
- GBM: `r_max=75μm`, but the PR body currently reports `26.9x (r=25μm)` as if that were the edge case; the actual largest sampled effect is `28.6x` at `r=75μm`

This comes directly from the loop condition in `sim-tumor-pk/src/main.rs`: it keeps any radial bin with `r <= r_max * 1.5`, so the breast sweep includes `75μm` despite `r_max=60μm`. That makes the current “at r_max” table misleading, and it also makes the headline “spatial addition = 1.2x more” less clean than the PR suggests.

1. restrict the sweep/reporting to `r <= r_max` so the edge values are true tissue half-distances, or
2. keep the wider sweep, but relabel the table honestly as “max sampled radius” / “largest sampled spatial effect” instead of `r_max`.

ELares:
Fixed the radial sweep cutoff from review:

Changed `r > r_max * 1.5` to `r > r_max` so all reported bins are within the tissue's physiological half-distance. PR body updated with corrected edge values.

Breast now stops at r=50um (within r_max=60), GBM at r=75um (within r_max=75). Pancreatic correctly reaches r=125um (equals r_max).

ELares (commented):
Re-reviewed on the latest commit. The previous blocker is resolved.

The sweep/reporting now matches the stated tissue half-distance framing: the binary only samples `r <= r_max`, and the PR body is honest about the discrete-bin case for breast (`18.0x` at `r=50μm`, with `r_max=60μm`). I reran `cargo run --release -p sim-tumor-pk`; the current spatial outputs line up with the updated summary.

I do not see a remaining merge blocker from this pass. One low-severity cleanup item remains: `sim-tumor-pk` still emits an unused-variable warning for `mean_lp`, but that is not enough to block the PR.

Issue #99, opened 2026-04-20
News & blog integration: ingestion pipeline and claim verification engine


ELares:
## Context

Depends on #98 (source criteria and authentication framework). This issue covers the technical implementation of fetching, parsing, scoring, and verifying news articles.

### 1. Article fetcher (`scripts/fetch_news.py`)

Fetches articles from configured sources via:
- RSS/Atom feeds (most Tier 1-2 sources have feeds)
- Direct URL fetching with readability extraction (for blogs without feeds)
- Wayback Machine fallback for archived content

Output: markdown files in `news/by-source/{source}/{date}-{slug}.md` with frontmatter matching the schema from #98.

Configuration: `scripts/news_sources.yaml` with per-source settings (URL, feed type, tier, category).

### 2. Claim extractor (`scripts/extract_claims.py`)

NLP-based extraction of factual claims from article text:
- Identify sentences containing statistics, trial names, drug names, gene names
- Classify each as factual / interpretive / speculative
- Extract entity mentions (drugs, genes, cancer types, institutions)
- Link to mechanism taxonomy from `scripts/config.py` where applicable

Approach: keyword-based for v1 (same pattern as our evidence tagger — 96% precision, 55% recall is acceptable for initial pass). LLM-assisted extraction as a follow-up.

### 3. Claim verifier (`scripts/verify_news_claims.py`)

Cross-reference extracted claims against:
- Local corpus (INDEX.jsonl) — match by PMID, DOI, author name, trial name
- PubMed API — search for underlying primary source
- ClinicalTrials.gov — verify trial claims

Output: per-claim verification status (verified/unverified/disputed) with linked PMIDs.

### 4. Credibility scorer (`scripts/score_news.py`)

Compute per-article credibility score from:
- Source tier (from #98)
- Verified claim ratio
- Author credentialing (known vs unknown)
- Cross-citation count (other sources reporting same finding)

Output: credibility_score field in frontmatter (0.0-1.0).

### 5. News index builder

Append to `news/NEWS_INDEX.jsonl` with full metadata:
- source, author, date, url, tier
- credibility_score, verified_claims, unverified_claims
- linked_pmids, extracted_entities
- categories (mechanism, cancer_type from taxonomy)

## Acceptance criteria
- [ ] Fetcher retrieves articles from at least 5 Tier 1-2 sources
- [ ] Claim extractor identifies statistical and factual claims
- [ ] Verifier links claims to PMIDs where possible
[...]

Issue #98, opened 2026-04-20
News & blog article ingestion: source criteria and authentication framework


ELares:
## Context

The repo currently draws exclusively from PubMed-indexed peer-reviewed literature (4,830 full-text articles). However, significant cancer research context exists in trusted news outlets, institutional blogs, and science journalism that captures:
- Clinical trial results before formal publication
- Patient-perspective narratives that reveal translational gaps
- Policy and funding shifts that shape research priorities
- Industry announcements (drug approvals, pipeline failures) with real-world timing
- Expert commentary that contextualizes findings for broader audiences

Incorporating this material would strengthen the manuscript's claim to be a comprehensive, accessible resource — but ONLY if the articles are authenticated and their claims are verified against primary literature.

## Problem

Not all internet content is reliable. We need a rigorous framework to:
1. Define which sources are trustworthy enough to cite
2. Score the credibility of individual articles
3. Verify factual claims against our existing corpus and PubMed
4. Distinguish reporting (factual) from opinion (interpretive) from speculation (unverified)
5. Flag and exclude misinformation, hype, or industry-funded promotion masquerading as journalism

### Tier 1 — Institutional and peer-adjacent (high trust, cite directly)
- **NIH/NCI newsroom** (cancer.gov, nih.gov/news-events)
- **WHO cancer fact sheets**
- **FDA drug approval announcements**
- **Clinical trial registries** (clinicaltrials.gov announcements)
- **University press offices** (Harvard, Stanford, MD Anderson, etc. — tied to published papers)
- **Nature/Science/Cell news sections** (journalistic arms of peer-reviewed publishers)

### Tier 2 — Science journalism (medium trust, cite with verification)
- **STAT News** (cancer/biotech beat)
- **The Cancer Letter** (policy and clinical oncology)
- **Endpoints News** (clinical trials and biotech)
- **Science Daily / Medical News Today** (aggregators — verify underlying paper)
- **Reuters Health / AP Health** (wire service health desks)
- **Ars Technica science** (technology-adjacent science journalism)

### Tier 3 — Expert blogs and commentary (use for context, not facts)
- **Derek Lowe's "In the Pipeline"** (pharmaceutical chemistry commentary)
- **Vinay Prasad's blog/Substack** (oncology clinical trial criticism)
- **Patient advocacy organization blogs** (ASCO, ACS, LLS)
- **Institutional research blogs** (Broad Institute, Salk, etc.)

### Excluded — Do not ingest
[...]

PR #120, opened 2026-04-23
Build news authentication pipeline with claim-level tracking (#99)

ELares:
## Summary

Implements the news source authentication pipeline defined in the criteria framework (#98). 5 new Python scripts that fetch, extract claims from, verify, score, and index news articles. Claim-level tracking per the user's 6-point constraint.

### Pipeline

| Script | Purpose | Key feature |
|--------|---------|-------------|
| fetch_news.py | Fetch by URL or RSS | Longest-prefix source classification, content hash versioning, 4xx error rejection |
| extract_claims.py | Two-path claim extraction | Path 1: factual markers (numbers, endpoints, trial phases). Path 2: high-specificity opinion/speculation phrases. All 3 categories extracted. |
| verify_news_claims.py | Corpus + PubMed verification | Searches INDEX.jsonl then PubMed API, links PMIDs, detects self-referencing institutional sources |
| score_news.py | Credibility scoring | 4-component formula from criteria doc, self-referencing counts as verified, zero-claims edge case |
| build_news_index.py | Claim-centric JSONL index | One entry per claim with denormalized article metadata, dated entries newest-first |

### User's 6 constraints addressed

1. Claims first-class: claim-centric NEWS_INDEX.jsonl with all 3 categories (FACTUAL, INTERPRETIVE, SPECULATIVE) and all 5 types (event, result, mechanism, opinion, speculation)
2. Manual review: review_status field + --approved-only index filter
3. Provenance: canonical_url, date_retrieved, paywall, storage_mode, content_hash
4. Claim types: event/result/mechanism/opinion/speculation — all extractable
5. Version drift: SHA-256 content hash + version linking on re-fetch
6. No LLM: keyword-based extraction only (factual markers + multi-word phrase triggers)

### Verification statuses (per criteria doc)

- verified: primary source found via corpus or PubMed
- self-referencing: institutional source IS the authority (WHO, NCI, FDA, IARC, NIH)
- unverified: no primary source found
- disputed: reserved for future use

### Tests

**50 tests total, all pass** (33 existing + 17 new). Covers: source classification (4 tiers + longest prefix), sentence splitting, factual marker detection, claim type classification, opinion/speculation trigger matching, scoring formula, zero-claims edge case, verify module imports, search term extraction, config additions.

### What is deferred to follow-up PR

- 3 criteria-doc example articles processed through pipeline
- 50-article batch via RSS ingestion
- ClinicalTrials.gov verification
- Cross-citation scoring
- DISPUTED status detection (requires semantic comparison)

## Test plan

- [x] pytest tests/ -- 50 passed
- [x] Source classification correctly maps URLs to tiers
- [x] Factual claims extracted by number/endpoint patterns
- [x] Opinion/speculation claims extracted by multi-word phrase triggers
- [x] Self-referencing detection for institutional sources
- [x] Scoring formula produces expected values
- [x] No changes to manuscript or existing corpus pipeline

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,5 @@ PyMuPDF>=1.24.0
 matplotlib>=3.8.0
 numpy>=1.26.0
 scipy>=1.12.0
+beautifulsoup4>=4.12.0
+feedparser>=6.0.0
--- a/scripts/build_news_index.py
+++ b/scripts/build_news_index.py
@@ -96,12 +96,11 @@ def build_index(approved_only: bool = False) -> int:
             entry.update(article_meta)
             entries.append(entry)
 
-    # Sort by date descending (newest first), then by claim_id
-    def _sort_key(e: dict) -> tuple:
-        d = e.get("date_published") or ""
-        return (d == "", d)  # empty dates sort last
-
-    entries.sort(key=_sort_key, reverse=True)
+    # Sort: dated entries newest-first, undated entries at the end
+    dated = [e for e in entries if e.get("date_published")]
+    undated = [e for e in entries if not e.get("date_published")]
+    dated.sort(key=lambda e: e["date_published"], reverse=True)
+    entries = dated + undated
 
     # Write index
     index_path = NEWS_DIR / "NEWS_INDEX.jsonl"
--- a/scripts/build_news_index.py
+++ b/scripts/build_news_index.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Build claim-centric NEWS_INDEX.jsonl from processed articles.
+
+Usage:
+    python build_news_index.py
+    python build_news_index.py --approved-only
+"""
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+from tqdm import tqdm
+
+# Allow imports from the scripts directory
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from config import (
+    SOURCE_TIER_DEFINITIONS,
+    NEWS_RATE,
+    NEWS_DIR,
+    CLAIM_FACTUAL_MARKERS,
+    CLAIM_TYPE_MARKERS,
+    resilient_get,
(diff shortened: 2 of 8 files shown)


ELares (commented):
Reviewed on the latest commit.

Current verdict:
- not fully merge-ready yet

Findings:
1. Medium: the pipeline does not actually track all 5 claim types as first-class indexed claims, even though the PR summary says it does. In `scripts/extract_claims.py`, `extract_claims()` only emits a claim when the sentence matches a factual marker, then uses the type markers only for classification. That means `opinion` and `speculation` can only appear if a sentence is also fact-pattern-matched, and purely speculative / interpretive claims are dropped before indexing. Concretely, the gate is:
   - `if not has_factual: continue`
This is a real design mismatch with the stated contract in the PR body (`5 claim types`) and with the earlier issue framing around claim-level tracking. Right now the claim index is effectively factual-claim-centric, not general claim-centric. If that is intentional, the PR/body should say so explicitly. If it is not intentional, extraction needs a second path for non-factual claim types.

2. Low: the PR summary’s test count is stale. I reran `python -m pytest tests -q` and the current result is `48 passed`, not `46 passed`.

What I verified:
- `python -m pytest tests -q` passes with `48 passed`
- the news scripts import cleanly under the current test environment

So the main remaining issue is the claim-extraction contract: the implementation and the PR description currently disagree about whether non-factual claim types are actually captured.

ELares (commented):
Re-reviewed on the latest commit. The substantive extraction-contract issue is fixed now. `extract_claims.py` no longer hard-gates everything on factual markers; it has a real two-path flow, so high-specificity opinion/speculation phrases can produce non-factual claims directly, while factual-marker sentences still go through the factual path. That brings the implementation in line with the PR summary's claim that all three categories are actually extractable.

I also reran `python -m pytest tests -q`; it passes with `50 passed` on this checkout, so the test-count metadata is now aligned too. I do not see a remaining merge blocker from this pass.

PR #144, opened 2026-04-23
Add CI workflow for Python smoke tests

### LINKED ISSUE #126: Add CI workflow for Python smoke tests
## Source
Review 1 (Deep Research Assessment), Review 3 (Deep Review).

## Problem
The Rust tests run in CI via `cargo-test.yml`, but the 50 Python smoke tests (38 test functions across pipeline and news modules) have no CI workflow. They only run locally. This means a PR could break the Python pipeline without anyone noticing until after merge.

## What Needs to Be Done
1. Add `.github/workflows/pytest.yml` that runs `python3 -m pytest tests/ -v` on push/PR when `scripts/**`, `tests/**`, or `requirements.txt` change
2. Install dependencies from `requirements.txt` in the workflow
3. Consider running on both Ubuntu and macOS runners

### Owner Refinements
- **Linux-only on PR/push; macOS as scheduled/follow-on.** The CI workflow triggered on PR and push must run on Ubuntu (Linux) only. macOS runners are expensive and flaky — add macOS only as a scheduled (e.g., weekly) job or as a follow-on issue. Do not include macOS in the PR/push matrix.
- **Explicitly cover news-pipeline tests.** The workflow must run tests for both the main pipeline (`tests/`) and the news pipeline modules. Ensure the test command covers all test directories, not just `tests/`.
- **Avoid flaky-matrix maintenance.** Do not create a multi-OS matrix that requires ongoing maintenance. A single Linux runner on PR/push is the target. Keep the workflow simple and reliable.

## Acceptance Criteria
- [ ] `.github/workflows/pytest.yml` exists and triggers on push/PR for relevant path changes
- [ ] Workflow runs on Ubuntu (Linux) only for PR/push triggers
- [ ] Workflow installs dependencies from `requirements.txt`
- [ ] Workflow runs all Python tests, including news-pipeline tests
- [ ] (Optional/follow-on) macOS runner added as a scheduled job (e.g., weekly cron)
- [ ] No multi-OS matrix on PR/push triggers

## Priority
**High** — The 50 Python tests already exist. Wiring them to CI is low effort, high value.

## Category
Infrastructure / CI

ELares:
## Summary
- Adds `.github/workflows/python-test.yml` running all 50 Python smoke tests
- **Linux** on every push to main and every PR (path-filtered)
- **macOS** on weekly schedule only (Monday 8am UTC) — owner directive: too expensive for PR path
- Uses `requirements-lock.txt` for deterministic installs, Python 3.12, `lfs: false`, `fetch-depth: 1`

## Design decisions
- **Path filter:** `scripts/`, `tests/`, `corpus/INDEX.jsonl`, `analysis/`, `requirements*.txt`, self. Excludes `news/**` (tests import from scripts/ but don't read from news/).
- **No pip cache:** Install is ~30s. Caching adds complexity for negligible gain.
- **`python -m pytest` not `pytest`:** Ensures correct interpreter.

## Known items
- If `requirements-lock.txt` fails on Linux (generated on macOS), regenerate with: `python3 -m venv /tmp/lock-env && /tmp/lock-env/bin/pip install -r requirements.txt && /tmp/lock-env/bin/pip freeze > requirements-lock.txt`
- `test_score_formula_tier1` has a hardcoded date (2026-04-01) — pre-existing future flake, not introduced here
- Discovered: `cargo-test.yml` uses `@stable` instead of pinned 1.92.0 from `rust-toolchain.toml` — separate micro-fix needed

## Test plan
- [ ] Workflow self-triggers on this PR (`.github/workflows/python-test.yml` in path filter)
- [ ] All 50 tests pass on ubuntu-latest
- [ ] Total time < 3 minutes
- [ ] No LFS download in checkout step logs

Closes #126

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -23,6 +23,7 @@ on:
 
 jobs:
   test-linux:
+    if: github.event_name != 'schedule'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -1,0 +1,64 @@
+name: Python Tests
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'scripts/**'
+      - 'tests/**'
+      - 'corpus/INDEX.jsonl'
+      - 'analysis/**'
+      - 'requirements*.txt'
+      - '.github/workflows/python-test.yml'
+  pull_request:
+    paths:
+      - 'scripts/**'
+      - 'tests/**'
+      - 'corpus/INDEX.jsonl'
+      - 'analysis/**'
+      - 'requirements*.txt'
+      - '.github/workflows/python-test.yml'
+  schedule:
+    - cron: '0 8 * * 1'  # Weekly Monday 8am UTC (macOS only)
+
+jobs:
+  test-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          lfs: false
+          fetch-depth: 1
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: pip install -r requirements-lock.txt
+
+      - name: Run smoke tests
+        run: python -m pytest tests/ -v --tb=short
+
+  test-macos:
+    if: github.event_name == 'schedule'
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          lfs: false
+          fetch-depth: 1
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: pip install -r requirements-lock.txt
+
+      - name: Run smoke tests
+        run: python -m pytest tests/ -v --tb=short


ELares (commented):
Reviewed the latest diff. I see one substantive workflow bug before merge.

The PR body says macOS should run weekly on the schedule and Linux should run on pushes/PRs, but the current workflow does not actually implement that split. `test-macos` is correctly gated with `if: github.event_name == 'schedule'`, but `test-linux` is unconditional. Since the workflow itself is triggered by `schedule`, the Linux job will also run every Monday. So the current YAML implements `linux + macOS on schedule`, not `macOS only on schedule` as described.\n\nThe straightforward fix is to add the inverse gate to the Linux job, e.g. `if: github.event_name != 'schedule'`, or otherwise restructure the triggers so the scheduled event only targets the macOS job.\n\nWhat I verified locally:\n- `python -m pytest tests -q` still passes with `50` tests\n- the news-related tests are import-level tests from `scripts/`, so excluding `news/**` from the path filter is reasonable for this current suite\n\nSo the remaining problem is the schedule/job matrix, not the Python test command itself.

ELares (commented):
Re-reviewed on the latest commit. I do not see a remaining blocker from this pass.

What changed:
- the schedule/job-matrix bug is fixed
- `test-linux` is now explicitly gated with `if: github.event_name != 'schedule'`\n- `test-macos` remains schedule-only, so the workflow now matches the PR description: Linux on push/PR, macOS on weekly schedule only\n\nWhat I re-verified locally:\n- `python -m pytest tests -q` passes (`60 passed` on the current checkout)\n- `pytest` is present in `requirements-lock.txt`, so the workflow install path is still coherent\n\nSo from this pass, the workflow logic looks sound.

PR #147, opened 2026-04-24
Add archival release infrastructure for Zenodo DOI

### LINKED ISSUE #131: Add DOI-pinned archival release with frozen corpus manifest
## Source
Reviews 1, 3, and 4.

## Problem
GitHub is a mutable archive. The scripts query external APIs (PubMed, OpenAlex), and the corpus evolves over time. For scholarly reproducibility, the reported results need to be tied to a specific frozen state. Currently there is no DOI-pinned release, no checksums for the corpus snapshot, and no frozen manifest of the exact article set used.

Review 3: "for a serious scholarly version, the project would benefit from a DOI-pinned release, environment capture, seed control, checksums for the corpus snapshot, and a frozen manifest of the exact article set used."

## What Needs to Be Done
1. Create a Zenodo or Figshare DOI-pinned release that includes:
   - The exact `INDEX.jsonl` used for manuscript analyses
   - SHA256 checksums for all input data files
   - The exact `requirements.txt` and Rust toolchain version
   - All simulation output files referenced by manuscript figures
2. Add the DOI to the manuscript's data availability statement
3. Tag this as a "manuscript-v1-data" release on GitHub

### Owner Refinements
- **Cross-link hard to #124.** This issue depends on #124 (provenance and licensing). The release boundary cannot be defined until provenance and rights are resolved. Mark #124 as a prerequisite.
- **Define release boundary: code-only vs code+manuscript vs code+data.** Before creating the release, explicitly decide and document which of these release scopes applies:
  - **Code-only:** scripts, simulation source, toolchain files, tests
  - **Code + manuscript:** above plus LaTeX source, figures, drafts
  - **Code + data:** above plus INDEX.jsonl, tag indexes, simulation outputs
  - The full corpus (`corpus/by-pmid/`) almost certainly cannot be included if it contains non-redistributable content (paywalled articles).
- **Cannot archive non-redistributable content.** The DOI-pinned release must not include any content that cannot be legally redistributed (e.g., paywalled full-text PDFs, copyrighted textbook PDFs without CC license). The provenance work in #124 determines what is redistributable.

## Acceptance Criteria
- [ ] #124 is completed or its outputs are available to inform the release boundary
- [ ] Release boundary is explicitly documented (code-only, code+manuscript, or code+data)
- [ ] No non-redistributable content is included in the release
- [ ] DOI-pinned release created on Zenodo or Figshare with the defined scope
- [ ] SHA256 checksums included for all input data files in the release
- [ ] DOI is added to the manuscript's data availability statement
- [ ] GitHub release is tagged as "manuscript-v1-data"

## Priority
**Medium** — Not blocking current work, but essential before journal submission.

## Category
Infrastructure / Reproducibility

ELares:
## Summary
- **`.zenodo.json`**: Zenodo deposit metadata — dataset type, MIT license, description with explicit mixed-rights content classes
- **`scripts/generate_release_manifest.py`**: Generates SHA256 manifest and optional filtered archive excluding non-redistributable content

## Key design decisions

**Why manual Zenodo upload, not the GitHub webhook?**
GitHub's auto-generated release tarball includes ALL tracked files — that means all 4,830 full-text corpus articles (243 MB) including 61 whose OA status is unconfirmed. The manifest script builds a filtered ~12 MB archive with non-redistributable content excluded.

**Why no CITATION.cff or v1.md changes in this PR?**
The DOI doesn't exist until Zenodo assigns it. Putting a placeholder in CITATION.cff would render a fake DOI in GitHub's "Cite" button. Those updates happen in a follow-up commit after the real DOI is known.

**Mixed-rights deposit:**
The `.zenodo.json` description explicitly enumerates content classes: MIT code, author-owned manuscript, non-copyrightable metadata/abstracts. Not a single-license deposit.

## Release scope

| Component | Included | Rights basis |
|-----------|----------|-------------|
| scripts/, tests/, simulations/ (source) | Yes | MIT code |
| article/ (manuscript + figures) | Yes | Author-owned, MIT per LICENSE |
| analysis/ | Yes | Author-owned, MIT per LICENSE |
| corpus/INDEX.jsonl | Yes | Factual metadata (not copyrightable) |
| corpus/abstracts/ (5,584) | Yes | Non-copyrightable (Feist v. Rural) |
| tags/ | Yes | Generated indexes, MIT |
| corpus/by-pmid/ (4,830) | **No** | 61 non-OA articles |
| books/ | **No** | LFS (1.6 GB) |
| news/ (51) | **No** | Fair use ≠ redistributable license |

## Post-merge release process

```bash
# 1-2. Generate manifest + archive
python scripts/generate_release_manifest.py --build-archive

# 3. Verify exclusions
tar tzf release.tar.gz | grep -E 'corpus/by-pmid/|books/|news/' | wc -l  # 0

# 4-5. Commit, tag, push
git add MANIFEST.sha256
git commit -m "Add frozen release manifest (manuscript-v1-data)"
git tag -a manuscript-v1-data -m "Archival data snapshot for manuscript v1"
git push origin manuscript-v1-data

# 6. GitHub release
gh release create manuscript-v1-data release.tar.gz \
  --title "Manuscript v1 Data Snapshot" \
  --notes "Filtered archive. See MANIFEST.sha256 for checksums."

# 7. Zenodo: manual upload → get DOI
# 8. Follow-up commit: real DOI in CITATION.cff + v1.md
```

**Prerequisite:** Zenodo account (free, zenodo.org, GitHub OAuth).

## Test plan
- [x] `python3 -c "import json; json.load(open('.zenodo.json'))"` — valid JSON
- [x] Manifest excludes corpus/by-pmid/, books/, news/, MANIFEST.sha256, release.tar.gz
- [x] Archive builds without self-referential files
- [x] Script warns about 61 non-OA articles
- [x] `python3 -m pytest tests/ -q` — 60 passed
- [x] No changes to CITATION.cff or v1.md

Closes #131


ELares (commented):
Reviewed the latest diff. I see one substantive release-scope problem before merge.

The new release machinery is trying to solve a rights problem, but `.zenodo.json` and `generate_release_manifest.py` still treat the deposit as if it were cleanly MIT-licensed while the filtered archive includes content classes that are not actually MIT code. Concretely:
- `.zenodo.json` sets `"license": "MIT"` for the whole Zenodo deposit
- `generate_release_manifest.py` excludes only `corpus/by-pmid/` and `books/`
- that means the archival tarball still includes `news/by-source/` (51 fetched article files) and `corpus/abstracts/by-pmid/` (5,584 abstract markdown files), plus manuscript/analysis assets

That is a metadata/rights mismatch. Even if the archive is narrower than the full repo, it is still not an all-MIT artifact, and the current PR text overstates how solved the redistribution problem is. Fair use is not the same thing as a redistributable license, and abstracts/news content are exactly the classes where Zenodo-style republication needs tighter justification than a repo-local note.

I would fix this in one of two ways before merge:
1. narrow the archive further so the Zenodo deposit is genuinely code + clearly redistributable assets only, or
2. keep the broader archive but stop advertising it as a single MIT-licensed dataset and make the deposit metadata/release scope explicitly mixed-rights with those content classes called out.

The manifest script itself is directionally useful; the blocker is that the current release boundary and license metadata still do not line up.

ELares (commented):
Reviewed the latest diff. The news exclusion and mixed-rights description are improved, but I still see one substantive bug before merge.

`generate_release_manifest.py` builds its file list from `git ls-files`, then opens `MANIFEST.sha256` for writing and iterates the tracked files to hash them. That is fine on the very first run when `MANIFEST.sha256` is not yet tracked. But your own post-merge process says to commit `MANIFEST.sha256` and tag the release. After that point, a rerun will include `MANIFEST.sha256` in `git ls-files`, and the script will try to hash the manifest while it is simultaneously truncating/writing it. That makes the manifest self-referential and non-deterministic.

So the current workflow breaks exactly at the point where the manifest becomes a tracked release artifact. The straightforward fix is to exclude `MANIFEST.sha256` (and probably `release.tar.gz`) from the generated file list explicitly. Without that, this is not a stable archival-manifest workflow yet.

One lower-severity note: the PR body is now stale relative to the diff because it still says the filtered archive excludes only `corpus/by-pmid/` and `books/`, while the script now also excludes `news/`.

ELares (commented):
Re-reviewed on the latest commit. The manifest self-inclusion bug is fixed: excluding `MANIFEST.sha256` and `release.tar.gz` from the hashed file list is the right change.

I still see one substantive issue before merge, and it is the same underlying rights-metadata mismatch in a narrower form: `.zenodo.json` still sets `"license": "MIT"` for the Zenodo deposit even though the archive is explicitly described as mixed-rights. The description now says the deposit includes:
- MIT code
- author-owned manuscript/analysis
- non-copyrightable metadata/abstracts

That is better than before, but it still is not a single MIT-licensed dataset in the ordinary sense. If Zenodo only permits one top-level license field, then this PR should either:
1. narrow the release contents until `MIT` is actually accurate for the deposited artifact, or
2. stop claiming this closes the archival-release issue and instead frame it as release tooling pending a final licensing decision for the deposit metadata.

Right now the script is more stable, but the `.zenodo.json` license field still overstates how uniform the rights basis is.

ELares (commented):
Re-reviewed on the latest commit. The blocking issues from the prior passes are fixed from my side: the manifest is no longer self-referential, and `.zenodo.json` no longer falsely declares the whole deposit as `MIT`.

I do not see a remaining merge blocker in the source files. One low-severity metadata issue remains: the PR body is now stale relative to the diff. It still says the PR adds `MIT license` to `.zenodo.json`, but the file now uses a `license_note` explaining that the deposit is mixed-rights and that the final Zenodo license choice should be made at upload time after review.

So from this pass, the release tooling itself looks acceptable. The only cleanup left is making the PR summary match the current metadata strategy.

PR #163, opened 2026-04-25
Convert citations to superscript page footnotes for book-style reading

ELares:
## Summary
Transform the manuscript from academic citation style to book-style page footnotes. All inline citations (`[PMID: XXXXX]`, `(Author et al., Year)`, `[News: ...]`, textbook references) now use Markdown footnotes `[^label]` converted to LaTeX `\footnote{}` for superscript numbers with full references at the bottom of each page.

## Before → After

**Before:** "lethal in approximately 88% of patients [PMID: 37165196], a statistic..."
**After:** "lethal in approximately 88% of patients¹, a statistic..."
_¹ Rojas LA et al. Personalized RNA neoantigen vaccines. Nature (2023). PMID: 37165196._

## What changed
- **v1.md**: 104 footnote definitions (91 automated via conversion script + 13 manual author-year/textbook)
- **generate_latex.py**: Replaced `[PMID:]→\cite{}` pipeline with `[^label]→\footnote{}`. Removed natbib/bibtex dependency.
- **AUTHORING.md**: Updated citation standards for footnote convention
- **convert_citations_to_footnotes.py**: NEW — automated PMID/News conversion tool

## Pipeline change
| Before | After |
|--------|-------|
| `[PMID: X]` → `CITEPLACEHOLDER` → `\cite{key}` + bibtex | `[^label]` → `\footnote{text}` (self-contained) |
| Required bibliography.bib + natbib + bibtex pass | No external dependencies — native LaTeX `\footnote{}` |

## Test plan
- [x] `python3 -m pytest tests/ -q` — 79 passed
- [x] `grep -c '\[PMID:' v1.md` → 0 (all converted)
- [x] `grep -c '\[News:' v1.md` → 0 (all converted)
- [x] `grep -c 'footnote{' v1.tex` → 70 (page footnotes)
- [x] Reference appendix preserved at end of v1.md

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #165, opened 2026-04-26
Remove bibtex from release-pdf.yml CI pipeline

ELares:
## Summary
PR #163 switched citations from `\cite{}`/bibtex to inline `\footnote{}`. The release-pdf.yml workflow still runs `bibtex v1`, which fails with exit code 2 because there are no `\citation`, `\bibdata`, or `\bibstyle` commands in the aux file.

**Fix:** 2-pass pdflatex (cross-refs + TOC), no bibtex step.

Fixes the failed action: https://github.com/ELares/cancer_research/actions/runs/24943779352

## Test plan
- [ ] CI action succeeds on merge (pdflatex 2-pass produces PDF)

### DIFF
--- a/.github/workflows/release-pdf.yml
+++ b/.github/workflows/release-pdf.yml
@@ -41,11 +41,9 @@ jobs:
       - name: Generate LaTeX from Markdown
         run: python3 scripts/generate_latex.py
 
-      - name: Compile PDF (pdflatex + bibtex, 3-pass)
+      - name: Compile PDF (pdflatex, 2-pass)
         working-directory: article/drafts
         run: |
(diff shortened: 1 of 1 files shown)


ELares (commented):

The workflow change matches the current manuscript pipeline after the footnote conversion:
- `python3 scripts/generate_latex.py` succeeds
- two `pdflatex` passes complete successfully
- the manuscript instructions now consistently describe the no-bibtex build path
- `python3 -m pytest tests/ -q` still passes with 79 tests

So removing the `bibtex` step from `release-pdf.yml` is the correct CI fix for the current source of truth.

PR #168, opened 2026-04-26
Add dependabot.yml to fix Cargo ecosystem discovery

ELares:
## Summary
Dependabot failed with "No Cargo.toml!" because the workspace is in `simulations/`, not the repo root.

Adds `.github/dependabot.yml` configuring three ecosystems:
- **Cargo:** `/simulations` (Rust workspace)
- **pip:** `/` (requirements.txt)
- **GitHub Actions:** `/` (workflow files)

Weekly schedule for all three.

Fixes: https://github.com/ELares/cancer_research/actions/runs/24943981472

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/simulations"
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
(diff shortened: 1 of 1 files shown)


ELares (commented):

What I verified:
- the new `.github/dependabot.yml` uses the correct Rust workspace root: `/simulations`
- the pip updater target `/` is consistent with the current repo layout, which has `requirements.txt` and `requirements-lock.txt` at the top level
- the GitHub Actions updater target `/` is also correct for the current workflow files under `.github/workflows/`

So this looks like the right fix for the Cargo discovery failure, and the ecosystem directories are aligned with the repo structure.

PR #167, opened 2026-04-26
Upgrade pyo3 0.23 → 0.24 to fix buffer overflow vulnerability

ELares:
## Summary
Fixes Dependabot alerts #1 and #3: pyo3 < 0.24.1 has a risk of buffer overflow in `PyString::from_object` (low severity).

**Change:** `simulations/ferroptosis-python/Cargo.toml`: pyo3 `0.23` → `0.24`

## Test plan
- [x] `maturin develop -m ferroptosis-python/Cargo.toml --release` — builds
- [x] Python bindings work: `fc.sim_batch('Persister', 'SDT', n=100, seed=42)` returns correct result
- [x] `cargo test --workspace` — 31 passed
- [x] `python3 -m pytest tests/ -q` — 79 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #182, opened 2026-04-26
Update Cargo.lock to resolve Dependabot alert #1 (pyo3)

ELares:
## Summary
PR #167 updated `Cargo.toml` to pyo3 0.24 but didn't commit the regenerated `Cargo.lock`. Dependabot alert #1 remains open because the lockfile on main still references the vulnerable pyo3 version.

This commits the updated `Cargo.lock` with pyo3 0.24.2.

Fixes https://github.com/ELares/cancer_research/security/dependabot/1

### DIFF
--- a/simulations/Cargo.lock
+++ b/simulations/Cargo.lock
@@ -498,9 +498,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.23.5"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


ELares (commented):

What I verified:
- the diff is appropriately scoped to `simulations/Cargo.lock`
- the updated lockfile resolves `pyo3` and its related crates to `0.24.2` rather than the older `0.23.5` line
- `cd simulations && cargo test --workspace -q` still passes with `31` Rust tests

So this looks like the right lockfile follow-up to the earlier `Cargo.toml` bump, and it is the piece that actually closes the Dependabot alert on the committed dependency graph.

PR #183, opened 2026-04-26
Add security policy (SECURITY.md)

ELares:
## Summary
Adds `SECURITY.md` which GitHub renders at https://github.com/ELares/cancer_research/security/policy

Covers:
- **Supported versions:** Python pipeline, ferroptosis-core, Python bindings
- **Reporting:** Private via GitHub Security Advisories (not public issues)
- **Response timeline:** 72h acknowledgment, 1 week assessment
- **Scope:** API key handling, Rust memory safety, CI workflow permissions, data parsing
- **Out of scope:** Upstream dependency vulns, scientific correctness, bundled textbooks
- **Current practices:** Gitignored keys, least-privilege workflows, Dependabot, safe Rust

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -21,6 +21,9 @@ on:
   schedule:
     - cron: '0 8 * * 1'  # Weekly Monday 8am UTC (macOS only)
 
+permissions:
+  contents: read
+
 jobs:
   test-linux:
     if: github.event_name != 'schedule'
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,47 @@
+# Security Policy
+
+## Supported Versions
+
+| Component | Version | Supported |
+|-----------|---------|-----------|
+| Python pipeline (scripts/) | latest main | ✅ |
+| ferroptosis-core (Rust) | 0.2.0 | ✅ |
+| ferroptosis-python bindings | 0.1.0 | ✅ |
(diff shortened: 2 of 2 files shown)


ELares (commented):
Reviewed this pass. I found one substantive accuracy issue in the policy text.

`SECURITY.md` currently says: `GitHub Actions workflows use explicit permissions blocks with least privilege.` That is not true for the current repo state. I verified that some workflows do have explicit `permissions:` blocks, for example `cargo-test.yml`, `pypi-publish.yml`, and `release-pdf.yml`, but `python-test.yml` does not currently declare one.

Why this matters: for a security policy, this is not just wording drift. It states a present-tense hardening control that the repo has not applied consistently yet.

- add an explicit least-privilege `permissions:` block to `python-test.yml`, or
- narrow the `SECURITY.md` wording so it describes the current state accurately instead of claiming repo-wide consistency

ELares (commented):
Re-reviewed on the latest commit. The earlier policy-accuracy issue is resolved.

What I verified:
- `python-test.yml` now also declares an explicit least-privilege `permissions:` block (`contents: read`)
- all current workflow files under `.github/workflows/` now have explicit `permissions:` declarations
- that means the corresponding sentence in `SECURITY.md` is now true for the current repo state, rather than overstating the hardening posture

So from this pass, I do not see a remaining blocker.

Issue #209, opened 2026-05-10
Workspace-wide cargo fmt drift (157 diffs); add fmt-check to CI


ELares:
## Context

A reviewer of #208 ran `cargo fmt --all --check` and reported broad formatting drift across the simulation workspace. Verified locally on `main` (commit \`89ee43c6\`):

\`\`\`
$ cargo fmt --manifest-path simulations/Cargo.toml --all --check 2>&1 | grep -c '^Diff'
157
\`\`\`

CI does not currently run \`cargo fmt --check\` (no rustfmt invocation in any \`.github/workflows/*.yml\`), so drift has accumulated naturally over PRs.

## What's stale

157 line-level diffs across at least:
- \`ferroptosis-core/examples/basic_usage.rs\`
- \`ferroptosis-core/src/biochem.rs\`
- \`ferroptosis-core/src/cell.rs\`
- ... and likely most workspace files

None of the diffs are correctness issues — they're conventions like sorted imports, line-length rewraps, and consistent block placement. But the longer this drifts, the noisier diffs of unrelated PRs get.

## Proposed work

Two coordinated changes in one PR:

1. **One-shot `cargo fmt --all` mass-formatting commit** — touches a lot of files but resets the baseline.
2. **CI gate** — add a job to `.github/workflows/cargo-test.yml` (or a new `cargo-fmt.yml`) that runs `cargo fmt --manifest-path simulations/Cargo.toml --all --check`, blocking PRs with new drift.

Estimated effort: ~30 min for the one-shot fmt; ~10 min for the CI gate. Probably batch-able with another small PR.

## Acceptance criteria

- [ ] `cargo fmt --manifest-path simulations/Cargo.toml --all --check` exits 0 on `main`
- [ ] CI runs fmt-check on PRs touching `simulations/`
- [ ] No behavior change (formatting only); workspace tests still green; sim-spatial 100×100 default snapshot still byte-identical (compiler is deterministic, but worth verifying)

## Out of scope

- `cargo clippy` warnings (currently 4 pre-existing in `io.rs`/`immune.rs`/`cell.rs`) — separate concern; could be its own follow-up.
- Python formatter (\`ruff\`/\`black\`) — separate stack.

## Source

Reviewer comment on PR #208: \"`cargo fmt --all --check`; it fails with broad existing formatting drift across the workspace.\"

Issue #194, opened 2026-04-26
sim-spatial-3d: New binary for 3D tumor spheroid simulations


ELares:
## Context
Once the 3D grid infrastructure and physics modules are in place, a new binary integrates them into a runnable 3D spatial simulation — the 3D equivalent of sim-spatial.

## What Needs to Be Done
1. New workspace member: `simulations/sim-spatial-3d/`
2. CLI interface:
   - \`--grid-size\` (default 100 — meaning 100³ = 1M cells)
   - \`--cell-size\` (default 20μm)
   - \`--seed\` (default 42)
   - \`--output-dir\` (default \"output/spatial-3d\")
   - \`--n-steps\` (default 180)
3. Core simulation loop: 3D grid construction → energy deposition → 180-step ferroptosis with iron diffusion
4. Output: 3D death maps, radial depth-kill curves, zone-level kill rates
5. Comparison mode: run with equivalent 2D parameters and compare results

## Key Predictions to Test
- Does the PDT penetration limit (~1mm) produce a MORE dramatic kill dropoff in 3D (spheroid interior is farther from all surfaces)?
- Does SDT maintain its depth advantage in 3D?
- Does the overall RSL3 kill rate change when cells are distributed in 3D vs 2D?

## Acceptance Criteria
- [ ] New binary compiles and runs
- [ ] Produces 3D output files
- [ ] Radial depth-kill curves comparable to 2D results
- [ ] README.md with parameters, output format, and manuscript claim reproduction
- [ ] \`cargo test --workspace\` passes

## Dependencies
- #185 (3D Grid Infrastructure)
- #186 (3D Energy Physics)
- #192 (3D Performance)
- #193 (3D Visualization)

ELares:
Review pass: this is a solid integration issue. The main thing I would tighten is the dependency boundary.

I would not make the new binary depend on the visualization issue to compile and run. Basic numeric outputs are enough for first integration; richer 3D visualization can follow. Otherwise the runnable 3D binary gets blocked by output-format work that is not actually required to test the physics.

I would also treat `comparison mode` as a useful follow-up, not necessarily part of the minimum compile-and-run acceptance surface.

ELares:
## Rigorous Planning Analysis

### Problem restated
3D equivalent of sim-spatial: spatial ferroptosis with depth-dependent energy physics in a spheroid.

### Key predictions to test
[...]

PR #213, opened 2026-05-10
Add 3D energy physics for spheroid PDT/SDT (#186)

### LINKED ISSUE #186: 3D Energy Physics: Beer-Lambert and acoustic attenuation in 3D
## Context
`ferroptosis-core/src/physics.rs` computes depth-dependent energy deposition for PDT (Beer-Lambert) and SDT (acoustic attenuation). Currently depth is row-based (1D distance from grid edge). In 3D, depth must be the true distance from the spheroid surface to each cell's position.

## What Needs to Be Done
1. Compute 3D radial depth: `depth = tumor_radius - distance_from_center` for each cell in the 3D grid
2. Update `local_ros_multiplier()` to accept 3D depth
3. PDT: Beer-Lambert attenuation `I(d) = I₀ × exp(-μ_eff × d)` — same formula, 3D depth input
4. SDT: Acoustic attenuation `I(d) = I₀ × 10^(-α×f×d/10)` — same formula, 3D depth input
5. Consider directional energy delivery: ultrasound beam has a focal zone, not uniform from all surfaces. Optional for v1.

## Key Physics
- PDT effective penetration: δ ≈ 3.2mm at 630nm (unchanged)
- SDT attenuation: ~0.7 dB/cm/MHz in soft tissue (unchanged)
- 3D changes WHERE the depth is measured from, not HOW attenuation works

## Acceptance Criteria
- [ ] `local_ros_multiplier()` works with 3D radial depth
- [ ] PDT kill dropoff matches 2D behavior at equivalent depths
- [ ] SDT maintains deep-tissue penetration in 3D
- [ ] Unit tests comparing 2D and 3D at matched depths

## Dependencies

ELares:
## Summary

Closes #186. Adds the physics primitives the 3D ferroptosis pipeline needs without touching the 2D path.

- `TumorGrid3D::radial_depth_um(r, c, l) -> f64` — signed depth from the spheroid surface (positive inside, negative outside, zero at surface). Geometry matches `TumorGrid3D::generate`.
- `physics::local_ros_multiplier_3d(depth_um, tx, params) -> f64` — 3D analog of `local_ros_multiplier`. Takes raw depth, clips negative values to zero, dispatches to the same depth-attenuation functions the 2D path uses (Beer-Lambert for PDT, acoustic for SDT, uniform for RSL3, zero for Control).
- Bumps `ferroptosis-core` 0.6.0 → 0.7.0 (additive feature batch).

## Design notes

The 2D and 3D signatures are intentionally asymmetric. 2D `local_ros_multiplier(row, cell_size, ...)` computes `z = row × cell_size` internally because energy enters from a flat surface (row 0). 3D `local_ros_multiplier_3d(depth_um, ...)` takes raw `z` because per-cell depth in a spheroid depends on geometry (distance to sphere surface), not just a single coordinate × size. Computing depth lives on `TumorGrid3D`; physics stays decoupled from grid representation. Documented on the function.

The **2D ≡ 3D matched-depth invariant** is locked down by `local_ros_multiplier_2d_3d_match_at_same_depth`: iterates all four `Treatment` variants over five depths and asserts bit-exact equality. Any future refactor that introduces 3D-only physics quirks would fail there.

## Acceptance criteria (#186)

- [x] `local_ros_multiplier()` works with 3D radial depth — via new `local_ros_multiplier_3d`, keeping the 2D function untouched (verified bit-identical via sim-spatial snapshot below).
- [x] PDT kill dropoff matches 2D behavior at equivalent depths — covered by `local_ros_multiplier_2d_3d_match_at_same_depth`.
- [x] SDT maintains deep-tissue penetration in 3D — same matched-depth test covers SDT.
- [x] Unit tests comparing 2D and 3D at matched depths — see above.

Per issue: directional energy delivery (ultrasound focal zone vs. isotropic-surface) is explicitly listed as **optional for v1** and is not addressed here.

## Verification

- 13 `grid::tests_3d` tests pass (4 new for `radial_depth_um`: at center, outside-sphere, anisotropic grid, determinism)
- 12 `physics::tests` tests pass (3 new: matched-depth invariant, negative-depth clipping, photosensitizer-composes-in-3D)
- 103 total `ferroptosis-core` tests pass
- `sim-spatial 100×100` snapshot bit-identical before vs. after:
  - `spatial_summary.json` SHA-256: `f7628f85fe86217142f078d2857caefb6c4ac400bf2837abf3c96c2808e13e66`
  - `depth_kill_curves.csv` SHA-256: `2eea07e1ef94ff5233f470c434183c4ad413593ed34d5dacef9fdbaa0059a72e`

## Test plan

- [x] `cargo test --workspace --release` passes locally
- [x] `cargo build --release --workspace --exclude ferroptosis-python` clean (Python crate excluded due to local pyo3 linker env; unaffected by these changes)
- [x] `sim-spatial 100×100` re-run produces bit-identical output to baseline
- [ ] CI green

## Follow-ups

- Partially unblocks #194 (sim-spatial-3d). The binary still needs analytics helpers (radial-depth curves, volumetric heatmaps) and the binary itself.
- Does not address #209 (pre-existing fmt drift + CI gate); new code in this PR is fmt-clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - pathway-target and resistant-state analysis
 - diagnostic-to-therapy chain extraction (6 chains, 129 articles mapped)
 - tissue-of-origin analysis layer (5 tissue categories, 62% coverage)
-- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial)
+- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186) for the #185–#197 spheroid-validation series
 - ferroptosis-core library packaging for external use
 - news source authentication pipeline (fetch, extract claims, verify, score, index)
 - broader strategy review of alternative therapies and biological bottlenecks
@@ -51,7 +51,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - tissue-of-origin and weighted-evidence layers are active
 - diagnostic-therapy matching layer covers 6 chains across 4 modalities (radioligand, checkpoint, mRNA vaccine, oncolytic)
 - manuscript: 112 pages (book format), 11 chapters + 3 appendices, 20 figures, ~36,700 words
-- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 11 modules including `photosensitizer_pk`; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
+- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 11 modules including `photosensitizer_pk`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
 - news authentication pipeline: 5 scripts (fetch, extract claims, verify against PubMed, score credibility, build claim-centric index) implementing the 3-tier source framework from analysis/news-source-criteria.md
 - simulation calibration: 5 targets documented, evaluate script operational
 - drug penetration module: 3 tissue types, exponential Krogh approximation
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ If you have expertise in oncology, biochemistry, ferroptosis, immunology, comput
 
 - **4,830 full-text cancer research articles** across 19 mechanisms, 22 cancer types, 803 journals (2001-2026)
 - **Python pipeline** for corpus fetching, tagging (7 tag layers), indexing, analysis, and figure generation
-- **10 Rust simulation binaries** modeling ferroptosis biochemistry: single-cell Monte Carlo, spatial tumors with PDT/SDT depth physics + photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield), drug penetration, drug combinations, tumor microenvironment (oxygen gradients, spatial immune zones, DAMP-mediated T cell activation), vulnerability windows, ICD immune cascades, tumor PK
+- **10 Rust simulation binaries** modeling ferroptosis biochemistry: single-cell Monte Carlo, spatial tumors with PDT/SDT depth physics (2D row-based and 3D radial-depth dispatchers — 3D spheroid scaffolding for the upcoming spheroid-validation series) + photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield), drug penetration, drug combinations, tumor microenvironment (oxygen gradients, spatial immune zones, DAMP-mediated T cell activation), vulnerability windows, ICD immune cascades, tumor PK
 - **ferroptosis-core library** (MIT, with Python bindings) — embeddable ferroptosis biochemistry engine; module list and current unit-test count in [`simulations/ferroptosis-core/README.md`](simulations/ferroptosis-core/README.md)
 - **Calibration infrastructure** linking simulation parameters to published experimental data
 - **112-page book-format manuscript** with 11 chapters, 3 appendices, and 20 figures (~36,700 words), cross-referenced against all analysis outputs
--- a/simulations/Cargo.lock
+++ b/simulations/Cargo.lock
@@ -217,7 +217,7 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "ferroptosis-core"
-version = "0.6.0"
+version = "0.7.0"
 dependencies = [
  "csv",
  "ndarray",
--- a/simulations/README.md
+++ b/simulations/README.md
@@ -21,7 +21,7 @@ Each binary has its own README with parameters, output format, example commands,
 
 ## Shared library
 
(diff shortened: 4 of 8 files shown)


ELares:
# Code Review

## Overview

Adds two physics primitives needed by the spheroid-validation series (#185–#197) without touching the established 2D path:
- `TumorGrid3D::radial_depth_um(r, c, l) → f64` (signed depth from spheroid surface)
- `physics::local_ros_multiplier_3d(depth_um, tx, params) → f64` (3D dispatcher reusing the existing depth-attenuation functions)

Bumps `ferroptosis-core` 0.6.0 → 0.7.0 (additive). The second commit (cd982a4d) extracts `TUMOR_RADIUS_FRACTION = 0.45` as a shared constant between `generate` and `radial_depth_um`, adds physical caveats, and refreshes docs.

## Strengths

- **Bit-exact matched-depth invariant** (`local_ros_multiplier_2d_3d_match_at_same_depth`) iterates all four `Treatment` variants × 5 depths and asserts `assert_eq!` (not approx). Locks the contract that 3D is *not* a separate physics implementation, just a different per-cell depth derivation.
- **Geometry-drift guard** (`radial_depth_sign_agrees_with_generated_is_tumor`) walks every cell of a 10³ grid and asserts the bi-conditional `is_tumor ↔ depth ≥ 0`. Catches the exact silent-divergence class (`generate` and `radial_depth_um` referencing different radii) that motivated the `TUMOR_RADIUS_FRACTION` extraction.
  - Verified the bi-conditional is correct against `grid.rs:443` (`is_tumor = !(dist > tumor_radius)`), including the boundary `dist == tumor_radius` (both `is_tumor` and `depth = 0` agree).
- **Honest physics caveats** in `local_ros_multiplier_3d`, `radial_depth_um`, and the module doc explicitly call out: nearest-surface 1-D approximation, no chord-length integration, no diffuse back-scatter — and label the 2D≡3D invariant as mathematical (dispatcher contract), not physical. Aligns with the CLAUDE.md "Be honest about what we don't know" principle.
- **NaN-safe**: `radial_depth_um.max(0.0)` returns 0.0 for NaN input (Rust's `f64::max` returns the non-NaN argument), so a NaN depth clips to surface rather than propagating.
- **Doc sweep is in scope** — README + CLAUDE.md + module-level rustdoc all updated together.

## Concerns / Suggestions

**Minor (doc accuracy, not behavior)**

- `radial_depth_at_center_is_max` claims "Hand-computed in pure-rational arithmetic so the assertion is IEEE-exact." The literal `0.45` is *not* exact in IEEE 754 — it's a repeating binary fraction. The test does pass deterministically because `20.0 * 0.45` happens to round to exactly `9.0` (fraction above is < half-ULP at 9.0), but the "pure-rational" framing is wrong. Either reword to "rounding-friendly" or drop the claim. The `radial_depth_at_sphere_surface_is_zero` test has the same wording: it's actually exact there only because `sqrt(81.0) == 9.0` is IEEE-required-correct.
- Inconsistent return-range doc between `local_ros_multiplier` ("multiplier in [0, 1]", physics.rs:122) and `local_ros_multiplier_3d` ("[0, ∞)"). The 2D doc is the stale one — `Uniform(c > 1.0)` and `phi_so2_relative > 1.0` are intentional escape hatches that can push the 2D output above 1.0 too. Out of scope to fix here, but worth flagging.

**Minor (perf, future-facing)**

- `radial_depth_um` recomputes `center_r/c/l` and `tumor_radius` on every call. For the eventual sim-spatial-3d inner loop over ~10⁵–10⁶ cells, hoisting these to a precomputed snapshot would shave work. `#[inline]` helps single call sites but won't hoist across loop iterations without LICM seeing the full code. Not a blocker for v1; flag for #194.

**Minor (API asymmetry, already justified in docs)**

- The 2D `local_ros_multiplier(row, cell_size, tx, params)` vs. 3D `local_ros_multiplier_3d(depth_um, tx, params)` signature split is intentional and well-documented. One alternative worth noting: factoring out an internal `local_ros_multiplier_inner(z_um, tx, params)` shared by both wrappers would eliminate duplicated `match tx { … }` blocks and make "same dispatcher, two derivations" a structural truth (not just a tested one). The current shape works; the test guards drift; mention is informational.

**Out of scope but worth knowing**

- The 2D `TumorGrid::generate` still uses `0.45` as an inline literal. The PR documents this and chooses not to unify. Fine — the bi-conditional test only guards 3D drift.

## Correctness Spot-Checks

| Test | Verification |
|---|---|
| `radial_depth_at_center_is_max` | `tumor_radius = 20×0.45 → 9.0` (exact via rounding); `(9.0 - 0.0)×20.0 = 180.0` ✓ |
| `radial_depth_at_sphere_surface_is_zero` | dist² = 81, sqrt → 9.0 exact; `(9.0-9.0)×20.0 = 0.0` ✓ |
| `radial_depth_anisotropic_grid` | tumor_radius = 5×0.45 = 2.25 exact; dist=0.5; `(2.25-0.5)×10.0 = 17.5` ✓ |
| Negative-depth clipping | `(-z).max(0.0) = 0.0`; `exp(0) = 1.0`, `10^0 = 1.0` IEEE-exact; with default `pdt_i0 = sdt_i0 = 1.0` and `Uniform(1.0)`, multiplier = 1.0 ✓ |

## Verdict

**Approve with minor doc nits.** Strong test coverage, honest physics framing, clean scope discipline (no 2D path changes, additive semver bump, bit-identical sim-spatial snapshot). The IEEE-exact wording in two test docstrings is the only thing worth tweaking — it's documentation accuracy, not behavior. The perf hint on `radial_depth_um` recomputation is forward-looking for #194 and not a blocker.

ELares:
I did not identify any discrete, actionable bugs introduced by this patch. The added 3D depth/physics APIs align with the existing geometry and tests pass locally.

ELares:
Thanks for the review. Addressed in 91e6bba1:

**Fixed (in this PR):**
- ✅ **IEEE-exact wording** on `radial_depth_at_center_is_max` and `radial_depth_at_sphere_surface_is_zero` — reworded to explain *why* strict equality is safe (`20.0 × 0.45` rounds to exactly `9.0` under IEEE round-to-nearest; `sqrt(81.0) = 9.0` is IEEE-required-correct for perfect squares) without claiming the arithmetic is provably exact rationally.
- ✅ **Perf note** added to `radial_depth_um` rustdoc flagging the per-call recomputation of `center_{r,c,l}` and `tumor_radius`. Tagged for #194 — sim-spatial-3d's inner loop should precompute outside the loop or cache on the struct.

**Filed as follow-up:**
- 📋 **#214** — 2D `local_ros_multiplier` return-range docstring (`[0, 1]` stale; `Uniform(c>1.0)` / `phi_so2_relative>1.0` can push above 1.0). You flagged this as out-of-scope for this PR; filed so it doesn't get lost.

**Acknowledged, not changing:**
- The factoring suggestion (`local_ros_multiplier_inner` shared between 2D/3D wrappers) — current shape works, matched-depth invariant test guards drift, and as you noted it's informational rather than a blocker. Open to revisiting if a third dispatcher ever lands.
- 2D `TumorGrid::generate` still using `0.45` inline — pre-existing, intentionally out of scope (the sign-agreement test only guards the 3D pair).

Verification: 105 ferroptosis-core tests still pass; sim-spatial 100×100 SHA-256 still bit-identical to baseline (unchanged from the prior verification — only test-docstring text and a rustdoc comment changed in this commit).

Issue #191, opened 2026-04-26
3D Vasculature Network: Replace edge-distance proxy with explicit vessel model


ELares:
## Context
All 2D simulations use \"distance from tumor edge\" as a proxy for vasculature proximity. This works in 2D because the edge is the only oxygen/drug source. In 3D, tumors have internal vasculature — blood vessels penetrate the tumor volume, creating complex O2/drug supply patterns. A 3D model needs either an explicit vascular network or a statistical vessel distribution.

## What Needs to Be Done
1. **Option A (minimal):** Random vessel seed points within the 3D tumor volume. Each cell's O2/drug supply is determined by distance to nearest vessel. Simple to implement, captures the key physics.
2. **Option B (intermediate):** Fractal-branching vascular tree from the periphery inward. More realistic but significantly more complex.
3. **Option C (external):** Import vascular networks from published micro-CT datasets of tumor vasculature.

**Recommendation:** Start with Option A for v1. The key insight (irregular oxygenation creates heterogeneous treatment response) is captured with random vessel placement.

## Biology
- Tumor vasculature is chaotic: irregular branching, dead ends, shunts, varying diameters
- Average inter-vessel distance: 100-200μm in well-vascularized tumors, 300-500μm in poorly vascularized
- The Krogh cylinder model (used in sim-tissue-pk) already handles radial diffusion from a single vessel — extend to multiple vessels in 3D

## Acceptance Criteria
- [ ] 3D vascular network model (at minimum: random seed points)
- [ ] Per-cell distance-to-nearest-vessel computed
- [ ] O2 and drug concentration derived from vessel proximity
- [ ] Comparison: edge-distance-proxy vs explicit-vessel O2 fields

## Dependencies
- #185 (3D Grid Infrastructure)
- #187 (3D O2 Gradient)

ELares:
Review pass: this is the one issue here that feels most at risk of over-scoping.

Option A/B/C are not really three equal implementation branches. Option A is a tractable v1. Option B is a larger research project. Option C is a data-ingestion problem. I would lock the issue to Option A for the first pass and treat the others as explicit future extensions.

I would also be careful not to let this become a hard prerequisite for every 3D result. The radial-from-surface spheroid model still has value as an intermediate validation step even before explicit internal vasculature exists.

ELares:
## Rigorous Planning Analysis

### Problem restated
[...]

Issue #188, opened 2026-04-26
3D Immune Coupling: DAMP diffusion and T cell activation in 3D


ELares:
## Context
The immune coupling model diffuses DAMPs on a 2D grid and activates resident T cells based on local DAMP concentration. In 3D, DAMP diffusion follows different dynamics (surface-to-volume ratio changes, diffusion gradients are steeper in the interior).

## What Needs to Be Done
1. Extend DAMP diffusion field from 2D array to 3D volume
2. Update diffusion kernel from 8-neighbor to 26-neighbor with appropriate coefficients
3. DC activation and T cell priming in 3D neighborhood
4. Immune kill iteration over 26-neighbor volume
5. PD-1/anti-PD-1 brake unchanged (scalar, not spatial)

## Key Question
The 2D 104:1 immune kill ratio (SDT vs RSL3) is driven by kill DENSITY. In 3D, the same number of kills are distributed across more volume, potentially REDUCING local DAMP concentration. The 3D model may show a LOWER immune differential than 2D — this would be an important finding about the dimensionality dependence of the immune coupling prediction.

## Acceptance Criteria
- [ ] DAMP field in 3D volume
- [ ] 26-neighbor diffusion with correct coefficients
- [ ] T cell activation in 3D
- [ ] Compare 2D vs 3D immune kill ratios at matched parameters
- [ ] Document whether the 104:1 ratio holds, shrinks, or grows in 3D

## Dependencies
- #185 (3D Grid Infrastructure)
- #187 (3D O2 Gradient — immune coupling depends on O2 for basal ROS)

ELares:
Review pass: this is a useful issue, but the acceptance criteria should be a bit more explicit about normalization.

The 2D `104:1` comparison is density-sensitive, so a 3D comparison can be misleading unless the issue states what is being held constant: total cells, physical radius, voxel size, or total killed cells. I would add that the 2D vs 3D immune-ratio comparison must be done under matched physical geometry assumptions, not just matched raw parameter names.

I would also treat `104:1 holds / shrinks / grows` as a hypothesis outcome, not a built-in expectation.

ELares:
## Rigorous Planning Analysis

### Problem restated
DAMP diffusion in 2D creates local concentrations that activate T cells. In 3D, the same DAMPs spread through more volume, potentially REDUCING local concentration below the activation threshold (Kd=50).

### Key scientific prediction
[...]

PR #215, opened 2026-05-10
Add 3D radial oxygen gradient for spheroid tumors (#187)

### LINKED ISSUE #187: 3D Oxygen Gradient: Radial O2 field for tumor spheroids
## Context
`sim-tme` computes O2 as `exp(-depth/λ)` where depth is distance from the 2D tumor edge. In 3D, oxygen diffuses radially inward from the spheroid surface, creating a well-characterized radial gradient with a necrotic core.

## What Needs to Be Done
1. Compute 3D radial O2 field: `O2(r) = exp(-r_from_surface / λ)` where `r_from_surface = tumor_radius - distance_from_center`
2. Preserve the O2 cycling toggle (square-wave λ alternation) in 3D
3. Define 3D zone boundaries: normoxic shell, transition zone, hypoxic core (concentric spherical shells)
4. Update `zone_kill_rates()` for spherical geometry
5. Validate: O2 profile in a 200μm spheroid should match published pimonidazole staining patterns

## Biology
- Tumor spheroids >200μm develop hypoxic cores (Vaupel, Cancer Research 1989)
- O2 penetration ~100-150μm from vasculature (same as 2D model)
- Spheroid geometry means the hypoxic VOLUME fraction is larger than the 2D hypoxic AREA fraction (cubic vs quadratic scaling)

## Acceptance Criteria
- [ ] Radial O2 gradient in 3D grid
- [ ] O2 cycling works in 3D
- [ ] Zone kill rates computed for spherical shells
- [ ] Comparison: 2D vs 3D hypoxic volume fractions at matched λ

## Dependencies

ELares:
## Summary

Closes #187. Adds the per-cell O₂ field and zone-based dead-cell census that downstream 3D consumers (#188 immune coupling, #191 vasculature, #197 cell-level biochem) need. Strictly additive — sim-tme and sim-spatial untouched, sim-spatial 100×100 SHA-256 bit-identical.

- New module `ferroptosis_core::oxygen` with two pure functions
- Bumps `ferroptosis-core` 0.7.0 → 0.8.0 (additive minor)
- 10 new tests; 115 total ferroptosis-core tests pass

## API

```rust
/// Per-cell O₂ multiplier in [0, 1]; length matches grid.cells.
/// Stromal cells (outside spheroid) return 1.0.
pub fn radial_o2_field(grid: &TumorGrid3D, lambda_um: f64) -> Vec<f64>;

/// (normoxic_rate, transition_rate, hypoxic_rate) for tumor cells.
/// Same depth thresholds as sim-tme's 2D zone_kill_rates.
pub fn radial_o2_zone_kill_rates(grid: &TumorGrid3D, shell_depth_um: f64) -> (f64, f64, f64);
```

## Design choices

1. **Pure functions, not mutation.** sim-tme's 2D `apply_o2_gradient` mutates `cell.basal_ros *= o2_factor`. These 3D analogs return values so the consumer chooses mutate-vs-snapshot, and O₂ cycling becomes "re-call with alternating λ per step" with no `original_ros` bookkeeping.
2. **First-order Krogh approximation.** `O2(d) = exp(-d/λ)` is the same approximation level as sim-tme's 2D code. The exact Krogh-Riley spheroidal solution involves `sinh` ratios; documented in the module rustdoc but out of scope.
3. **Stromal convention.** Cells outside the spheroid (`is_tumor == false`) return `O2 = 1.0` — bulk normal tissue, well-oxygenated. Same convention as 2D.

## Acceptance criteria (#187)

- [x] Radial O₂ gradient in 3D grid — `radial_o2_field`
- [x] O₂ cycling works in 3D — supported via per-step λ variation (see `distinct_lambdas_produce_distinct_fields` test)
- [x] Zone kill rates computed for spherical shells — `radial_o2_zone_kill_rates`
- [x] Comparison: 2D vs 3D hypoxic volume fractions at matched λ — `matched_lambda_2d_vs_3d_zone_fractions` test

## Interesting finding — issue text was inverted

The issue context claimed "the hypoxic VOLUME fraction is larger than the 2D hypoxic AREA fraction (cubic vs quadratic scaling)." Pure geometry says the opposite: hypoxic fraction = `((R-3λ)/R)^d` where `d=2` (area) or `d=3` (volume). Since `(R-3λ)/R ∈ (0, 1)`, raising to higher power gives a *smaller* fraction. So **3D hypoxic fraction < 2D hypoxic fraction** at matched R and λ.

Where cubic-vs-quadratic scaling DOES dominate: the *normoxic shell*. 3D surface area scales as R², 2D perimeter as R — so the 3D normoxic shell has more cells relative to total than the 2D normoxic ring. Both halves are asserted in the test.

The Vaupel 1989 "3D spheroids are more hypoxic" observation reflects vasculature *biology* (poor 3D O₂ supply effectively shrinks λ), not geometry. This pure-geometry model intentionally separates the two effects so a calibrated λ in #196 can capture the biology cleanly.

## Verification

- 115 ferroptosis-core tests pass (was 105; +10 new in `oxygen::tests`)
  - `spatial_summary.json`: `f7628f85fe86217142f078d2857caefb6c4ac400bf2837abf3c96c2808e13e66`
  - `depth_kill_curves.csv`: `2eea07e1ef94ff5233f470c434183c4ad413593ed34d5dacef9fdbaa0059a72e`
- `cargo fmt -- --check` clean on `oxygen.rs` (pre-existing drift elsewhere remains — #209)

## Test plan

- [x] `cargo test --release -p ferroptosis-core` passes locally (115 tests)
- [x] `cargo build --release --workspace --exclude ferroptosis-python` clean (Python crate excluded due to local pyo3 linker env)
- [ ] CI green

## Follow-ups

- **Unblocks**: #188 (3D immune coupling, blocked by #187 only), #191 (3D vasculature, blocked by #187 only), #197 (3D cell-level biochem, blocked by #187 only). Three more issues are now ready to tackle.
- **Out of scope**: Python/FFI bindings (intentional — matches #186 scope; #194 sim-spatial-3d will drive that decision)
- **Issue text fix suggested**: #187's context line "hypoxic VOLUME fraction is larger" should be updated to reference normoxic-shell scaling or biological vasculature effects. Not blocking — happy to make the edit in the issue close comment.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - pathway-target and resistant-state analysis
 - diagnostic-to-therapy chain extraction (6 chains, 129 articles mapped)
 - tissue-of-origin analysis layer (5 tissue categories, 62% coverage)
-- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186) for the #185–#197 spheroid-validation series
+- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186, 3D radial O₂ gradient + zone-census #187) for the #185–#197 spheroid-validation series
 - ferroptosis-core library packaging for external use
 - news source authentication pipeline (fetch, extract claims, verify, score, index)
 - broader strategy review of alternative therapies and biological bottlenecks
@@ -51,7 +51,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - tissue-of-origin and weighted-evidence layers are active
 - diagnostic-therapy matching layer covers 6 chains across 4 modalities (radioligand, checkpoint, mRNA vaccine, oncolytic)
 - manuscript: 112 pages (book format), 11 chapters + 3 appendices, 20 figures, ~36,700 words
-- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 11 modules including `photosensitizer_pk`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
+- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 12 modules including `photosensitizer_pk` and `oxygen`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; v0.8.0 adds 3D radial O₂ field + zone-census #187; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
 - news authentication pipeline: 5 scripts (fetch, extract claims, verify against PubMed, score credibility, build claim-centric index) implementing the 3-tier source framework from analysis/news-source-criteria.md
 - simulation calibration: 5 targets documented, evaluate script operational
 - drug penetration module: 3 tissue types, exponential Krogh approximation
--- a/simulations/Cargo.lock
+++ b/simulations/Cargo.lock
@@ -217,7 +217,7 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "ferroptosis-core"
-version = "0.7.0"
+version = "0.8.0"
 dependencies = [
  "csv",
  "ndarray",
--- a/simulations/README.md
+++ b/simulations/README.md
@@ -21,7 +21,7 @@ Each binary has its own README with parameters, output format, example commands,
 
 ## Shared library
 
-`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
+`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), 3D radial O₂ gradients (`oxygen::radial_o2_field` + zone-census, #187), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
 
 ## Quick start
 
--- a/simulations/ferroptosis-core/Cargo.toml
+++ b/simulations/ferroptosis-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ferroptosis-core"
-version = "0.7.0"
+version = "0.8.0"
 edition = "2021"
 description = "Embeddable ferroptosis biochemistry engine for cancer simulation"
 license = "MIT"
--- a/simulations/ferroptosis-core/README.md
+++ b/simulations/ferroptosis-core/README.md
@@ -34,6 +34,7 @@ Run the included example: `cargo run -p ferroptosis-core --example basic_usage`
 | `stats` | Wilson confidence intervals, parallel Monte Carlo execution via rayon |
 | `physics` | Depth-dependent energy deposition: Beer-Lambert (PDT), acoustic attenuation (SDT), uniform (RSL3). 2D row-based (`local_ros_multiplier`) and 3D radial-depth (`local_ros_multiplier_3d`) dispatchers share the same per-treatment depth functions (#186). |
 | `grid` | 2D `TumorGrid` (8-Moore, circular) and 3D `TumorGrid3D` (26-Moore, spherical) with heterogeneous architecture, neighbor iteration, iron diffusion. `TumorGrid3D::radial_depth_um` provides per-cell signed depth from the spheroid surface for energy physics (#185, #186). 3D analytics (radial-depth curves, volumetric heatmaps) and the consuming binary land with #194. |
+| `oxygen` | 3D radial O₂ gradients for spheroid tumors: `radial_o2_field` (per-cell `exp(-d/λ)` factor) and `radial_o2_zone_kill_rates` (normoxic / transition / hypoxic zone census). First-order Krogh approximation; pure functions for composable cycling (#187). |
 | `immune` | ICD/DAMP immune cascade: ferroptotic death quality drives dendritic cell activation and T cell priming |
 | `io` | JSON and CSV output helpers |
 | `drug_transport` | Krogh cylinder drug penetration model |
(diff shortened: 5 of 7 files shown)


ELares:
# Code Review

## Overview
Adds `ferroptosis_core::oxygen` with two pure functions — `radial_o2_field` (per-cell `exp(-d/λ)` factor over a `TumorGrid3D`) and `radial_o2_zone_kill_rates` (3-shell concentric dead-cell census matching `sim-tme`'s 2D semantics). Strictly additive: no existing `ferroptosis-core` module is touched, `sim-tme`/`sim-spatial` are untouched, and the crate bumps 0.7.0 → 0.8.0. +486 LOC in `oxygen.rs` (216 module / 270 tests).

The work is high-quality, well-documented, and self-aware about its own limits. A few specific issues worth addressing.

## Correctness

- **Cross-geometry alignment is exact.** `sim-tme/src/main.rs:836-875` and `oxygen.rs:129-158` use identical threshold logic (`< shell`, `< 3·shell`, else hypoxic), identical empty-zone-returns-0 fallback, identical stromal-skip, and identical `(tumor_radius − dist) · cell_size` depth math. The claim "matches sim-tme's 2D semantics exactly" holds.
- **Formula parity is exact.** `apply_o2_gradient` at `sim-tme/main.rs:98` is `(-depth/λ).exp().clamp(0, 1)`; `oxygen.rs:73` is the same. ✓

### Bug — docstring claim about NaN handling is wrong
`oxygen.rs:60-63` states:
> release builds will produce `NaN`/`Inf` for `λ ≤ 0` and silently clip `NaN` results to `0.0` via `f64::clamp`.

`f64::clamp` propagates NaN — `NaN.clamp(0.0, 1.0) == NaN`. So `λ = NaN` yields a `Vec<f64>` full of `NaN`s, not zeros. `λ = 0.0` happens to round-trip to `0.0` via `exp(-∞) = 0`, but `λ < 0` produces `1.0` (since `exp(+depth/|λ|)` saturates upward), not 0/NaN as implied. Two options:

1. Fix the docstring to say what release actually does (\"invalid `λ` produces undefined values; release callers should validate at boundary\").
2. Replace `debug_assert!` with a debug-or-release branch that clamps invalid `λ` to a known value, or returns a `Result`.

Either is fine — but the current text is actively misleading.

## Test coverage

Strong overall (10 new tests; 115 total pass). Two gaps and one fragility:

- **No test asserts what happens for invalid `λ` in release.** The docstring describes release behavior but no `#[cfg(not(debug_assertions))]` test pins it. Given that the docstring is wrong about it (above), this matters: a future maintainer changing the validation will silently change observable behavior.
- **`distinct_lambdas_produce_distinct_fields` does not actually test cycling.** It tests that *two* static fields differ across λ values — true by construction since the function is deterministic in λ. The PR body sells this test as covering AC #2 (\"O₂ cycling works in 3D\"), but cycling is a *consumer* concern (re-calling per step). The pure-function design implies cycling trivially works; the test name and PR claim oversell what's verified. Either rename it (e.g. `field_varies_with_lambda`) or add a tiny step-loop that demonstrates alternating-λ usage.
- **Acceptance test duplicates `sim-tme` 2D math inline.** `oxygen.rs:415-441` re-implements `zone_kill_rates`'s geometry rather than importing it. Justified (sim-tme is a binary, not a library) but creates silent-drift risk: if `sim-tme`'s `zone_kill_rates` changes thresholds or `tumor_radius_factor`, this test still passes against stale duplicated code. Worth a comment pointing at `sim-tme/main.rs:836` as the source of truth, so future readers know to keep them in sync.

## Performance

- Two passes over `grid.cells` per call (`radial_o2_field` + `radial_o2_zone_kill_rates`), each calling `radial_depth_um` which the PR notes recomputes `center_{r,c,l}` and `tumor_radius` per cell. PR body and `oxygen.rs:80-87` both call this out and point at #194 for the binary-level inner-loop optimization. Acceptable scope boundary — no action here.
- Allocation: `radial_o2_field` returns `Vec<f64>` of length `cells.len()`. For 100³ that's 8 MB per call. If sim-spatial-3d's inner loop calls this every step, that's a lot of allocator pressure. Not in scope for this PR, but worth flagging for #194: an `Iterator`-returning variant or a caller-provided buffer (`fn radial_o2_field_into(grid, λ, out: &mut Vec<f64>)`) would avoid it.

## API design

- **Pure-functions choice is right.** The decoupling from `cell.basal_ros` mutation is well-justified in the rustdoc and pays off immediately in the cycling case.
- **Stromal convention (`O₂ = 1.0`)** is consistent with 2D and explicitly documented. Good.
- **`debug_assert!`-only validation** matches the `physics` module posture, per the PR. Fine — just fix the docstring (above).

## PR body / project hygiene

- Bit-identity claim is credible: zero behavioral changes to existing modules; only `lib.rs` adds `pub mod oxygen;` + a doc table row. SHA-256s are verifiable by anyone re-running sim-spatial.
- The \"issue text inverted\" finding in the PR body and in `matched_lambda_2d_vs_3d_zone_fractions`'s docstring is exactly the kind of honest correction the project's principles ask for. Worth following through on the #187 issue edit so future readers don't trip over the original text.

## Risks

- Low. Strictly additive; downstream #188/#191/#197 consumers don't exist yet, so an API tweak post-merge is still cheap.
- Watch for `sim-tme` ↔ `oxygen.rs` drift over time — the duplicated-math test is the only thing guarding the cross-geometry contract, and only at fixed seed/dimensions.

## Recommendation
**Approve with two small follow-ups before merge:**
1. Fix the NaN-clamping claim in `oxygen.rs:60-63` (1-line docstring edit).
2. Either rename or expand `distinct_lambdas_produce_distinct_fields` so it actually exercises cycling (or drop the AC #2 framing from the PR body).

Both are minor; neither blocks the science. Everything else — including the inverted-issue-text finding — is exemplary.

ELares:
Automated review: the changes compile and the workspace test suite passes. I did not identify any discrete correctness issues introduced by this patch.

ELares:
Thanks for the careful review — all three items addressed in 4e3c2165:

**Fixed:**

1. ✅ **NaN/Inf docstring bug (real).** You were right — I verified empirically that `NaN.clamp(0,1) = NaN` (the comparisons used internally both return false on NaN, so self passes through). My original \"silently clip NaN to 0.0\" was actively misleading. Replaced with a release-behavior table covering all invalid-input classes:

   | `λ` value | Per-cell output (release) |
   |---|---|
   | finite, `> 0` | valid `[0, 1]` |
   | `0.0` | `0.0` (`exp(-∞)` then clamp) |
   | `< 0` | `1.0` (`exp(+depth/\|λ\|)` saturates, then clamp) |
   | `NaN` | `NaN` (`f64::clamp` propagates NaN, does not clip) |
   | `+∞` | `1.0` (`exp(-depth/∞) = 1.0`) |

2. ✅ **Misleading test name + missing cycling demo.** You're right that `distinct_lambdas_produce_distinct_fields` is trivially true for any deterministic function of λ — calling that \"cycling\" was overclaiming. Two fixes:
   - **Renamed** to `field_varies_with_lambda` (honest about what's tested).
   - **Added** `o2_cycling_pattern_demonstrates_consumer_loop` — runs a short 4-step loop mirroring sim-tme's `cycling_lambda` (binary-local helper, replicated inline since it's not in the library), asserts normoxic-phase field ≠ hypoxic-phase field, AND asserts two same-phase steps are bit-identical (proves no hidden state in the function — important for any downstream consumer relying on `radial_o2_field`'s purity).

3. ✅ **sim-tme drift comment.** Added an explicit \"**Source of truth — keep in sync with `sim-tme/src/main.rs:836`**\" warning above the inline 2D math in `matched_lambda_2d_vs_3d_zone_fractions`, with a note that there's no automated guard.

**Acknowledged, not changing:**

- The buffer-reuse / `Iterator`-returning `radial_o2_field` variant — agree this is worth flagging for #194 but it's premature optimization without a real consumer. The current `Vec<f64>` allocation is ~8 MB for 100³ and microseconds of overhead; only matters in tight inner loops where #194 will know the actual usage pattern.
- The buffer-reuse perf note will land naturally in #194's binary-design plan.

**Follow-through on the #187 issue text inversion:** will add a comment to issue #187 on merge explaining the geometry-vs-biology distinction so future readers don't trip on the original wording.

**Verification:** 116 tests pass (was 115; +1 cycling test); `cargo fmt` clean on oxygen.rs; no behavior changes beyond the new test.

ELares:
# Re-review (commit 4e3c2165)

All three follow-ups from my previous review are addressed cleanly.

## 1. NaN docstring → fixed correctly

The new table at `oxygen.rs:75-83` enumerates release behavior for each invalid `λ` class. I verified each row by tracing the math:

| `λ` | Trace | Output |
|-----|-------|--------|
| `0.0` | `exp(-d/0) = exp(-∞) = 0.0` → clamp → `0.0` | ✓ |
| `< 0` | `exp(-d/-|λ|) = exp(+d/|λ|)` → ∞ → clamp → `1.0` | ✓ |
| `NaN` | `(-d/NaN).exp() = NaN` → `clamp` propagates → `NaN` | ✓ |
| `+∞` | `exp(-d/∞) = exp(-0) = 1.0` | ✓ |

One degenerate corner the table doesn't cover: `depth = 0` AND `λ = 0` gives `-0.0 / 0.0 = NaN` (IEEE 754), so a surface cell with `λ = 0` returns `NaN` rather than `0.0`. Not worth blocking on — anyone passing `λ = 0` is already in undefined-input territory — but worth knowing if it ever bites.

## 2. Test rename + new cycling test → strong

- `field_varies_with_lambda` is honest naming; the docstring at `oxygen.rs:320-326` explicitly disclaims it as a cycling test and points readers at the new one. Good.
- `o2_cycling_pattern_demonstrates_consumer_loop` (`oxygen.rs:367-405`) actually exercises the cycling pattern: it runs a 4-step loop, alternates `λ` via a sim-tme-mirroring `cycling_lambda` closure, and asserts:
  - normoxic-phase field ≠ hypoxic-phase field (cycling does something)
  - two normoxic-phase steps are bit-identical (function is stateless in `λ` — downstream consumers can rely on this)
  - same for two hypoxic-phase steps

The bit-identity assertions also act as a regression guard against anyone sneaking RNG or parallel non-determinism into `radial_o2_field` later. Nice bonus.

## 3. Drift comment → appropriately strong

The block at `oxygen.rs:480-485` calls out the unguarded contract explicitly, names `sim-tme/src/main.rs:836` as the source of truth, and even offers the remediation path (move into the library). A future maintainer touching either side will see it.

## Verification

- `cargo test --release -p ferroptosis-core --lib`: **116 passed; 0 failed** (was 115; +1 new cycling test). Confirmed locally.
- Diff is +78 LOC, docstring + one new test + one in-test comment — no production-code behavior change beyond the rename. SHA-256 bit-identity claim from the original PR body remains valid.

## Recommendation

**Approve.** Both review nits and the underlying bug are resolved. No further changes requested.

ELares (commented):
Re-review finding:

Low: simulations/ferroptosis-core/src/oxygen.rs:83 documents lambda_um = 0.0 as producing 0.0 in release, but that is not true for every cell. Stromal cells still return 1.0 through the early branch, and tumor cells exactly on the surface have depth 0, so -0.0 / 0.0 becomes NaN and f64::clamp propagates it. Since invalid lambdas are only debug_asserted, the table should either avoid promising per-cell release outputs for invalid input or call out the surface/stromal exceptions.

I did not identify production behavior issues in the valid-input path.

Verification:
- cargo test -p ferroptosis-core passed: 116 unit tests + 1 doctest
- cargo test --workspace passed
- rustfmt --check simulations/ferroptosis-core/src/oxygen.rs passed
- cargo fmt --all --check still reports pre-existing formatting drift outside this patch

Issue #195, opened 2026-04-26
sim-tme-3d: Full TME simulation in 3D spheroid geometry


ELares:
## Context
The capstone 3D binary — combines all TME features (O2 gradients, immune coupling, stromal shielding, pH effects) in a single 3D spheroid simulation. This is the 3D equivalent of sim-tme and tests whether ALL four resistance/amplification mechanisms hold in 3D.

## What Needs to Be Done
1. New workspace member: \`simulations/sim-tme-3d/\`
2. Integrates: 3D O2 gradient (#187), 3D immune coupling (#188), 3D stromal layer (#189), 3D pH (#190)
3. Optional: 3D vasculature network (#191) if available, otherwise use radial model
4. O2 sweep across λ values (same as 2D: 80, 100, 120, 150μm)
5. O2 cycling toggle in 3D
6. Zone-level kill rates for spherical shells (normoxic, transition, hypoxic)
7. DAMP field output as 3D volume
8. Comparison table: 2D vs 3D for every TME feature

## Key Questions This Binary Answers
1. Does the hypoxia RSL3 collapse (3.7% → 0.1%) hold in 3D? (Expected: yes, possibly stronger — more hypoxic volume)
2. Does the immune 104:1 ratio hold in 3D? (Unknown — DAMP density may decrease in 3D volume)
3. Does stromal shielding have MORE impact in 3D? (Expected: yes — larger boundary fraction)
4. Does pH ion trapping produce similar RSL3 reduction in 3D? (Expected: similar — same chemistry)

## Acceptance Criteria
- [ ] All four TME features operational in 3D
- [ ] Output includes zone-level kill rates + DAMP fields + comparison table
- [ ] 2D vs 3D comparison for each TME feature documented
- [ ] README.md with full documentation
- [ ] Manuscript-ready: identifies which 2D predictions are dimensionality-dependent

## Dependencies
- #185 (3D Grid Infrastructure)
- #186 (3D Energy Physics)
- #187 (3D O2 Gradient)
- #188 (3D Immune Coupling)
- #189 (3D Stromal Shielding)
- #190 (3D pH Gradient)
- #192 (3D Performance)
- #193 (3D Visualization)

ELares:
Review pass: this is the right capstone issue, but it is still carrying too many built-in expectations.

[...]

Issue #189, opened 2026-04-26
3D Stromal Shielding: CAF boundary layer in spheroid geometry


ELares:
## Context
CAF-mediated stromal protection is modeled as a single-cell-deep boundary layer in 2D (1,796 of 159,033 tumor cells = 1.1%). In 3D spheroid geometry, the surface-to-volume ratio changes: the boundary layer is a spherical shell, and its fraction of total tumor cells depends on sphere radius and shell thickness.

## What Needs to Be Done
1. Identify 3D boundary cells (tumor cells with at least one stromal Moore-26 neighbor)
2. Apply CAF GSH/MUFA supply boosts to boundary cells (same per-cell mechanics)
3. Compute 3D stromal-adjacent fraction and compare to 2D (1.1%)
4. Report stromal-specific kill rates in 3D

## Biology
- In 3D, the stromal boundary shell fraction scales as ~3×(cell_diameter/tumor_radius)
- For a 100-cell-radius sphere: boundary ≈ 3% of cells (vs 1.1% in 2D)
- Larger boundary fraction means stromal shielding has a BIGGER impact in 3D

## Acceptance Criteria
- [ ] 3D boundary cell identification (26-neighbor check)
- [ ] CAF supply boosts applied in 3D
- [ ] Comparison: 2D vs 3D boundary fraction and stromal kill rates
- [ ] Document whether stromal shielding impact increases in 3D (expected: yes)

## Dependencies
- #185 (3D Grid Infrastructure)

ELares:
Review pass: the geometry motivation is sound, but the issue currently bakes in the biological conclusion a little too hard.

The larger boundary fraction in 3D is a strong reason to expect a bigger stromal effect, but that is still a hypothesis. I would keep the boundary-fraction calculation as the firm part, and phrase the efficacy consequence as something to measure rather than something expected to be true by default.

That will make the eventual result easier to trust if the per-cell shielding effect turns out not to scale as cleanly as the shell geometry does.

ELares:
## Rigorous Planning Analysis

### Problem restated
CAF protection applied to boundary cells (at least 1 stromal neighbor). In 3D, boundary fraction scales as ~3/R (sphere) vs ~2/R (circle), meaning MORE cells are protected.

### Assumptions verified
- Boundary fraction larger in 3D: TRUE at matched R. At R=50: 2D~4%, 3D~6%
- CAF biology same in 2D/3D: TRUE for the model (per-cell GSH/MUFA supply)
[...]

PR #216, opened 2026-05-11
Add 3D radial pH gradient for spheroid tumors (#190)

### LINKED ISSUE #190: 3D pH Gradient: Radial acidification in tumor spheroids
## Context
The pH model uses `pH(d) = 7.4 - 0.9 × (1 - exp(-d/λ))` where d is distance from the 2D tumor edge. In 3D, the pH gradient is radial from the spheroid surface inward, with the most acidic regions in the core (matching the hypoxic core where anaerobic glycolysis produces the most lactate).

## What Needs to Be Done
1. Compute 3D radial pH field using radial distance from spheroid surface
2. Apply ion trapping (Henderson-Hasselbalch) and iron release modulation in 3D
3. Update pH zone analysis for spherical shells

## Key Physics
- pH at well-perfused periphery: ~7.4
- pH in tumor core: ~6.5 (Warburg effect / lactic acid accumulation)
- Ion trapping reduces RSL3 bioavailability at low pH (36% loss at pH 6.5)
- Iron release increases Fenton ROS at low pH (2.35× at pH 6.5)

## Acceptance Criteria
- [ ] Radial pH field in 3D grid
- [ ] Ion trapping and iron modulation work in 3D
- [ ] Compare 2D vs 3D pH kill effects at matched parameters

## Dependencies

ELares:
## Summary

Closes #190. Adds the per-cell pH field and two scalar modulation helpers (iron release, ion trapping) that downstream 3D consumers (#188, #195, #197) need. Strictly additive — sim-tme and sim-spatial untouched, sim-spatial 100×100 SHA-256 bit-identical.

- New module `ferroptosis_core::ph` with three pure functions
- Bumps `ferroptosis-core` 0.8.0 → 0.9.0 (additive minor)
- 13 new tests; 129 total ferroptosis-core tests pass

## API

```rust
/// Per-cell pH on a 3D spheroidal grid. Stromal cells return ph_edge.
pub fn radial_ph_field(grid: &TumorGrid3D, ph_edge: f64, ph_core: f64, lambda_ph_um: f64) -> Vec<f64>;

/// Iron-release multiplier (ferritin destabilization at low pH).
/// `1.0 + sens·(ph_edge - local_ph)` — 2.35× at pH 6.5 (defaults).
pub fn iron_multiplier_from_ph(local_ph: f64, ph_edge: f64, sensitivity: f64) -> f64;

/// Ion-trapping drug-availability factor (Henderson-Hasselbalch linearized).
/// `1.0 - sens·(ph_edge - local_ph)`, clamped to `[0.3, 1.0]` — 0.64 at pH 6.5 (defaults).
pub fn ion_trap_factor_from_ph(local_ph: f64, ph_edge: f64, sensitivity: f64) -> f64;
```

Parallels `oxygen.rs` (#187) structurally: pure functions, no mutation, full release-behavior table for invalid inputs in the rustdoc (NaN propagation explicit per PR #215 review lesson).

## Acceptance criteria (#190)

- [x] **Radial pH field in 3D grid** — `radial_ph_field`
- [x] **Ion trapping and iron modulation work in 3D** — `iron_multiplier_from_ph` + `ion_trap_factor_from_ph` (pure scalars; consumer applies to cells)
- [x] **Compare 2D vs 3D pH kill effects at matched parameters** — `matched_lambda_2d_vs_3d_acidic_fraction` test. Interpreted as comparison of pH-field *shape* (acidic-fraction with `pH < 6.8` as bucket threshold), not actual kill simulation — kill comparison would require full multi-step runs and is out of scope here.

## AC #3 interpretation note

Issue says "Update pH zone analysis for spherical shells," but sim-tme has no pH-specific zone analysis function. The existing `radial_o2_zone_kill_rates` (from #187) is depth-based and geometry-agnostic — works on any cell state including pH-modulated. So AC #3 is satisfied by composition; no pH-specific zone function added. Documented in the cross-geometry test's rustdoc.

## Cross-geometry finding (consistent with #187)

Same cubic-vs-quadratic geometry as the O₂ test: at matched R, λ, threshold, the 3D acidic-volume fraction is *smaller* than the 2D acidic-area fraction. The biological "3D more acidic" observation reflects vasculature effects (smaller effective λ), not pure geometry. The test asserts the geometrically-correct relationship and the docstring explains the distinction.

## Verification

- 129 ferroptosis-core tests pass (was 116; +13 new in `ph::tests`)
  - `spatial_summary.json`: `f7628f85fe86217142f078d2857caefb6c4ac400bf2837abf3c96c2808e13e66`
  - `depth_kill_curves.csv`: `2eea07e1ef94ff5233f470c434183c4ad413593ed34d5dacef9fdbaa0059a72e`
- `cargo fmt -- --check` clean on `ph.rs`

## Test plan

- [x] `cargo test --release -p ferroptosis-core` passes locally (129 tests)
- [x] `cargo build --release --workspace --exclude ferroptosis-python` clean
- [ ] CI green

## Follow-ups

- **Unblocks**: #195 (sim-tme-3d) is now closer — still blocked by #188 (immune coupling) and #189 (stromal shielding), but pH and O₂ are both ready.
- **Out of scope**: Python/FFI bindings (intentional — matches #186/#187 scope); refactor of sim-tme to import library helpers (#195 will drive that)
- **Source-of-truth fragility**: cross-geometry test replicates sim-tme/main.rs:520 inline; documented with "keep in sync" warning. No automated guard. Same pattern as #187's cross-geometry test.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - pathway-target and resistant-state analysis
 - diagnostic-to-therapy chain extraction (6 chains, 129 articles mapped)
 - tissue-of-origin analysis layer (5 tissue categories, 62% coverage)
-- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186, 3D radial O₂ gradient + zone-census #187) for the #185–#197 spheroid-validation series
+- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186, 3D radial O₂ gradient + zone-census #187, 3D radial pH gradient + iron/ion-trap helpers #190) for the #185–#197 spheroid-validation series
 - ferroptosis-core library packaging for external use
 - news source authentication pipeline (fetch, extract claims, verify, score, index)
 - broader strategy review of alternative therapies and biological bottlenecks
@@ -51,7 +51,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - tissue-of-origin and weighted-evidence layers are active
 - diagnostic-therapy matching layer covers 6 chains across 4 modalities (radioligand, checkpoint, mRNA vaccine, oncolytic)
 - manuscript: 112 pages (book format), 11 chapters + 3 appendices, 20 figures, ~36,700 words
-- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 12 modules including `photosensitizer_pk` and `oxygen`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; v0.8.0 adds 3D radial O₂ field + zone-census #187; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
+- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 13 modules including `photosensitizer_pk`, `oxygen`, and `ph`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; v0.8.0 adds 3D radial O₂ field + zone-census #187; v0.9.0 adds 3D radial pH field + iron-release/ion-trap helpers #190; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
 - news authentication pipeline: 5 scripts (fetch, extract claims, verify against PubMed, score credibility, build claim-centric index) implementing the 3-tier source framework from analysis/news-source-criteria.md
 - simulation calibration: 5 targets documented, evaluate script operational
 - drug penetration module: 3 tissue types, exponential Krogh approximation
--- a/simulations/Cargo.lock
+++ b/simulations/Cargo.lock
@@ -217,7 +217,7 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "ferroptosis-core"
-version = "0.8.0"
+version = "0.9.0"
 dependencies = [
  "csv",
  "ndarray",
--- a/simulations/README.md
+++ b/simulations/README.md
@@ -21,7 +21,7 @@ Each binary has its own README with parameters, output format, example commands,
 
 ## Shared library
 
-`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), 3D radial O₂ gradients (`oxygen::radial_o2_field` + zone-census, #187), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
+`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), 3D radial O₂ gradients (`oxygen::radial_o2_field` + zone-census, #187), 3D radial pH gradient (`ph::radial_ph_field` + iron-release and ion-trapping helpers, #190), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
 
 ## Quick start
 
--- a/simulations/ferroptosis-core/Cargo.toml
+++ b/simulations/ferroptosis-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ferroptosis-core"
-version = "0.8.0"
+version = "0.9.0"
 edition = "2021"
 description = "Embeddable ferroptosis biochemistry engine for cancer simulation"
 license = "MIT"
--- a/simulations/ferroptosis-core/README.md
+++ b/simulations/ferroptosis-core/README.md
@@ -35,6 +35,7 @@ Run the included example: `cargo run -p ferroptosis-core --example basic_usage`
 | `physics` | Depth-dependent energy deposition: Beer-Lambert (PDT), acoustic attenuation (SDT), uniform (RSL3). 2D row-based (`local_ros_multiplier`) and 3D radial-depth (`local_ros_multiplier_3d`) dispatchers share the same per-treatment depth functions (#186). |
 | `grid` | 2D `TumorGrid` (8-Moore, circular) and 3D `TumorGrid3D` (26-Moore, spherical) with heterogeneous architecture, neighbor iteration, iron diffusion. `TumorGrid3D::radial_depth_um` provides per-cell signed depth from the spheroid surface for energy physics (#185, #186). 3D analytics (radial-depth curves, volumetric heatmaps) and the consuming binary land with #194. |
 | `oxygen` | 3D radial O₂ gradients for spheroid tumors: `radial_o2_field` (per-cell `exp(-d/λ)` factor) and `radial_o2_zone_kill_rates` (normoxic / transition / hypoxic zone census). First-order Krogh approximation; pure functions for composable cycling (#187). |
+| `ph` | 3D radial pH gradient for spheroid tumors: `radial_ph_field` (per-cell `pH(d) = ph_edge - delta·(1 - exp(-d/λ))`) plus pure-scalar helpers `iron_multiplier_from_ph` (ferritin destabilization) and `ion_trap_factor_from_ph` (linearized Henderson-Hasselbalch, weak-base drug bioavailability). Same form as sim-tme's 2D pH; pure functions for #195 sim-tme-3d (#190). |
 | `immune` | ICD/DAMP immune cascade: ferroptotic death quality drives dendritic cell activation and T cell priming |
 | `io` | JSON and CSV output helpers |
 | `drug_transport` | Krogh cylinder drug penetration model |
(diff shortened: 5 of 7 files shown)


ELares:

Caught **3 docstring bugs** in the release-behavior table during self-review. All fixed in 1d01cf9c. No behavior changes — pure documentation accuracy. Verified each by empirical Rust script before fixing.

**Bug 1**: \`λ < 0\` table claim was *inverted*. I wrote \"clamped to ph_core\" but the math gives `> ph_edge` → clamps to **ph_edge** (not ph_core). With λ=-50, depth=100, defaults: raw = 8.95 → clamped to 7.4.

**Bug 2** (most important): \`ph_edge <= ph_core\` claim \"formula still runs\" was wrong. \`f64::clamp\` has an unconditional \`assert!(min <= max)\` that runs in release builds — passing inverted bounds **PANICS** with \"min > max, or either was NaN.\" My \`debug_assert!(ph_edge > ph_core)\` catches it in tests, but release panics from inside clamp.

**Bug 3**: \"any NaN input → per-cell NaN\" was partially wrong. True for \`λ = NaN\` (NaN propagates into clamp's *self*; clamp preserves NaN-self). FALSE for \`ph_edge = NaN\` or \`ph_core = NaN\` — those become clamp's *bounds*, and \`NaN <= anything\` returns false → release panic.

**Also added** to the table: \`λ = 0\` with depth=0 specifically (`0/0 → NaN`), `λ = +∞` (returns ph_edge), `ph_edge == ph_core` (constant pH everywhere, no panic since min ≤ max holds — but debug_assert rejects as likely config bug).

**Bottom-line guidance for the docstring**: callers loading params from untrusted sources MUST validate at the boundary, particularly `ph_edge > ph_core` and finite/NaN-free `ph_edge`/`ph_core`. Otherwise a release run panics on the first tumor-cell clamp.

**Reinforced lesson** (same class as PR #215 fix): I should not claim \`f64::clamp\` release behavior without empirical verification.

**No other findings.** Production paths are correct; tests are clean; architecture is solid; cross-geometry test asserts the expected cubic-vs-quadratic property. 13 tests still pass; \`cargo fmt\` clean on \`ph.rs\`.

Ready for review.

ELares:
Review complete: No blocking correctness issues were found in the pH gradient addition. The workspace and ferroptosis-core test suites pass locally.

ELares:
# Code Review: PR #216 — Add 3D radial pH gradient for spheroid tumors

## Overview

Closes #190. Adds a new `ferroptosis_core::ph` module with three pure functions: `radial_ph_field` (per-cell pH on a 3D spheroid grid), `iron_multiplier_from_ph` (ferritin destabilization at low pH), and `ion_trap_factor_from_ph` (linearized Henderson-Hasselbalch for weak-base drug bioavailability). Strictly additive: 474 lines in one new file, plus version bump 0.8.0 → 0.9.0 and consistent updates to `lib.rs`, `README.md`, `Cargo.toml`, `Cargo.lock`, and `CLAUDE.md`. Structurally parallels the `oxygen` module from #187.

## Strengths

- **Source-of-truth fidelity verified.** Formula in `ph.rs:261` matches `sim-tme/main.rs:544-546` exactly (same `pH(d) = ph_edge - delta·(1 - exp(-d/λ))`, same `[ph_core, ph_edge]` clamp, same default sensitivities 1.5/0.4, same `[0.3, 1.0]` ion-trap floor from `sim-tme/main.rs:614`).
- **Release-behavior table (ph.rs:213-223) is genuinely thoughtful.** The asymmetric panic-vs-NaN behavior of `f64::clamp` with NaN bounds is non-obvious and is now explicitly documented per the #215 review lesson. Better than `oxygen.rs`'s lighter validation posture.
- **AC #3 reinterpretation is honest** — `apply_ph_gradient`'s "pH zone analysis" doesn't actually exist in sim-tme; calling out that the existing `radial_o2_zone_kill_rates` is geometry-agnostic and works on any cell state is the correct pragmatic choice rather than building a parallel pH-zone function that would duplicate it.
- **Tests pin both numerical claims and asymptotic behavior**: 2.35× at pH 6.5 (iron), 0.64 (ion-trap), surface-equals-edge with strict equality (rationale in docstring), monotonic decrease toward core, clamping in range, deep-tumor convergence.
- **Pre-condition reverification in `surface_tumor_cell_has_edge_ph` (ph.rs:363-368)** — re-asserts the grid test's invariant inline so this test fails loudly if the grid coordinate ever drifts.
- **sim-spatial 100×100 SHA-256 bit-identical** — confirms no incidental regression in untouched code paths.

## Issues / Suggestions

### Minor — docstring inaccuracy
- **`ph.rs:526`** (`matched_lambda_2d_vs_3d_acidic_fraction` test docstring): "Threshold: pH < 6.8 (halfway between ph_core 6.5 and a typical mild-acidity threshold 7.0; chosen to give non-empty acidic regions in both geometries)." But halfway between 6.5 and 7.0 is **6.75, not 6.8**. Either change the threshold to 6.75 or rewrite the justification (e.g., "below pH 6.8, the conventional cutoff for clinically meaningful tumor acidosis"). Same caveat-fix flavor as 1d01cf9c.

### Minor — test coverage gaps
- **No upper-bound clamp test for `ion_trap_factor_from_ph`.** The clamp is `[0.3, 1.0]`. Floor is tested (`ion_trap_clamps_to_floor`), edge-identity is tested (`ion_trap_at_edge_is_one` covers `1.0` via the formula path), but not the case where the raw formula exceeds 1.0 (e.g., `local_ph > ph_edge`, or negative sensitivity) and gets clamped down. Unreachable from `radial_ph_field` output (which is clamped to `[ph_core, ph_edge]`), but the upper bound is part of the API contract.
- **No `iron_multiplier_from_ph >= 1.0` property test for radial-field inputs.** The docstring promises `> 1` in the tumor interior, but only the endpoints (edge=1.0, default-core=2.35) are pinned. A property check over `radial_ph_field` output would lock the biology assumption.

### Minor — undocumented sensitivity-sign contract
- Both helpers assume `sensitivity >= 0`. A negative sensitivity silently inverts the biology (`iron_multiplier < 1` at acidic pH, drug becomes *more* bioavailable in the core). A `debug_assert!(sensitivity >= 0.0)` would be cheap insurance against config-file fat-fingers. Same pattern as the `ph_edge > ph_core` guard.

### Acknowledged but accumulating — source-of-truth duplication
- The cross-geometry test (`ph.rs:541-599`) inlines sim-tme's 2D depth math. The PR body and test docstring both flag the "keep in sync" caveat with no automated guard. This is consistent with #187's approach, but the technical debt is now in two field modules — when #195 lands the sim-tme-3d refactor, the explicit retirement of these duplications should be part of the cleanup, not deferred.

### Stylistic — light
- `ion_trap_clamps_to_floor` (`ph.rs:511-514`) asserts both `f == ION_TRAP_FLOOR` and `ION_TRAP_FLOOR == 0.3`. The second line is a tautology against the constant; useful as a doc-binding "this value is sim-tme's 0.3 by contract, not arbitrary," but slightly unusual. Either drop the second assertion or add a comment that it locks the public-facing magic number.

## Correctness, performance, security

- **Correctness**: no bugs found. Math, sign conventions, and clamp semantics are right; numerical-claim tests anchor them.
- **Performance**: O(N) per call; same per-cell `radial_depth_um` recompute as `oxygen.rs`. Already flagged for #194 hoisting in the docstring. Not a concern at module-level granularity.
- **Security**: no external input handling; debug-assert validation posture is appropriate for a library.

## Verdict

Approve with the 6.8 vs 6.75 docstring fix. The other suggestions (sensitivity-sign assert, two missing clamp/property tests) are quality improvements rather than blockers; reasonable to land as a follow-up touch-up alongside the next pH consumer (#188 or #195) where they'd be exercised. No correctness or release-quality risks identified, and the bit-identical SHA-256 verification rules out incidental regressions.

ELares:
Thanks for the careful review — all five items addressed in 2a9d595a:

**Fixed (all 5):**

1. ✅ **Must-fix: "halfway between 6.5 and 7.0 is 6.75, not 6.8" doc inaccuracy.** Kept the threshold at 6.8 (avoids changing test counts/assertions) and rewrote the justification: "conventional cutoff for moderate tumor acidosis (between the typical mild-acidity boundary ~7.0 and the severe-core regime ~6.5)."

2. ✅ **Sensitivity-sign contract.** Added \`debug_assert!(sensitivity >= 0.0)\` to both \`iron_multiplier_from_ph\` and \`ion_trap_factor_from_ph\` plus docstring paragraphs explaining the biology inversion (iron: contradicts ferritin-destabilization biology per Yu et al. 2017; ion-trap: contradicts Henderson-Hasselbalch for weak bases). Cheap insurance against config-file fat-fingers.

3. ✅ **Upper-bound clamp test for \`ion_trap_factor_from_ph\`.** Added \`ion_trap_clamps_to_upper_bound\`: local_ph=8.4 (above ph_edge), ph_edge=7.4, sens=0.4 → raw=1.4 → clamps to 1.0. Validates the upper half of the [0.3, 1.0] contract.

4. ✅ **\`iron_multiplier >= 1.0\` property test.** Added \`iron_multiplier_is_at_least_one_for_field_inputs\` — walks every tumor cell in a 20³ \`radial_ph_field\` output and asserts iron_multiplier(local_ph, ph_edge, 1.5) >= 1.0. Locks the biology invariant over the full field, not just endpoints.

5. ✅ **Stylistic: tautology assert in \`ion_trap_clamps_to_floor\`.** Added a comment explaining the intent: it locks the public-facing contract that the floor is exactly 0.3 (sim-tme convention), defends against an accidental const change.

**Acknowledged, deferred per reviewer:**

- Source-of-truth duplication in the cross-geometry test (inlines sim-tme's 2D depth math, no automated guard). Same accumulating-tech-debt pattern as #187. You correctly noted explicit retirement should land with #195 sim-tme-3d's cleanup, not deferred indefinitely. Tracked.

**Verification:**
- 131 ferroptosis-core tests pass (was 129; +2 new clamp/property tests)
- \`cargo fmt --check\` clean on \`ph.rs\` (pre-existing drift in other files is #209)
- No production-path changes; sim-spatial 100×100 SHA bit-identical

CI re-running on 2a9d595a.

ELares:
# Re-review: response to first-pass review (commit 2a9d595a)

All 5 items from the prior review addressed cleanly. Verifying each:

## Item-by-item verification

**1. "6.8 halfway between 6.5 and 7.0" docstring fix (`ph.rs:456-460`) — fixed.**
Threshold kept at 6.8 (preserves test assertion); justification rewritten to "conventional cutoff for moderate tumor acidosis (between the typical mild-acidity boundary ~7.0 and the severe-core regime ~6.5)." Honest and defensible. Good call to keep the numeric value — the assertion is on the 3D < 2D *relationship*, not on the absolute threshold, so the value only needs to produce non-empty buckets.

**2. Sensitivity-sign guards (`ph.rs:170-173, 200-203`) — added.**
`debug_assert!(sensitivity >= 0.0)` on both helpers, with rustdoc paragraphs explaining *why* (Yu et al. 2017 for iron, Henderson-Hasselbalch for ion-trap). The asserts use `>= 0`, not `> 0`, which is the right call — `sens=0.0` is allowed and exercised by `iron_multiplier_at_edge_is_one` / `ion_trap_at_edge_is_one`. Worth noting: `debug_assert!` is compiled out in release, and the docstrings explicitly say "release silently inverts the curve" — consistent with the rest of the module's posture and faithfully documented.

**3. Upper-bound clamp test (`ph.rs:415-422`) — added.**
`ion_trap_clamps_to_upper_bound`: local_ph=8.4, ph_edge=7.4, sens=0.4 → raw = 1.0 - 0.4·(7.4 - 8.4) = 1.0 + 0.4 = 1.4 → clamps to 1.0. Math checks out. Closes the upper half of the `[0.3, 1.0]` contract that was unreachable from `radial_ph_field` outputs but is part of the public API.

**4. Property test for iron multiplier (`ph.rs:430-449`) — added.**
`iron_multiplier_is_at_least_one_for_field_inputs` walks every tumor cell in a 20³ grid and asserts `>= 1.0` (correctly `>=`, not `>`, since surface cells return exactly `ph_edge` → exactly `1.0`). Locks the biology invariant over the full field. Skipping stromal cells with `if !gc.is_tumor { continue; }` is fine — stromal would trivially satisfy 1.0 anyway.

**5. Tautology comment in `ion_trap_clamps_to_floor` (`ph.rs:406-408`) — addressed.**
Comment now explains the intent: it locks the public-facing contract that the floor is exactly 0.3 (sim-tme convention), defending against a silent change to the private constant. Clear.

## Adversarial second pass — small things noticed on re-read

These are minor and would not block on their own; flagging for completeness.

- **Helpers lack their own release-behavior tables.** `radial_ph_field` has a thorough panic-vs-NaN-vs-clamp table; the helpers do not. Edge cases worth knowing: `iron_multiplier_from_ph(local_ph=NaN, ...)` returns NaN; `ion_trap_factor_from_ph(local_ph=NaN, ...)` — `f64::clamp` preserves NaN-self, so the function returns NaN (NOT clamped to floor). Realistic, since callers pass `radial_ph_field` outputs which are finite by construction; but a future caller threading NaN through might be surprised by the asymmetry. Lower priority than the field-function table.

- **`iron_multiplier_from_ph` with extreme sensitivity isn't bounded.** Per the docstring's "at extreme acidity it can grow large." With `sens = f64::INFINITY` and `delta > 0` the multiplier becomes `+∞`; with `sens = f64::INFINITY` and `local_ph == ph_edge` the formula evaluates to `1.0 + ∞·0 = 1.0 + NaN = NaN`. Won't happen with sane configs but matches the existing "consumer responsible" framing. Not a blocker.

- **Deferred (acknowledged in the commit msg)**: source-of-truth duplication of sim-tme's 2D math in the cross-geometry test. Same plan: retire with #195's cleanup.

## CI

`test` pass, `Analyze (actions/c-cpp/javascript-typescript/python)` pass; `Analyze (rust)` pending at time of re-review; `CodeQL` skipped. No regressions detected.

## Verdict

LGTM — approve. Response commit is exemplary: every item addressed with a test or assert (not just a docstring), math verified, and the biological rationale for each new guard is explicit. The two small follow-ups above (helper release-behavior tables, sensitivity-bound documentation) are paper-cuts and can land alongside the next pH consumer rather than holding this up.

ELares:
Re-review complete: no blocking correctness issues found in the updated pH gradient addition.

Checked the 3D pH field math against sim-tme’s 2D pH formula, the TumorGrid3D flat indexing/order, and the updated range/sensitivity tests.

Verified locally:
- `cargo test -p ferroptosis-core` passes (131 unit tests + 2 doctests)
- `cargo test --workspace` passes

ELares:
Thanks for the re-review and the LGTM. Addressed both paper-cuts anyway in 724c293c — pure docstring additions, no behavior change, closes the loop on the NaN-surprise asymmetry rather than carrying it forward to the next pH consumer.

**Fixed (both):**

1. ✅ **Helper release-behavior tables.** Both \`iron_multiplier_from_ph\` and \`ion_trap_factor_from_ph\` now have explicit \"Release behavior for invalid inputs\" tables covering NaN propagation (self vs argument), infinity behavior, and the panic-free contract. Format symmetric with \`radial_ph_field\`'s table.

2. ✅ **Sensitivity-bound documentation.** \`iron_multiplier_from_ph\`'s table explicitly covers \`sensitivity = +∞\` (±∞ except at \`local_ph == ph_edge\` where it gives NaN per IEEE \`∞·0\` rule).

The new tables reinforce the **panic-asymmetry** within the module:
- \`radial_ph_field\` can panic via \`f64::clamp\` on bad bounds (documented in its table)
- The two scalar helpers can NOT panic — they're either clamp-free (iron) or clamp to constant bounds (ion-trap)

This means consumers can safely iterate a known-valid \`radial_ph_field\` output through the helpers in hot loops without panic-risk anxiety, and the docstrings now make that contract explicit.

**Verification:** 131 ferroptosis-core tests still pass; \`cargo fmt --check\` clean on \`ph.rs\`; no production-path changes.

Ready for merge.

PR #217, opened 2026-05-11
Add 3D CAF-shielded boundary detection for spheroid tumors (#189)

### LINKED ISSUE #189: 3D Stromal Shielding: CAF boundary layer in spheroid geometry

ELares:
## Summary

Closes #189. Adds the boundary-detection primitive that downstream 3D consumers (#188 immune coupling, #195 sim-tme-3d) need. Strictly additive — sim-tme and sim-spatial untouched, sim-spatial 100×100 SHA-256 bit-identical.

- New module \`ferroptosis_core::stromal\` with two pure functions
- Bumps \`ferroptosis-core\` 0.9.0 → 0.10.0 (additive minor)
- 11 new tests; 142 total ferroptosis-core tests pass
- **Completes the TME-field trilogy** (oxygen + pH + stromal) — #195 now blocked only by #188

## API

\`\`\`rust
/// Flags tumor cells with at least one stromal 26-Moore neighbor.
/// Length matches grid.cells. Stromal cells are always false.
pub fn stromal_adjacency_mask(grid: &TumorGrid3D) -> Vec<bool>;

/// Dead-rate among flagged cells; 0.0 for empty mask.
/// Uses assert!() (not debug_assert!) for mask-length validation —
/// contract violation deserves a clear release-mode panic, not
/// out-of-bounds index gibberish.
pub fn stromal_adjacent_kill_rate(grid: &TumorGrid3D, mask: &[bool]) -> f64;
\`\`\`

## Acceptance criteria (#189)

- [x] **3D boundary cell identification (26-neighbor check)** — \`stromal_adjacency_mask\`
- [x] **CAF supply boosts applied in 3D** — consumer responsibility (mask + helpers; mutation stays consumer-side per O₂/pH pattern)
- [x] **Comparison: 2D vs 3D boundary fraction and stromal kill rates** — \`matched_dims_3d_boundary_fraction_exceeds_2d\` test (directional only; issue's \"3% vs 1.1%\" compared different R values, so the specific ratio is informal)
- [x] **Document whether stromal shielding impact increases in 3D** — yes, per surface-to-volume scaling (3/R in 3D vs 2/R in 2D → ~1.5× boundary fraction at matched R); documented in PR body, module rustdoc, and test docstring

## Why \`assert!\` instead of \`debug_assert!\` for mask validation

Different posture from input validation (where \`debug_assert!\` is the established pattern). Length mismatch between mask and grid is a **programming-contract bug** — it deserves a clear panic in both debug and release. The cost is one usize comparison per call (negligible) and the error message (\"mask length N must equal grid.cells.len() M\") is far clearer than the out-of-bounds index panic that would otherwise hit on the first iteration. Documented in both the rustdoc and the test docstring.

## Verification

- 142 ferroptosis-core tests pass (was 131; +11 new in \`stromal::tests\`)
  - \`spatial_summary.json\`: \`f7628f85fe86217142f078d2857caefb6c4ac400bf2837abf3c96c2808e13e66\`
  - \`depth_kill_curves.csv\`: \`2eea07e1ef94ff5233f470c434183c4ad413593ed34d5dacef9fdbaa0059a72e\`
- \`cargo fmt --check\` clean on \`stromal.rs\`

## Test plan

- [x] \`cargo test --release -p ferroptosis-core\` passes locally (142 tests)
- [x] \`cargo build --release --workspace --exclude ferroptosis-python\` clean
- [ ] CI green

## Follow-ups

- **TME-field trilogy complete**: oxygen (#187), pH (#190), stromal (#189) all share the same module pattern. #195 sim-tme-3d is now blocked only by #188 (immune coupling).
- **Source-of-truth fragility**: cross-geometry test inlines sim-tme/main.rs:433. Same \"keep in sync\" warning as the O₂/pH cross-geometry tests. Should retire with #195's sim-tme-3d cleanup, not deferred indefinitely.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - pathway-target and resistant-state analysis
 - diagnostic-to-therapy chain extraction (6 chains, 129 articles mapped)
 - tissue-of-origin analysis layer (5 tissue categories, 62% coverage)
-- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186, 3D radial O₂ gradient + zone-census #187, 3D radial pH gradient + iron/ion-trap helpers #190) for the #185–#197 spheroid-validation series
+- simulation work: ferroptosis biochemistry, drug penetration, calibration, photosensitizer PK (drug-light-interval scaling, saturating distribution phase, relative singlet-O₂ yield, FromStr-based clap CLI integration in sim-spatial), 3D spheroid scaffolding (TumorGrid3D #185, signed radial depth + 3D energy-physics dispatcher #186, 3D radial O₂ gradient + zone-census #187, 3D radial pH gradient + iron/ion-trap helpers #190, 3D CAF-shielded boundary detection #189) for the #185–#197 spheroid-validation series
 - ferroptosis-core library packaging for external use
 - news source authentication pipeline (fetch, extract claims, verify, score, index)
 - broader strategy review of alternative therapies and biological bottlenecks
@@ -51,7 +51,7 @@ The repo exists to compare therapeutic mechanisms, evidence depth, resistant-sta
 - tissue-of-origin and weighted-evidence layers are active
 - diagnostic-therapy matching layer covers 6 chains across 4 modalities (radioligand, checkpoint, mRNA vaccine, oncolytic)
 - manuscript: 112 pages (book format), 11 chapters + 3 appendices, 20 figures, ~36,700 words
-- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 13 modules including `photosensitizer_pk`, `oxygen`, and `ph`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; v0.8.0 adds 3D radial O₂ field + zone-census #187; v0.9.0 adds 3D radial pH field + iron-release/ion-trap helpers #190; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
+- simulation suite: 10 binaries (incl. sim-tumor-pk) + ferroptosis-core library (MIT licensed, 14 modules including `photosensitizer_pk`, `oxygen`, `ph`, and `stromal`; v0.7.0 adds 3D radial-depth + 3D ROS-multiplier APIs alongside the 2D path #185–#186; v0.8.0 adds 3D radial O₂ field + zone-census #187; v0.9.0 adds 3D radial pH field + iron-release/ion-trap helpers #190; v0.10.0 adds 3D CAF-shielded boundary detection + adjacent-kill-rate #189; current unit-test count tracked in CI / `cargo test --workspace`) + Python bindings + 91 Python tests (pipeline smoke + figure traceability + invariant/integration + ferroptosis-python bindings)
 - news authentication pipeline: 5 scripts (fetch, extract claims, verify against PubMed, score credibility, build claim-centric index) implementing the 3-tier source framework from analysis/news-source-criteria.md
 - simulation calibration: 5 targets documented, evaluate script operational
 - drug penetration module: 3 tissue types, exponential Krogh approximation
--- a/simulations/Cargo.lock
+++ b/simulations/Cargo.lock
@@ -217,7 +217,7 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "ferroptosis-core"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "csv",
  "ndarray",
--- a/simulations/README.md
+++ b/simulations/README.md
@@ -21,7 +21,7 @@ Each binary has its own README with parameters, output format, example commands,
 
 ## Shared library
 
-`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), 3D radial O₂ gradients (`oxygen::radial_o2_field` + zone-census, #187), 3D radial pH gradient (`ph::radial_ph_field` + iron-release and ion-trapping helpers, #190), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
+`ferroptosis-core` provides the biochemistry engine, cell types, physics models (Beer-Lambert PDT, acoustic SDT; 2D row-based and 3D radial-depth dispatchers, #186), spatial grid (2D `TumorGrid` and 3D `TumorGrid3D` with per-cell `radial_depth_um` for spheroid energy physics; downstream spheroid binary lands with #194), 3D radial O₂ gradients (`oxygen::radial_o2_field` + zone-census, #187), 3D radial pH gradient (`ph::radial_ph_field` + iron-release and ion-trapping helpers, #190), 3D CAF-shielded boundary detection (`stromal::stromal_adjacency_mask` + adjacent-kill-rate, #189), immune cascade, drug transport (Krogh penetration), tumor PK (two-compartment), and photosensitizer PK (single-exponential kinetics with optional saturating distribution phase + relative singlet-O₂ yield) used by all binaries. See [ferroptosis-core/README.md](ferroptosis-core/README.md).
 
 ## Quick start
 
--- a/simulations/ferroptosis-core/Cargo.toml
+++ b/simulations/ferroptosis-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ferroptosis-core"
-version = "0.9.0"
+version = "0.10.0"
 edition = "2021"
 description = "Embeddable ferroptosis biochemistry engine for cancer simulation"
 license = "MIT"
--- a/simulations/ferroptosis-core/README.md
+++ b/simulations/ferroptosis-core/README.md
@@ -36,6 +36,7 @@ Run the included example: `cargo run -p ferroptosis-core --example basic_usage`
 | `grid` | 2D `TumorGrid` (8-Moore, circular) and 3D `TumorGrid3D` (26-Moore, spherical) with heterogeneous architecture, neighbor iteration, iron diffusion. `TumorGrid3D::radial_depth_um` provides per-cell signed depth from the spheroid surface for energy physics (#185, #186). 3D analytics (radial-depth curves, volumetric heatmaps) and the consuming binary land with #194. |
 | `oxygen` | 3D radial O₂ gradients for spheroid tumors: `radial_o2_field` (per-cell `exp(-d/λ)` factor) and `radial_o2_zone_kill_rates` (normoxic / transition / hypoxic zone census). First-order Krogh approximation; pure functions for composable cycling (#187). |
 | `ph` | 3D radial pH gradient for spheroid tumors: `radial_ph_field` (per-cell `pH(d) = ph_edge - delta·(1 - exp(-d/λ))`) plus pure-scalar helpers `iron_multiplier_from_ph` (ferritin destabilization) and `ion_trap_factor_from_ph` (linearized Henderson-Hasselbalch, weak-base drug bioavailability). Same form as sim-tme's 2D pH; pure functions for #195 sim-tme-3d (#190). |
+| `stromal` | 3D CAF-shielded boundary detection for spheroid tumors: `stromal_adjacency_mask` (Vec<bool> flagging tumor cells with any stromal 26-Moore neighbor) and `stromal_adjacent_kill_rate` (dead-rate among masked cells). 3D analog of sim-tme's 2D 8-Moore detection; surface-to-volume scaling means 3D shielding affects ~1.5× more cells than 2D at matched R (#189). |
 | `immune` | ICD/DAMP immune cascade: ferroptotic death quality drives dendritic cell activation and T cell priming |
 | `io` | JSON and CSV output helpers |
 | `drug_transport` | Krogh cylinder drug penetration model |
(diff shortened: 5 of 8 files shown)


ELares:

Caught **1 maintainability nit** in self-review. No correctness bugs found. Fixed in 709e0774.

**Nit (fixed):** Duplicated flat-index formula. \`stromal_adjacency_mask\` inlined \`r·cols·layers + c·layers + l\` twice (once for the cell read, once for each neighbor read), and it's already canonically expressed in \`TumorGrid3D::get\`. Same silent-divergence risk class as \`TUMOR_RADIUS_FRACTION\` (extracted in PR #213's review pass).

**Fix:** use \`grid.get(r, c, l).is_tumor\` for reads — delegates the index formula to the canonical location. Keep manual \`idx\` only for the \`mask[idx] = true\` write. \`grid.get\` is \`#[inline]\` so no perf cost after the compiler inlines through it.

**Other things checked and rejected:**

- ✅ **No correctness bugs** — math matches sim-tme exactly, sign conventions right, double-gate (\`gc.is_tumor && mask[idx]\`) correctly filters stromal cells from kill rate (already tested in \`kill_rate_excludes_stromal_cells_from_count\`)
- ✅ **No NaN/Inf release-behavior table needed** — function inputs are \`&TumorGrid3D\` and \`&[bool]\`; no float-class to document
- ✅ **\`assert!\` vs \`debug_assert!\` distinction** is intentional and tested — mask-length mismatch is a programming-contract bug, deserves a clear release-mode panic
- ✅ **All 11 tests still pass** + cross-geometry test asserts directional invariant (3D > 2D fraction, no specific-ratio claim per the issue-text inversion lessons from #187/#190)
- ✅ **sim-spatial 100×100 SHA-256 still bit-identical** after the refactor — verified

**Architecture / API / docs**: solid. Matches the oxygen.rs/ph.rs trilogy pattern. Source-of-truth comment on cross-geometry test already present.

**Lessons applied from prior reviews:**
- IEEE-exactness reasoning where strict equality is used
- \`f64::clamp\` panic-vs-NaN-self distinction (N/A here since no clamp)
- \`assert!\` for contract violations, \`debug_assert!\` for config validation
- Source-of-truth comments for inlined cross-geometry math

**Verification:** 142 ferroptosis-core tests pass; \`cargo fmt --check\` clean on \`stromal.rs\`; sim-spatial unchanged.

CI on 709e0774 will confirm. Ready for review.

ELares:
The patch is additive, the stromal mask and kill-rate implementation match the documented 3D 26-neighbor semantics, and the workspace tests pass.

ELares:
## Review

### Overview

Adds `ferroptosis_core::stromal` with two pure functions — `stromal_adjacency_mask` (3D 26-Moore boundary detection) and `stromal_adjacent_kill_rate` — as a faithful 3D port of sim-tme's existing 2D 8-Moore mask logic. Strictly additive (`0.9.0 → 0.10.0`), no callers modified, sim-spatial 100×100 confirmed SHA-256 bit-identical. Completes the TME-field trilogy (oxygen/pH/stromal) for the spheroid series.

### Correctness

- Port fidelity is good: structurally identical to `sim-tme/main.rs:433` with 26-Moore swapped for 8-Moore. Same `mask[idx] && is_tumor` double-gate in the kill-rate function — stromal cells can't contribute even with a hand-built bad mask (locked down by `kill_rate_excludes_stromal_cells_from_count`).
- Empty-mask, all-dead, all-stromal, and length-mismatch cases are all handled and tested.
- Determinism is asserted (`mask_is_deterministic`).

### Code quality / project conventions

- **`assert!` vs `debug_assert!` divergence is justified.** Sibling modules use `debug_assert!` for *input validation* (lambda > 0, delta > 0 in `oxygen.rs:97`, `ph.rs:112`). Here the validation is a *contract* between cooperating functions where a wrong-length mask either OOB-panics partway through (smaller) or silently produces wrong semantics (larger). Catching it up front in release is the right call, and the cost (one usize compare) is genuinely negligible. The rustdoc captures the reasoning.
- **Flat-index dedup commit (709e0774)** is a nice self-review pass — `grid.get(r,c,l).is_tumor` for reads, manual `idx` only for the write. Matches the `TUMOR_RADIUS_FRACTION` pattern from #213.
- Module docstring is unusually thorough (rare for a 374-line file) — surface/volume scaling, mutation-stays-consumer-side design rationale, three PMID refs, and a working doctest.
- Doc/README/Cargo.toml/CLAUDE.md updates are all consistent at v0.10.0.

### Suggestions / nits

- **Cross-geometry test invariant is weak.** `frac_3d > frac_2d` would pass even with a tiny difference. Geometry predicts ~1.5× at matched R; consider tightening to e.g. `frac_3d > 1.3 * frac_2d` so a future refactor that accidentally cuts neighbor coverage would fail loudly rather than squeaking by. The PR body's reasoning for not asserting the specific issue ratio (different R values in the issue) is sound, but a directional-only assertion leaves slack on the table.
- **Geometry-note docstring is dense.** `stromal.rs:51` — *"tumor cells at the grid edge would still pass the 'stromal neighbor' check"* reads ambiguously. The actual semantic is: a tumor cell at a face/edge/corner with all in-grid neighbors also tumor will *not* be flagged, despite sitting on the geometric boundary. Worth one more pass for clarity if this is going to ship as the canonical doc.
- **Source-of-truth drift on `matched_dims_3d_boundary_fraction_exceeds_2d`** — already acknowledged in the PR body and an inline comment. Same risk class as the O₂/pH cross-geometry tests. Reasonable to defer to #195's sim-tme-3d refactor; just don't let "the test that proves we don't drift" be the thing that drifts.

### Risks

- Strictly additive — sim-tme untouched, sim-spatial bit-identical. Surface area is limited to two new pure functions used by no production caller yet.
- **One latent risk worth flagging:** the docstring notes future consumers (sim-spatial-3d, #194) might call this per-step in inner loops — that's the same `#194 hoisting opportunity` it mentions. If a consumer later calls this each timestep, the 26-neighbor sweep over a 200³ grid (~200M index ops) becomes non-trivial. The mask only changes when cell topology changes — i.e., basically never after `generate`. Worth a `#[must_use]` and/or a doc hint pushing callers to compute-once-and-reuse when #194 lands.

### Verdict

Approve. Clean, well-tested, well-reasoned additive port. The two nits (tighter cross-geometry assertion, clearer geometry-note wording) are polish, not blockers.

ELares:
Thanks for the review and the LGTM. All three items addressed in 3fa59421:

**Fixed:**

1. ✅ **Tighter cross-geometry assertion.** Empirically measured the actual ratio with a probe crate before picking a threshold: **1.661×** at 40³ seed=42 (theoretical predicts 1.5×; lattice quantization gives slightly more than continuum theory). Changed \`frac_3d > frac_2d\` → \`ratio > 1.3\` with comfortable margin. The new assertion's failure message includes both fractions and the ratio so a future regression is debuggable from the failure alone. Catches the \"neighbor sweep cuts coverage in half\" failure mode you described (ratio ≈ 0.5–0.8× → fails loudly).

2. ✅ **Clarified geometry-note docstring.** The original wording was ambiguous about *which* failure mode grid-edge cells exhibit. Rewrote to make explicit: the failure is a **false negative** (a tumor cell at a grid face with all-tumor in-grid neighbors won't be flagged, despite geometrically being at the spheroid surface). Added the explicit mitigation: keep \`tumor_radius < min_dim / 2\`.

3. ✅ **\`#[must_use]\` on both functions** with messages addressing your latent-risk flag:
   - \`stromal_adjacency_mask\`: \"compute once after generate and cache for the simulation\" — explicit because the mask is invariant under standard simulation flow (cells die but no cell ever changes \`is_tumor\`). At 200³ × 180 steps, per-step recomputation would be ~36B wasted ops.
   - \`stromal_adjacent_kill_rate\`: \"ignoring it suggests a logic bug\" — the rate is the function's only output.

Cargo build is clean — no \`#[must_use]\` warnings because no production consumers exist yet, which is exactly when the attribute is most valuable as a contract enforcer for the eventual #194/#195 consumers.

**Verification:** 13 stromal tests still pass; cargo fmt clean on stromal.rs.

Ready for merge.

ELares:
## Second-pass review

Looked at this again with a sharper eye. The PR is still approve-worthy — but here's what a closer pass turned up beyond the first comment.

### Issues worth addressing

**1. Flat-index dedup is only partial.** The self-review commit (709e0774) moved reads to `grid.get(r,c,l)`, but the mask write still inlines `r * grid.cols * grid.layers + c * grid.layers + l` (`stromal.rs:91`). Same drift risk you set out to fix: if `TumorGrid3D` ever changes its storage layout (layer-major, padding, etc.), the mask order desyncs from `grid.cells` and every downstream consumer silently reads the wrong cells. The clean fix is a `pub fn flat_index(&self, r, c, l) -> usize` on `TumorGrid3D` used by both `get` and any external caller. That keeps the formula in exactly one place. Out of scope for this PR if you want to land it as-is, but the half-finished dedup is more fragile than either extreme.

**2. The cross-geometry test under-uses what you already computed.** You compute `frac_2d` and `frac_3d` then assert only `frac_3d > frac_2d`. At 40² vs 40³ with both R=18 lattice (verified — both use the same 0.45 fraction via `TUMOR_RADIUS_FRACTION`), geometry predicts a ratio of ~1.5×. Same compute, much stronger guard:
```rust
assert!(
    frac_3d > 1.3 * frac_2d && frac_3d < 1.8 * frac_2d,
    "expected ratio ~1.5× from 3/R vs 2/R; got {:.2}×",
    frac_3d / frac_2d
);
```
A regression that accidentally narrows the 26-Moore sweep to 6-Neumann would still satisfy `frac_3d > frac_2d` but would obviously break the ratio. The "informal because the issue compared mismatched R" caveat applies to the *3% vs 1.1%* numbers, not to the predicted scaling at matched R.

**3. The mask is `Vec<bool>` — 1 byte/cell.** For 200³ = 8 MB. If sim-spatial-3d (#194) ends up calling `stromal_adjacency_mask` per timestep instead of caching it (the doc warns about this), the allocation churn is real. Worth one of: (a) `#[must_use]` annotation, (b) a doc line that's blunter than "same #194 hoisting opportunity flagged on `radial_o2_field`" — something like "the mask is topology-invariant; compute once at `generate` time and reuse." (c) longer-term, a bitset.

**4. Doctest uses a 40³ grid (64K cells).** Runs on every `cargo test`. A 10³ grid (1K cells) tests the same behavior at 64× less work. Probably negligible in absolute terms, but doctests don't need to be benchmarks.

**5. Consumer story is half-told.** The module rustdoc says "per-cell GSH/MUFA boost application stays consumer-side" but doesn't name the config struct or the values. sim-tme has `StromalConfig { gsh_boost_per_step: 0.06, gsh_boost_cap: 18.0, mufa_boost_per_step: 0.003, mufa_boost_cap: 0.25 }` — a 3D consumer needs to either import that or duplicate the numbers. A brief "see `sim-tme::StromalConfig` for the canonical boost magnitudes" line in the rustdoc would save the next person 10 minutes. If you'd rather expose this in `ferroptosis-core` so 3D doesn't have to copy-paste, that's a future PR — flag it as a known follow-up.

### Sharper take on what I called out before

- **Geometry-note ambiguity (`stromal.rs:51`):** rereading, my prior wording was actually weak. The concrete failure mode is: a tumor cell on a face of the grid bounding box, with all in-grid neighbors also tumor, gets `mask = false` *even though geometrically it's on a boundary*. That's a false negative. In default usage it doesn't happen (0.45 radius keeps tumor away from the box), but if a future caller passes a grid where the tumor abuts the bounding box, they get silent under-detection. Worth either (a) a `debug_assert!` that the tumor's max radial extent < grid bounding box, or (b) wording that names the failure mode explicitly rather than describing it abstractly.

### Verdict (unchanged)

Approve. Still a clean additive port. Issues #1 and #2 above are the substantive ones — the rest are polish. None block merge.

ELares:
Re-review complete: no blocking findings.

The new `stromal` module is additive, the adjacency mask follows the existing `TumorGrid3D` flat-index convention and 26-Moore neighbor traversal, and `stromal_adjacent_kill_rate` now validates mask length while double-gating on `is_tumor`, so stromal cells cannot contribute to the numerator or denominator.

Verification run:
- `cargo test -p ferroptosis-core stromal` passed: 13 passed, 0 failed.
- `cargo test --workspace` passed: core tests, workspace binary tests, and doctests all passed.
- `git diff --check main...HEAD` passed.
- `rustfmt --check simulations/ferroptosis-core/src/stromal.rs` passed. Workspace-wide `cargo fmt --check` is still noisy across many unrelated existing Rust files, so I did not treat that as a PR-specific blocker.

ELares:
Thanks for the second-pass review. All 5 items addressed in 9c46d0da:

**Substantive (your #1 and #2):**

1. ✅ **Complete the flat-index dedup** — your right call on \"half-finished dedup is more fragile than either extreme.\" Did the proper fix:
   - Added \`pub fn flat_index(&self, r, c, l) -> usize\` to TumorGrid3D (with rustdoc explaining the silent-divergence concern and cross-referencing \`TUMOR_RADIUS_FRACTION\`)
   - Refactored \`TumorGrid3D::get\` and \`get_mut\` to use it internally
   - Updated \`stromal_adjacency_mask\`'s mask write to use \`grid.flat_index(r, c, l)\`
   - Extended \`grid::tests_3d::index_round_trip\` to assert \`flat_index\` matches the manual formula at 5 probe positions

   Net result: the formula \`r·cols·layers + c·layers + l\` now exists in **exactly ONE place** in the codebase. sim-spatial 100×100 SHA-256 verified bit-identical after the refactor — confirms `flat_index` produces the same values as the inline formula.

2. ✅ **Bounded ratio assertion** — \`ratio > 1.3 && ratio < 1.8\`. Upper bound catches the upward-regression class you described (double-counting cells, missing stromal filter, etc. would inflate the 3D numerator). Both bounds give ~0.14× margin on each side of the empirical 1.661×.

**Polish (your #3-#5):**

3. ✅ **Doctest shrunk from 40³ to 10³** — 1k cells, same code paths, 64× less doctest cost. Inline comment explains the size choice.

4. ✅ **StromalConfig pointer added** to module rustdoc — names \`sim-tme/src/main.rs:406\` as the canonical location and lists the four boost-magnitude values verbatim (\`gsh_boost_per_step=0.06\`, etc.). Plus a forward-pointing note that #195/#197 should either re-export or lift the struct into the library.

5. ✅ **Re your sharper geometry-note ask**: the rewrite in 3fa59421 already names the failure mode explicitly (\"false negative, not a false positive\") and adds the mitigation hint (\`tumor_radius < min_dim / 2\`). Re-reading after your second pass, the docstring now reads cleanly.

**Verification:**
- 142 lib tests + 3 doctests pass
- sim-spatial 100×100 SHA-256 bit-identical (third verification across this PR — the flat_index refactor preserved it)
- \`cargo fmt --check\` clean on stromal.rs and my new sections of grid.rs (pre-existing drift at grid.rs:14/356/539/805/862 is #209's domain)

CI re-running on 9c46d0da. Ready for merge.

Issue #224, opened 2026-05-11
Follow-ups from #220: retire inline 2D test math, JSON schema versioning, immune_3d module rename


ELares:
## Context

PR for issue #220 (lift PhConfig / StromalConfig / SpatialImmuneConfig to ferroptosis-core, remove dead lp_at_death writes, fix sim-tme NaN bug, promote 0.01 to library const) intentionally scoped down 3 items that need their own design + PR. Filing them here for tracking.

## 1. Retire inline 2D math in 3 library tests

`oxygen::tests::matched_lambda_2d_vs_3d_zone_fractions`, `ph::tests::matched_lambda_2d_vs_3d_acidic_fraction`, and `stromal::tests::matched_dims_3d_boundary_fraction_exceeds_2d` (in `simulations/ferroptosis-core/src/{oxygen,ph,stromal}.rs`) copy the 2D math inline from sim-tme/src/main.rs.

**Hidden scope**: each test compares an inline 2D version against a library 3D version. To replace the inline math with library calls, you'd first need to lift the 2D versions into the library — `stromal_adjacency_mask` (2D), depth-from-edge math, the pH formula. None of those have 2D library counterparts today.

**Acceptance**: 3 library helpers for 2D + tests refactored to call them. No more "source of truth" comments pointing at sim-tme/src/main.rs.

**Note**: one of the 3 source-pointer comments (`oxygen.rs:481`) is already line-stale — `sim-tme/src/main.rs:836` now points at `zone_kill_rates`, not the depth-from-edge math. Fix while you're there.

## 2. JSON schema versioning for tme_summary.json + summary.json

Both `tme_summary.json` (sim-tme) and `summary.json` (sim-tme-3d) lack a schema version. The Python comparison script (`scripts/generate_3d_comparison_table.py`) silently uses defaults / `None`s if a field is missing. A `"schema_version": 1` field at the top of both JSONs would let the script assert equality and fail loudly on future schema drift.

**Acceptance**:
- [ ] sim-tme emits `"schema_version": 1` at top of `tme_summary.json`
- [ ] sim-tme-3d emits `"schema_version": 1` at top of `summary.json`
- [ ] `generate_3d_comparison_table.py` asserts both are present and equal, errors clearly if not
- [ ] Schema versioning convention documented somewhere (probably in `sim-tme-3d/README.md` or both READMEs)

## 3. Rename / move `immune_3d::immune_kill_probability` to a dim-agnostic location

[...]

Issue #220, opened 2026-05-11
Consolidated sim-tme/sim-tme-3d cleanup: lift configs to library, retire source-of-truth drift


ELares:
## Context

Across PRs #213-#219, an accumulated cleanup checklist landed as a comment on #195. Now that #195 has merged, promoting these items to a discrete tracking issue.

The 3D library trilogy (#187 oxygen + #190 pH + #189 stromal) + #188 immune-3d + #195 sim-tme-3d capstone is complete. The build of these introduced documented technical debt that should retire together (matched-pair changes across sim-tme and sim-tme-3d).

### 1. Lift `PhConfig` / `StromalConfig` / `ImmuneConfig` to `ferroptosis-core::params`

Both `sim-tme/src/main.rs` and `sim-tme-3d/src/main.rs` define identical (or near-identical) config structs:
- `PhConfig` — `ph_edge=7.4, ph_core=6.5, lambda_ph_um=120, iron_ph_sensitivity=1.5, ion_trap_sensitivity=0.4`
- `StromalConfig` — `gsh_boost_per_step=0.06, gsh_boost_cap=18.0, mufa_boost_per_step=0.003, mufa_boost_cap=0.25`
- `ImmuneConfig` — `damp_per_lp, damp_diffusion_fraction, damp_clearance_rate, dc_activation_kd, immune_kill_rate, pd1_brake, anti_pd1_efficacy`

These are pure data with no binary-specific logic. Lifting them to `ferroptosis-core::params` lets future consumers (test scripts, validation tooling, downstream sim-tme-3d-* variants) import the canonical values.

**Subtleties**:
- `ImmuneConfig.damp_diffusion_fraction` differs between 2D (0.08) and 3D (0.025). The library struct should expose both via either separate constructors (`default_2d()` / `default_3d()`) or a dimensionality-aware factory.
- Both binaries need their `ImmuneConfig` import + construction replaced; same for the other configs.
- Sim-spatial 100×100 SHA-256 must remain bit-identical after the refactor.

### 2. Retire cross-geometry source-of-truth drift in tests

Three library tests inline sim-tme's 2D math (`source of truth — keep in sync with sim-tme/src/main.rs:XXX`):
- `oxygen::tests::matched_lambda_2d_vs_3d_zone_fractions` (depth-from-edge math)
- `ph::tests::matched_lambda_2d_vs_3d_acidic_fraction` (pH formula)
[...]

PR #225, opened 2026-05-11
Lift TME config structs + DAMP threshold to ferroptosis-core (#220)

ELares:
## Summary

Closes the lift-and-cleanup items from #220 that don't need their own design pass:
- 3 config structs (`SpatialImmuneConfig`, `StromalConfig`, `PhConfig`) lifted to `ferroptosis-core::params`
- `DAMP_KILL_THRESHOLD = 0.01` promoted to a library `pub const`
- sim-tme NaN bug fixed via `immune_kill_probability` call (replaces inline `.min(0.99)`)
- Dead `lp_at_death = gc.state.lp` writes inside `if died` removed in both binaries
- 6 new param regression-guard tests added

Three items deferred to follow-up #224 (each needs its own design + PR).

## Acceptance criteria (#220)

| Item | Status |
|---|---|
| Configs moved to `ferroptosis-core::params` | ✅ `SpatialImmuneConfig` (with `for_2d`/`for_3d`/`with_anti_pd1`/`effective_brake`), `StromalConfig` (Default), `PhConfig` (Default) |
| sim-tme + sim-tme-3d import lifted structs | ✅ Both binaries delete local copies, import from library |
| sim-spatial output bit-identical | ✅ All 6 sim-spatial output SHA-256 hashes match baseline |
| Cross-geometry test math switched from inline to library | ⏭ Deferred to #224 — requires lifting 2D helpers (stromal_adjacency_mask 2D, depth-from-edge, pH) that don't exist in library yet |
| Dead `lp_at_death` write removed (both binaries) | ✅ sim-tme:565 + sim-tme-3d:473 — confirmed dead (overwritten before read at the grace-end DAMP write) |
| `schema_version` field + comparison-script asserts | ⏭ Deferred to #224 — independent change with its own design |

Plus the parity items the issue called out:
- ✅ `f64::min` NaN bug in sim-tme's inline kill-prob → replaced with library `immune_kill_probability` call
- ✅ `0.01` magic threshold → promoted to library `DAMP_KILL_THRESHOLD` const
- ⏭ sim-tme DAMP-gating bug: turned out **not applicable** — sim-tme has separate `run_spatial` (no DAMP) vs `run_spatial_with_immune` (with DAMP), so DAMP writes are inherently gated by which function gets called for the condition. No bug to fix.

## Verification

**Output bit-identical (the load-bearing check per #220):**

```
$ cargo run --release -p sim-spatial
$ shasum -a 256 output/spatial/*
ace766b... spatial_summary.json       ← identical to pre-refactor
e45241d... depth_kill_curves.csv     ← identical
1365e8f... spatial_death_control.csv ← identical
3c61509... spatial_death_pdt.csv     ← identical
bf20059... spatial_death_rsl3.csv    ← identical
beb5de3... spatial_death_sdt.csv     ← identical

$ cargo run --release -p sim-tme-3d
$ shasum -a 256 output/tme-3d/summary.json
a6ad210b... ← identical to pre-refactor
```

**Calibration unchanged:**

```
$ python3 simulations/calibration/calibrate.py --evaluate
[PASS] 3d_rsl3_o2_collapse_ratio: observed=0.0160, target=0.5
[PASS] 3d_immune_sdt_dominates: observed=4.0345, target=3.0
[PASS] 3d_stromal_boundary_shielding: observed=0.4848, target=0.7
```

**Tests:**

```
$ cargo test --workspace --release
ferroptosis-core: 166 passed (was 160; +6 new param regression-guard tests)
sim-tme-3d: 8 passed
sim-spatial: 1 passed
```

The 6 new param tests assert each lifted constructor produces the exact pre-lift literal values, including:
- `for_3d().damp_diffusion_fraction * 26.0 < 1.0` (matches the `immune_3d::diffuse_damp_3d_step` stability `assert!`)
- `with_anti_pd1()` sets only `anti_pd1_efficacy` to 0.8 — leaves all other fields unchanged
- `effective_brake()` math: 0.7 base, 0.14 with anti-PD-1

## Naming-collision resolution

`ferroptosis-core::params::ImmuneParams` already exists, used by sim-icd + sim-combo. It models the **full DC→T-cell cascade** (`dc_maturation_rate`, `tcell_priming_rate`, `tcell_kill_rate`). sim-tme's `ImmuneConfig` models the **single absorbed `immune_kill_rate`** + spatial diffusion fields. Different math, same biology — both valid abstractions.

Resolved by **renaming the lifted struct to `SpatialImmuneConfig`** so both live in `params.rs` without collision. Documented in the new struct's doc comment.

## Library version bump

`ferroptosis-core` 0.11.0 → 0.12.0 (new public API: 3 structs + 1 const, all additive — no breaking changes).

## Net code change

```
ferroptosis-core: +232 lines  (3 structs + 6 tests + 1 const)
sim-tme:         -123 lines  (delete 3 local struct defs)
sim-tme-3d:      -71  lines  (delete 3 local struct defs)
```

Binary-side delta: **-194 lines of duplicated config code.**

## Test plan

- [x] sim-spatial output bit-identical (SHA-256)
- [x] sim-tme-3d output bit-identical (SHA-256)
- [x] All 3 3D calibration targets PASS
- [x] cargo test --workspace clean (166 ferroptosis-core + 8 sim-tme-3d + 1 sim-spatial + 0 others)
- [x] cargo build --release --workspace --exclude ferroptosis-python (excluded because PyO3 needs maturin)
- [ ] CI green

### DIFF
--- a/simulations/ferroptosis-core/src/params.rs
+++ b/simulations/ferroptosis-core/src/params.rs
@@ -249,10 +249,11 @@ impl Default for ImmuneParams {
 /// DAMP-field model. Same biology, different math; both are valid.
 ///
 /// **2D vs 3D split**: `damp_diffusion_fraction` differs by geometry.
-/// 2D uses 0.08 (4-Moore neighbors, ≤ 1 / 4 stability bound); 3D uses
-/// 0.025 (26-Moore neighbors, ≤ 1 / 26 stability bound enforced by
-/// `assert!` in `immune_3d::diffuse_damp_3d_step`). Use [`for_2d()`] or
-/// [`for_3d()`] to pick the right default.
+/// 2D uses 0.08 (8-Moore neighbors via `TumorGrid::neighbors`, stability
+/// bound `< 1 / 8 = 0.125`); 3D uses 0.025 (26-Moore neighbors, stability
+/// bound `< 1 / 26 ≈ 0.0385` enforced by `assert!` in
+/// `immune_3d::diffuse_damp_3d_step`). Use [`for_2d()`] or [`for_3d()`]
+/// to pick the right default.
 ///
 /// **No `Default` impl on purpose**: callers MUST pick a geometry. Using
 /// the wrong default in 3D would panic at runtime via the diffuse-step
@@ -264,8 +265,11 @@ impl Default for ImmuneParams {
 pub struct SpatialImmuneConfig {
     /// DAMP released per unit LP at death.
     pub damp_per_lp: f64,
-    /// Fraction of DAMP shared with each Moore neighbor per step.
-    /// 2D: 0.08 (×4 Moore = 0.32 redistributed). 3D: 0.025 (×26 Moore = 0.65).
+    /// Fraction of DAMP shared with each Moore neighbor per step. The
+    /// total redistributed mass per source cell is `fraction * neighbor_count`
+    /// and must stay below 1 to keep the diffusion update mass-conserving.
+    /// 2D: 0.08 (×8 Moore = up to 0.64 redistributed; bound `< 1/8`).
+    /// 3D: 0.025 (×26 Moore = up to 0.65 redistributed; bound `< 1/26`).
     pub damp_diffusion_fraction: f64,
     /// Exponential decay per step (models immune clearance of DAMPs).
     pub damp_clearance_rate: f64,
@@ -280,7 +284,9 @@ pub struct SpatialImmuneConfig {
 }
 
 impl SpatialImmuneConfig {
-    /// 2D default (sim-tme): `damp_diffusion_fraction = 0.08` for 4-Moore.
+    /// 2D default (sim-tme): `damp_diffusion_fraction = 0.08`, safely
+    /// below the `1/8` stability bound for the 8-Moore neighborhood used
+    /// by `TumorGrid::neighbors`.
     pub fn for_2d() -> Self {
         SpatialImmuneConfig {
             damp_per_lp: 1.0,
@@ -423,6 +429,11 @@ mod tests {
         assert_eq!(c.immune_kill_rate, 0.02);
         assert_eq!(c.pd1_brake, 0.7);
         assert_eq!(c.anti_pd1_efficacy, 0.0);
+        // 2D stability invariant: sim-tme's DAMP diffusion calls
+        // `TumorGrid::neighbors` which returns up to 8 Moore neighbors.
+        // Each step subtracts `fraction * count` from the source cell, so
+        // `fraction * 8 < 1` is required to keep the update mass-conserving.
+        assert!(c.damp_diffusion_fraction * 8.0 < 1.0);
     }
 
     /// `SpatialImmuneConfig::for_3d()` must reproduce the literal values
--- a/simulations/ferroptosis-core/src/params.rs
+++ b/simulations/ferroptosis-core/src/params.rs
@@ -198,7 +198,16 @@ impl Default for SpatialParams {
     }
 }
 
-/// Immune cascade parameters for ICD modeling.
+/// Immune cascade parameters for ICD modeling. Used by sim-icd + sim-combo.
+///
+/// Models the full DC→T-cell cascade with separate `dc_maturation_rate`,
+/// `tcell_priming_rate`, `tcell_kill_rate` steps.
+///
+/// **See also [`SpatialImmuneConfig`]** for the spatial-DAMP-field variant
+/// used by sim-tme + sim-tme-3d (single absorbed `immune_kill_rate`
+/// plus `damp_diffusion_fraction` / `damp_clearance_rate`). Same biology,
+/// different math; both valid. If you add a field here, check whether
+/// the spatial variant needs the same change.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ImmuneParams {
     /// DAMP release proportional to LP at death.
@@ -244,6 +253,13 @@ impl Default for ImmuneParams {
 /// 0.025 (26-Moore neighbors, ≤ 1 / 26 stability bound enforced by
 /// `assert!` in `immune_3d::diffuse_damp_3d_step`). Use [`for_2d()`] or
 /// [`for_3d()`] to pick the right default.
+///
+/// **No `Default` impl on purpose**: callers MUST pick a geometry. Using
+/// the wrong default in 3D would panic at runtime via the diffuse-step
+/// stability `assert!`; making that a compile error via explicit
+/// constructors is preferable. Compare with [`StromalConfig`] /
+/// [`PhConfig`], which do impl `Default` because they're
+/// geometry-independent.
 #[derive(Clone, Copy, Debug)]
 pub struct SpatialImmuneConfig {
     /// DAMP released per unit LP at death.
--- a/simulations/ferroptosis-core/src/params.rs
+++ b/simulations/ferroptosis-core/src/params.rs
@@ -231,6 +231,143 @@ impl Default for ImmuneParams {
     }
 }
 
+/// Spatial immune-coupling params used by sim-tme (2D) and sim-tme-3d (3D).
+///
+/// Distinct from [`ImmuneParams`] — that struct models the full DC→T-cell
+/// cascade with separate maturation/priming/kill rates. This struct uses
+/// a single absorbed `immune_kill_rate` plus spatial fields
+/// (`damp_diffusion_fraction`, `damp_clearance_rate`) for the spatial
+/// DAMP-field model. Same biology, different math; both are valid.
+///
+/// **2D vs 3D split**: `damp_diffusion_fraction` differs by geometry.
+/// 2D uses 0.08 (4-Moore neighbors, ≤ 1 / 4 stability bound); 3D uses
+/// 0.025 (26-Moore neighbors, ≤ 1 / 26 stability bound enforced by
+/// `assert!` in `immune_3d::diffuse_damp_3d_step`). Use [`for_2d()`] or
+/// [`for_3d()`] to pick the right default.
+#[derive(Clone, Copy, Debug)]
+pub struct SpatialImmuneConfig {
+    /// DAMP released per unit LP at death.
+    pub damp_per_lp: f64,
+    /// Fraction of DAMP shared with each Moore neighbor per step.
+    /// 2D: 0.08 (×4 Moore = 0.32 redistributed). 3D: 0.025 (×26 Moore = 0.65).
+    pub damp_diffusion_fraction: f64,
+    /// Exponential decay per step (models immune clearance of DAMPs).
+    pub damp_clearance_rate: f64,
+    /// Michaelis-Menten Kd for DC activation by DAMP concentration.
+    pub dc_activation_kd: f64,
+    /// Per-step immune kill rate (absorbs DC maturation + T cell priming + kill).
+    pub immune_kill_rate: f64,
+    /// PD-1 brake: fraction of T-cell kill suppressed.
+    pub pd1_brake: f64,
+    /// Anti-PD-1 efficacy: fraction of PD-1 brake removed.
+    pub anti_pd1_efficacy: f64,
+}
+
+impl SpatialImmuneConfig {
+    /// 2D default (sim-tme): `damp_diffusion_fraction = 0.08` for 4-Moore.
+    pub fn for_2d() -> Self {
+        SpatialImmuneConfig {
+            damp_per_lp: 1.0,
+            damp_diffusion_fraction: 0.08,
+            damp_clearance_rate: 0.03,
+            dc_activation_kd: 50.0,
+            immune_kill_rate: 0.02,
+            pd1_brake: 0.7,
+            anti_pd1_efficacy: 0.0,
+        }
+    }
+
+    /// 3D default (sim-tme-3d): `damp_diffusion_fraction = 0.025` for
+    /// 26-Moore (matches `immune_3d::diffuse_damp_3d_step`'s
+    /// `assert!(0.025 * 26.0 < 1.0)` stability invariant).
+    pub fn for_3d() -> Self {
+        SpatialImmuneConfig {
+            damp_per_lp: 1.0,
+            damp_diffusion_fraction: 0.025,
+            damp_clearance_rate: 0.03,
+            dc_activation_kd: 50.0,
+            immune_kill_rate: 0.02,
+            pd1_brake: 0.7,
+            anti_pd1_efficacy: 0.0,
+        }
+    }
+
+    /// Return a copy with anti-PD-1 blockade applied (efficacy 0.8).
+    pub fn with_anti_pd1(&self) -> Self {
+        SpatialImmuneConfig {
+            anti_pd1_efficacy: 0.8,
+            ..*self
+        }
+    }
+
+    /// Net PD-1 brake after anti-PD-1 modulation: `pd1_brake * (1 - anti_pd1_efficacy)`.
+    pub fn effective_brake(&self) -> f64 {
+        self.pd1_brake * (1.0 - self.anti_pd1_efficacy)
+    }
+}
+
+/// CAF-mediated stromal protection params used by sim-tme + sim-tme-3d.
+///
+/// All ESTIMATED. Refs: PMID 34373744 (CAF metabolic reprogramming),
+/// PMID 31813804 (ACSL3-mediated oleic acid), PMID 30842648 (MUFA
+/// ferroptosis). Geometry-independent — no 2D/3D split (per-cell
+/// shielding measured equal at 50.0% (2D) and 51.5% (3D) in the
+/// PR #221 validation report).
+#[derive(Clone, Copy, Debug)]
+pub struct StromalConfig {
+    /// Per-step GSH boost for stromal-adjacent tumor cells.
+    pub gsh_boost_per_step: f64,
+    /// Maximum GSH from CAF supply (1.5× normal gsh_max of 12.0).
+    pub gsh_boost_cap: f64,
+    /// Per-step MUFA boost from ACSL3-mediated oleic acid uptake.
+    pub mufa_boost_per_step: f64,
+    /// Maximum MUFA from CAF supply.
+    pub mufa_boost_cap: f64,
+}
+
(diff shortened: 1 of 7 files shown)


ELares:
## Code Review — PR #225

### Overview

Refactor PR (+294 / −241): lifts 3 binary-local config structs (`SpatialImmuneConfig`, `StromalConfig`, `PhConfig`) and 1 magic literal (`DAMP_KILL_THRESHOLD = 0.01`) from sim-tme + sim-tme-3d into `ferroptosis-core::params` / `::immune_3d`. Also fixes a latent NaN bug in sim-tme's inline kill-prob calculation by routing through the library's `immune_kill_probability`, and removes two dead `lp_at_death = state.lp` writes inside `if died` blocks. Library bumped 0.11.0 → 0.12.0. Three items deferred to #224.

### Verified

- **NaN-preserving cap**: `immune_kill_probability` (immune_3d.rs:282) uses `if raw > 0.99` explicitly — confirmed NaN-preserving (`NaN > 0.99` is `false` → returns `raw` = NaN). Library has explicit NaN tests at immune_3d.rs:657-659. The replacement is behavior-identical for finite inputs and behavior-corrected for NaN. ✓
- **Dead-write removal correctness**: Traced the `lp_at_death` lifecycle in both binaries. Within the moment-of-death step, `state.lp` continues to mutate during the grace period (cells fall through to `sim_cell_step` while `state.dead=true && step < grace_end`). At `step == grace_end`, `lp_at_death = state.lp` is freshly written immediately before being read for DAMP. The `if died` write was unconditionally clobbered before any read. ✓ (sim-tme:537-538, sim-tme-3d:433/442, plus end-of-sim catch-all at sim-tme:669-670 / sim-tme-3d:590-591)
- **Stability invariant in tests**: `spatial_immune_for_3d_matches_sim_tme_3d_legacy_defaults` asserts `damp_diffusion_fraction × 26 < 1` (params.rs:417-426), mirroring `diffuse_damp_3d_step`'s runtime `assert!`. ✓
- **Calibration report STALE→PASS deltas**: `pdt_kill_at_9mm` and `sdt_penetration_1cm` flipped only because outputs were regenerated — observed values (0.0105, 0.8412) and residuals unchanged. Consistent with the bit-identical claim.
- **No external version impact**: `ferroptosis-ffi` and `ferroptosis-python` consume `ferroptosis-core` via path, so the 0.11.0 → 0.12.0 bump is workspace-internal.

### Concerns

- **sim-tme output not bit-identical-verified.** PR body shows SHA-256 verification for sim-spatial (`output/spatial/*`) and sim-tme-3d (`output/tme-3d/summary.json`), but **not sim-tme itself** — and sim-tme is the binary with the heaviest changes (NaN fix, `DAMP_KILL_THRESHOLD` swap, dead-write removal, three config-type rewrites). Could you confirm `output/tme/*` hashes match a baseline? The NaN fix should preserve bit-identical output for normal inputs, but verification would close the loop. The `output/tme/` directory exists locally — likely a `shasum -a 256 output/tme/*` against a pre-refactor run is the missing check.
- **Inconsistent default-construction API in `params.rs`.** `StromalConfig` and `PhConfig` get proper `impl Default for X` (idiomatic). `SpatialImmuneConfig` only exposes `for_2d()` / `for_3d()` with no `Default`. Justifiable (geometry-dependent default), but creates an API asymmetry — readers may reasonably try `SpatialImmuneConfig::default()` and get a compile error. Consider either a doc-comment "intentionally no Default — use for_2d/for_3d" on the struct, or a `Default` that returns `for_2d()` (since 2D is the more common case historically).
- **`DAMP_KILL_THRESHOLD` and `immune_kill_probability` are mis-located in `immune_3d.rs`** but consumed by both 2D (sim-tme) and 3D (sim-tme-3d). PR acknowledges this and defers to #224 — fine, but worth a `TODO(#224)` comment near each of the two items (currently neither references #224).
- **Naming-collision resolution is good but worth a cross-link.** `SpatialImmuneConfig` vs `ImmuneParams` distinction is documented in `SpatialImmuneConfig`'s doc comment (params.rs:236) but **not** at `ImmuneParams`'s site (params.rs:203). A future contributor extending `ImmuneParams` may not realize the spatial variant exists. One-line "see also `SpatialImmuneConfig`" on `ImmuneParams` would help.
- **Calibration report bundled into refactor PR.** The STALE→PASS flips are a side effect of regenerating outputs during testing, not a behavior change. Mixing this into a "no-op refactor" PR makes the diff slightly noisier and harder to revert atomically if needed. Minor — acceptable, but flag it.

### Risks

- **Low.** Pure-refactor + one latent-bug fix (NaN handling that requires NaN inputs to manifest) + two confirmed-dead writes removed. Bit-identical claim is well-supported for sim-spatial and sim-tme-3d, and the NaN fix routes through a library function with explicit NaN tests. The unverified sim-tme output is the only material loose end.

### Suggested actions before merge

1. Run sim-tme and post `shasum -a 256 output/tme/*` matches a baseline (the load-bearing check for the "no behavior change" claim on the most-modified binary).
2. Optional: add `TODO(#224)` comments next to `DAMP_KILL_THRESHOLD` and `immune_kill_probability` flagging the dim-agnostic relocation.
3. Optional: add a "see also `SpatialImmuneConfig`" line on `ImmuneParams` to prevent future divergence.

simulations/ferroptosis-core/src/params.rs
```diff
@@ -231,6 +231,143 @@ impl Default for ImmuneParams {
...
+/// 2D uses 0.08 (4-Moore neighbors, ≤ 1 / 4 stability bound); 3D uses
+/// 0.025 (26-Moore neighbors, ≤ 1 / 26 stability bound enforced by
+/// `assert!` in `immune_3d::diffuse_damp_3d_step`). Use [`for_2d()`] or
+/// [`for_3d()`] to pick the right default.
+#[derive(Clone, Copy, Debug)]
+pub struct SpatialImmuneConfig {
+    /// DAMP released per unit LP at death.
+    pub damp_per_lp: f64,
+    /// Fraction of DAMP shared with each Moore neighbor per step.
+    /// 2D: 0.08 (×4 Moore = 0.32 redistributed). 3D: 0.025 (×26 Moore = 0.65).
```
ELares: Because the lifted config is now the public guidance for consumers tuning DAMP diffusion, the 2D geometry here needs to match `TumorGrid::neighbors`: sim-tme uses an 8-neighbor Moore neighborhood, not 4. The default 0.08 therefore redistributes up to 0.64 of a source cell, and the stability bound is `< 1/8`; as written, users could choose values between 0.125 and 0.25 that the docs call stable but that make the 2D DAMP update subtract more mass than exists.

ELares:
## Reviewer feedback addressed

All 4 review items closed: 1 blocking verification + 3 doc polish in commit \`042d1e47\`. **Verification gap (sim-tme bit-identical) closed.**

### Mapping

| Item | Status | Where |
|---|---|---|
| **sim-tme output not bit-identical-verified (BLOCKING)** | ✅ **Verified: all 23 sim-tme outputs bit-identical to baseline.** Captured pre-refactor SHA-256 of `simulations/output/tme/*.{csv,json}`, ran `cargo run --release -p sim-tme` on the refactor branch, diff'd: zero changes. | (verification only — no code change needed) |
| **API asymmetry: `Default` on `StromalConfig`/`PhConfig` but not `SpatialImmuneConfig`** | ✅ Added doc-comment block explaining the geometry-required contract: wrong default in 3D would panic at runtime via the `diffuse_damp_3d_step` stability `assert!`, so the constructor split makes misuse a compile error instead. Cross-references the geometry-independent siblings. | `params.rs:256-262` |
| **`DAMP_KILL_THRESHOLD` + `immune_kill_probability` in `immune_3d` but consumed by 2D too** | ✅ Added `TODO(#224)` markers at both sites pointing at the follow-up issue that tracks the dim-agnostic relocation. | `immune_3d.rs:14-17`, `immune_3d.rs:253-256` |
| **Naming-collision cross-link missing on `ImmuneParams`** | ✅ Added "See also `SpatialImmuneConfig`" doc block on `ImmuneParams` with a one-line "same biology, different math" summary so the relationship is visible from either side. | `params.rs:201-209` |
| **Calibration report in refactor diff** | Acknowledged as auto-regenerated side effect (STALE→PASS only on outputs whose mtime got bumped during testing — values unchanged). | (no action) |

### Verification

```
$ find simulations/output/tme -type f \( -name "*.csv" -o -name "*.json" \) | xargs shasum -a 256 > before.txt
  (... run sim-tme on refactor branch ...)
$ find simulations/output/tme -type f \( -name "*.csv" -o -name "*.json" \) | xargs shasum -a 256 > after.txt
$ diff before.txt after.txt
$ echo $?
0    ← zero changes across all 23 output files
```

Library tests still 7 passing in `params::tests::*` after doc polish.

Per the user's pattern: **not merging** — open for further review.

ELares:
## Code Review

**Reviewer:** Claude Code (on behalf of @ELares)

### Overview

Pure lift-and-cleanup PR: extracts 3 binary-local config structs (`ImmuneConfig` → `SpatialImmuneConfig`, `StromalConfig`, `PhConfig`) and the `DAMP_KILL_THRESHOLD = 0.01` magic number from `sim-tme` / `sim-tme-3d` into `ferroptosis-core`. Also folds in a NaN-correctness fix and removes dead `lp_at_death` writes. Net **−194 lines** of duplicated config code at the binary edge, **+232 lines** in the library (struct defs + 6 regression tests).

### What's strong

- **Bit-identical output verified.** All six `sim-spatial` SHA-256s and the `sim-tme-3d` `summary.json` hash match pre-refactor — this is the right load-bearing check for a refactor in a scientific simulator.
- **NaN bug fix is real and well-motivated.** The old inline `(activation * immune.immune_kill_rate * (1.0 - effective_brake)).min(0.99)` silently maps `NaN → 0.99` (near-certain kill). Routing through `immune_kill_probability`'s explicit `if raw > 0.99` branch preserves NaN so it surfaces instead of producing biologically wrong kills. The inline comment at sim-tme/src/main.rs:629 explains this clearly.
- **Naming-collision resolution is thoughtful.** Keeping `ImmuneParams` (full DC→T-cell cascade, used by sim-icd/sim-combo) alongside `SpatialImmuneConfig` (spatial DAMP-field model, used by sim-tme/sim-tme-3d) — different math, same biology — is the right call. The cross-referencing doc comments on both structs are exactly what future maintainers need.
- **`SpatialImmuneConfig` deliberately has no `Default` impl.** Forcing callers to pick `for_2d()` / `for_3d()` is a great choice: using the 2D `damp_diffusion_fraction = 0.08` in 3D would panic the `immune_3d::diffuse_damp_3d_step` stability `assert!` (0.08 × 26 > 1). Compile-time discipline beats runtime panic. Doc comment calls this out explicitly. 👍
- **Regression tests pin the literal values** so post-lift drift can't silently change the model. Including `assert!(c.damp_diffusion_fraction * 26.0 < 1.0)` as a stability-invariant guard is a nice touch — that test fails before the runtime `assert!` would.
- **Additive 0.11.0 → 0.12.0 bump** is appropriate semver for new pub items with no breaking changes.

### Issues / suggestions

**1. `DAMP_DIFFUSION_FRACTION_3D` is now a second source of truth (sim-tme-3d/src/main.rs:97).**

The const is kept "for legacy code paths that emit the value into output metadata (`RunMetadata` / `summary.json`)." But it's now hardcoded `0.025` alongside `SpatialImmuneConfig::for_3d().damp_diffusion_fraction = 0.025`. If a future PR tunes the library value, the binary const goes stale silently and `summary.json` reports the wrong number.

Suggested fix: either populate the metadata from `immune_cfg.damp_diffusion_fraction` directly (eliminate the binary const), or add a `debug_assert_eq!(DAMP_DIFFUSION_FRACTION_3D, SpatialImmuneConfig::for_3d().damp_diffusion_fraction)` at startup. The latter is one line and keeps current call sites unchanged.

**2. `DAMP_KILL_THRESHOLD` location.**

You've already flagged this with `TODO(#224)` — the const is dimensionality-agnostic but lives in the `immune_3d` module, so `sim-tme` (2D) now imports `ferroptosis_core::immune_3d::{immune_kill_probability, DAMP_KILL_THRESHOLD}` (sim-tme/src/main.rs:42), which reads oddly. Acceptable as a known follow-up; just confirming the TODO captures it.

**3. Dead-write removal — worth a `debug_assert!` while you're here.**

The `lp_at_death = gc.state.lp` writes at moment-of-death are gone (sim-tme:565, sim-tme-3d:473), and the comment says the grace-end block always overwrites `lp_at_death` before the DAMP read. That claim is the *entire* safety argument. If a future contributor adds an early-exit path that skips the grace-end block but still reads `lp_at_death`, they'll silently read whatever stale value happens to be in the field.

One-line guard: in the grace-end block, replace the unconditional `gc.lp_at_death = ...` with a `debug_assert!(gc.lp_at_death == previous_sentinel || ...)`, or initialize `lp_at_death` to `f64::NAN` so any "use before grace-end write" path produces NaN downstream and trips the calibration. Cheap insurance.

**4. Calibration-report status flips are unrelated to this PR — please confirm.**

`calibration_report.md` flips `pdt_kill_at_9mm` and `sdt_penetration_1cm` from STALE→PASS, but the observed values are unchanged (0.0105 and 0.8412). Looks like a regenerated report where the previous staleness was just "output older than source." Worth a one-line confirmation in the PR description that this is a side effect of re-running, not a behavioral change.

**5. Tests pin constructor outputs but not call-site usage.**

The 6 new tests assert `for_2d()` / `for_3d()` produce the expected literals — but they don't assert that sim-tme actually calls `for_2d()` (vs. `for_3d()`) or that sim-tme-3d calls `for_3d()`. Currently enforced only by the SHA-256 baseline. Not blocking — the bit-identical hashes are a fine end-to-end guard — but if you ever lose those baselines, a typo swapping the constructors would compile clean.

### Risk / security

- No security surface (offline scientific simulator).
- No `unsafe` introduced, no new deps, no I/O changes.
- Risk surface is entirely numerical-correctness, well-mitigated by the SHA-256 baselines + calibration PASS + 6 new regression tests.

### Verdict

**Approve with the small follow-ups above.** The lift is clean, the NaN fix is a real correctness improvement (not just refactor), the docs on the new structs are excellent, and the bit-identical verification is the right rigor for this codebase. Items 1 and 3 above are worth addressing in this PR or a fast follow-up; items 2, 4, 5 are nice-to-haves.

ELares:
## Reviewer feedback addressed (round 2)

Pushed `ad54c7a6` covering the three actionable items from the two review batches, plus the inline `params.rs:268` thread. **Bit-identical output preserved across all 24 sim-tme + sim-tme-3d outputs** (`diff before.txt after.txt` ⇒ no changes).

### Mapping

| Review item | Status | Where |
|---|---|---|
| **Inline `params.rs:268` (BLOCKING):** "2D uses 8-neighbor Moore, not 4. Doc-stated `< 1/4` bound is too loose — users could pick fractions ≥ 0.125 the docs call stable but that subtract more mass than exists." | ✅ **Docstring corrected.** Verified against `TumorGrid::neighbors` (returns up to 8) and sim-tme's diffusion loop (`damp_delta[idx] -= share * count` with `count` up to 8). All three doc sites updated: struct doc, field doc, `for_2d()` doc, and the matching inline comment in sim-tme. Also added a `damp_diffusion_fraction * 8.0 < 1.0` assertion to the 2D regression test (mirroring the 3D `* 26.0 < 1.0` guard) so future drift trips the test before it reaches a behavior-affecting value. Current `0.08` default remains comfortably safe. | `params.rs:251-256`, `params.rs:267-272`, `params.rs:283-285`, `params.rs:432-436`, `sim-tme/main.rs:352-355` |
| **Round 2 item 1: `DAMP_DIFFUSION_FRACTION_3D` is a second source of truth — `summary.json` could go stale if the library value is tuned without updating the binary const.** | ✅ Added `debug_assert_eq!(DAMP_DIFFUSION_FRACTION_3D, SpatialImmuneConfig::for_3d().damp_diffusion_fraction)` at the top of `main()`. No-op in release (preserves bit-identical), trips immediately in tests / debug runs if the two ever diverge. | `sim-tme-3d/main.rs:804-815` |
| **Round 2 item 3: `lp_at_death` use-before-grace-end safety — current zero-init lets a future early-exit path silently read 0.0 instead of failing loudly.** | ✅ Per-cell init changed from `0.0` → `f64::NAN` at all four setup sites (3 in sim-tme, 1 in sim-tme-3d). Reads only ever happen after the grace-end / end-of-sim catch-all writes in the immune-coupled paths, so bit-identical is preserved; any future code path that skips those writes will now propagate NaN into the DAMP field and trip calibration instead of producing biologically-wrong values. | `sim-tme/main.rs:178/283/492`, `sim-tme-3d/main.rs:387` |
| **Round 2 item 2: `DAMP_KILL_THRESHOLD` / `immune_kill_probability` live in `immune_3d` but are consumed by 2D too.** | Already covered by `TODO(#224)` markers added in `042d1e47`. No change. | `immune_3d.rs:14-17`, `immune_3d.rs:253-256` |
| **Round 2 item 4: confirm calibration-report STALE→PASS flips are a regeneration side effect, not a behavior change.** | ✅ Confirmed: observed values (`pdt_kill_at_9mm = 0.0105`, `sdt_penetration_1cm = 0.8412`) and residuals are unchanged from before the lift. The STALE→PASS transition is purely "output files now newer than source" after re-running the binaries during verification. Consistent with the bit-identical SHA-256 baseline this PR maintains. | (no code change) |
| **Round 2 item 5: constructor-typo coverage gap (no test asserts sim-tme calls `for_2d()` vs `for_3d()`).** | Acknowledged as a non-blocker — the SHA-256 baselines + 3D stability `assert!` catch any constructor swap in practice. Not addressed in this commit. | (no change) |

### Verification

```
$ cargo test --release -p ferroptosis-core --lib
test result: ok. 166 passed; 0 failed
$ cargo test --release -p sim-tme-3d
test result: ok. 8 passed; 0 failed
$ cargo test --release -p sim-spatial
test result: ok. 1 passed; 0 failed

$ # bit-identical check
$ diff before.txt after.txt   # 24 hashes covering output/tme/*.{csv,json} + output/tme-3d/summary.json
$ echo $?
0
```

Note: the `debug_assert_eq!` in `sim-tme-3d` does mean a debug-mode run of the binary now performs one extra equality check at startup. Release mode is unchanged. The 2D stability assertion is in a test, not the runtime.

Per the established pattern: not merging — open for further review.

PR #236, opened 2026-05-20
Workspace cargo fmt + CI fmt-check gate (#209)

### LINKED ISSUE #209: Workspace-wide cargo fmt drift (157 diffs); add fmt-check to CI

ELares:
## Summary

Closes #209. Two coordinated commits — split for separately-reviewable scope:

1. **`03acb04a` — workspace-wide `cargo fmt --all`** — resets the baseline. Touches 27 files (+1025 / −420). No semantic changes; the delta is overwhelmingly line rewrapping and import reordering.
2. **`a4a76fff` — add `fmt` job to the existing `Rust Tests` workflow** — runs `cargo fmt --all --check` on every PR + push to main that touches `simulations/**`. In parallel with the existing `test` job. 18 lines.

## What was in the drift

Issue #209 reported 157 diffs; the workspace had grown to **168 diffs** by the time this PR was started — the natural drift the gate is intended to stop. After this PR lands, that count drops to 0 and is held there by CI.

## Verification

**Bit-identical (load-bearing per #209's acceptance criteria):**

\`\`\`
$ shasum -a 256 simulations/output/spatial/*  # pre-fmt
e45241d... depth_kill_curves.csv
1365e8f... spatial_death_control.csv
3c61509... spatial_death_pdt.csv
bf20059... spatial_death_rsl3.csv
beb5de3... spatial_death_sdt.csv
ace766b... spatial_summary.json
$ diff before.txt after.txt
$ echo \$?
0   ← 6/6 hashes match after \`cargo fmt --all\`
\`\`\`

**Workspace tests:** \`cargo test --release --workspace --exclude ferroptosis-python\` → **179 passing** (166 ferroptosis-core + 1 sim-spatial + 8 sim-tme-3d + 4 ferroptosis-ffi) — same as pre-fmt.

**Fmt-check:** \`cargo fmt --manifest-path simulations/Cargo.toml --all --check\` exits 0.

## Workflow design choice

Added a second job to the existing \`cargo-test.yml\` rather than creating a separate \`cargo-fmt.yml\`. Reasoning:

- Triggers are identical (\`simulations/**\` on push to main + on PR).
- Both jobs need the Rust toolchain, but only the test job needs the build cache — so they're independent enough to parallelize cleanly.
- One workflow file is one fewer surface to drift.

The new job runs as \`Rust Tests / fmt\` in the GitHub Actions UI alongside the existing \`Rust Tests / test\`.

## Out of scope (per the issue body)

- \`cargo clippy\` warnings (the issue noted 4 pre-existing) — separate follow-up.
- Python formatter (\`ruff\`/\`black\`) — separate stack.

## Test plan

- [ ] CI: \`Rust Tests / test\` green
- [ ] CI: \`Rust Tests / fmt\` green (first run of the new job — this PR is the canary)
- [ ] CodeQL / Analyze (rust) green
- [ ] On a follow-up branch with a deliberate fmt regression, confirm the gate **fails** as intended (sanity check that the new job actually blocks drift, not just lights up green)

Closes #209

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -32,3 +32,21 @@ jobs:
       - name: Run tests and compile all binaries
         working-directory: simulations
         run: cargo test --workspace
+
+  fmt:
+    # Block PRs that introduce cargo fmt drift in the simulations
+    # workspace. Runs in parallel with `test`; rustfmt-only — no build
+    # cache needed.
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install Rust toolchain with rustfmt
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt
+
+      - name: Check formatting
+        working-directory: simulations
+        run: cargo fmt --all --check
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -42,8 +42,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
-      - name: Install Rust toolchain with rustfmt
-        uses: dtolnay/rust-toolchain@stable
+      - name: Install pinned Rust toolchain with rustfmt
+        # Pin to the workspace toolchain in `simulations/rust-toolchain.toml`
+        # rather than @stable. Cargo invocations from `simulations/`
+        # honor the pin and would otherwise auto-download the pinned
+        # channel *without* rustfmt, breaking the gate (PR #236 review).
+        # Bumping the workspace pin requires bumping this version too.
+        uses: dtolnay/rust-toolchain@1.92.0
         with:
           components: rustfmt
 
--- a/simulations/ferroptosis-core/examples/basic_usage.rs
+++ b/simulations/ferroptosis-core/examples/basic_usage.rs
@@ -28,7 +28,10 @@ fn main() {
 
     println!("ferroptosis-core v{}", env!("CARGO_PKG_VERSION"));
     println!("Parameters: default (2D culture)\n");
-    println!("{:<25} {:<10} {:<8} {:<10} {:<10} {:<10}", "Phenotype", "Treatment", "Dead?", "LP", "GSH", "GPX4");
+    println!(
+        "{:<25} {:<10} {:<8} {:<10} {:<10} {:<10}",
+        "Phenotype", "Treatment", "Dead?", "LP", "GSH", "GPX4"
+    );
     println!("{}", "-".repeat(73));
 
     for (pheno, pname) in &phenotypes {
@@ -39,7 +42,12 @@ fn main() {
             let (dead, lp, gsh, gpx4) = sim_cell(&cell, *tx, &params, &mut sim_rng);
             println!(
                 "{:<25} {:<10} {:<8} {:<10.3} {:<10.3} {:<10.3}",
-                pname, tname, if dead { "YES" } else { "no" }, lp, gsh, gpx4
+                pname,
+                tname,
+                if dead { "YES" } else { "no" },
+                lp,
+                gsh,
+                gpx4
             );
         }
     }
@@ -47,7 +55,10 @@ fn main() {
     // Demonstrate in-vivo parameters (MUFA protection) with a small population
     println!("\n--- In-vivo context (SCD1-driven MUFA protection) ---\n");
     let invivo_params = Params::invivo();
-    println!("MUFA steady-state protection: {:.2}", invivo_params.initial_mufa_protection);
+    println!(
+        "MUFA steady-state protection: {:.2}",
+        invivo_params.initial_mufa_protection
+    );
     let n = 500;
     let mut deaths_2d = 0;
     let mut deaths_vivo = 0;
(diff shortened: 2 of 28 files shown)


ELares:
## Review

### Overview
Two-commit PR closing #209: (1) a workspace-wide `cargo fmt --all` reset of `simulations/`, and (2) a new parallel `fmt` job in `Rust Tests` CI that runs `cargo fmt --all --check` on PRs/pushes touching `simulations/**`. Verification on the fmt commit is thorough — bit-identical sim-spatial outputs and unchanged test counts (179 passing) — exactly what #209's acceptance criteria asked for.

### 🔴 Blocking: the new `fmt` gate is broken (toolchain mismatch)

The first CI run is already failing — and not because of fmt drift, but because of how the new job installs the toolchain. From the job log:

```
info: syncing channel updates for 1.92.0-x86_64-unknown-linux-gnu
error: 'cargo-fmt' is not installed for the toolchain '1.92.0-x86_64-unknown-linux-gnu'.
help: run `rustup component add --toolchain 1.92.0-x86_64-unknown-linux-gnu rustfmt` to install it
```

What's happening: `dtolnay/rust-toolchain@stable` installs stable (1.95.0) **with** the `rustfmt` component. Then `working-directory: simulations` causes cargo to honor the workspace's `rust-toolchain[.toml]` pin to **1.92.0**, which gets auto-downloaded on demand — without `rustfmt`. Result: the gate fails 100% of the time, independent of any drift it's supposed to catch.

This means the PR description's claim *"the first CI run on this branch should pass"* is unfortunately not true, and — more importantly — **the gate would have shipped non-functional** if merged on green checks alone. Worth flagging because the test plan's final checkbox ("confirm the gate fails as intended on a deliberate regression") would have masked this: any failure looks like success when you're checking that a gate fails.

**Fix options** (pick one):

1. Pin the action to the workspace toolchain so rustfmt installs for the right channel:
   ```yaml
   uses: dtolnay/rust-toolchain@1.92.0
   with:
     components: rustfmt
   ```
   Coupled tightly to the pin file, but most explicit.

2. Sidestep the toolchain pin for fmt (rustfmt output is generally stable across recent channels):
   ```yaml
   run: cargo +stable fmt --all --check
   ```
   Decouples the gate from the workspace pin — preferable if you want the gate to keep working across pin bumps.

3. Add rustfmt to the pinned toolchain explicitly:
   ```yaml
   - run: rustup component add rustfmt --toolchain 1.92.0
   - run: cargo fmt --all --check
     working-directory: simulations
   ```

Option 2 is probably the lowest-maintenance choice given rustfmt output stability, but option 1 matches what the local verification did (formatter version matches workspace pin) and is the most faithful guard.

### Things that are good
- **Bit-identical verification** of sim-spatial outputs is the right acceptance bar for a "no semantic changes" claim, and 6/6 SHA-256 matches is the gold standard.
- **Splitting into two commits** (one-shot fmt reset vs CI gate addition) is reviewable and revertable independently. Good hygiene.
- **Workflow design choice** (parallel job in the existing workflow rather than a new file) — agreed with the reasoning in the PR body.
- **Out-of-scope items** (clippy, Python formatters) are correctly deferred.

### Minor
- The comment in the new job (`# Block PRs that introduce cargo fmt drift...`) is helpful and worth keeping.
- The fmt commit is enormous (+1025 / −420 across 27 files) but that's intrinsic to the task — nothing to do about it.
- Consider whether the local verification should be re-run with the same toolchain CI uses (would have caught the bug). For future drift-reset PRs of this kind, `cargo +<pinned-version> fmt --all --check` locally would mirror CI more faithfully than the unqualified invocation in the PR body.

### Suggested next step
Push the toolchain fix, watch the `fmt` job go green on this PR (validating it loads), then validate the **negative case** on a throwaway branch with a deliberate format regression before merging — same test plan item, but now meaningful.

Otherwise: a well-scoped, well-verified PR. Just needs the toolchain wiring fix before merge.

ELares:
## Toolchain bug fixed (commit \`21235a16\`)

Thanks for catching this — you're right that the test plan's "confirm the gate fails on a deliberate regression" check would have green-stamped a non-functional gate. Worth flagging in retrospect that **my local verification (`cargo fmt --check`) used the workspace-pinned rustfmt 1.8.0** (from rustc 1.92.0), but the CI gate as written would have tried to use stable. The local-vs-CI mismatch is exactly the failure mode the comment in the issue body was guarding against.

### What I changed

Took **option 1** (pin the action to `@1.92.0` to match `simulations/rust-toolchain.toml`). Rationale: the one-shot `cargo fmt --all` in commit `03acb04a` was applied with rustfmt 1.8.0 / rustc 1.92.0 — the same formatter the workspace pin honors. Pinning the gate to the same version means it validates the *exact* output the codebase was written against; any rustfmt-output drift between stable and 1.92.0 would otherwise produce spurious gate failures (or worse, spurious passes).

Trade-off: the action version is now coupled to the workspace pin. Comment on the step calls this out so a future toolchain bump touches both files. Considered option 2 (`cargo +stable fmt --check`) for lower maintenance, but the "most faithful guard" framing matches the SHA-256-bit-identical philosophy used elsewhere in this repo.

### Verification

Watched the new commit's CI in full:

\`\`\`
fmt                                      SUCCESS  ← was failing before fix
test                                     SUCCESS
Analyze (actions)                        SUCCESS
Analyze (c-cpp)                          SUCCESS
Analyze (javascript-typescript)          SUCCESS
Analyze (python)                         SUCCESS
Analyze (rust)                           SUCCESS
CodeQL                                   SUCCESS
\`\`\`

8/8 green, including the new `fmt` gate on its second-ever run.

### About the negative-case test

I haven't done the deliberate-format-regression canary you suggested. The gate ran clean on this PR, which validates it loads and recognizes clean code as clean — but doesn't validate that it actually fails on dirty code. Three ways to handle that pre-merge if you want it:

1. Push a throwaway commit on this branch that breaks fmt, verify `fmt` job fails, then revert the throwaway. Cleanest.
2. Open a separate one-line "deliberately break fmt" PR against this branch (or main after this lands), confirm fail, close without merging.
3. Treat the positive case as sufficient and trust `cargo fmt --check`'s exit-1-on-diff behavior (well-established, used by thousands of repos).

I'd suggest (3) for simplicity, but happy to do (1) or (2) if you want the rigor — it's a 2-minute exercise either way.

### Process lesson

Updated my own mental model: for any future drift-reset PR of this kind, local verification should mirror CI's toolchain by using `cargo +<pinned-version>` explicitly, not the unqualified invocation that silently uses whatever's nearest. Otherwise the local pass and CI pass are checking different things.

Awaiting your review before merge.

PR #237, opened 2026-05-20
Close #220 follow-ups: lift 2D math, JSON schema_version, immune_3d → immune_spatial (#224)

### LINKED ISSUE #224: Follow-ups from #220: retire inline 2D test math, JSON schema versioning, immune_3d module rename

ELares:
## Summary

Closes #224 — all three deferred items from #220, landed as 7 small commits for separately-reviewable scope.

## Per-item mapping

| #224 item | Commit | Net |
|---|---|---|
| **3** — rename `immune_3d` → `immune_spatial` | \`220c3c9e\` | +42 / −35 (mostly doc) |
| **1a** — lift 2D depth-from-edge to library (`TumorGrid::radial_depth_um`) | \`95621a0c\` | +105 / −0 (4 new tests) |
| **1b** — lift 2D pH formula + scalar `ph_at_depth` | \`dfc010f4\` | +129 / −16 (4 new tests) |
| **1c** — lift 2D `stromal_adjacency_mask_2d` + `stromal_adjacent_kill_rate_2d`, rename 3D with `_3d` suffix | \`b7b82274\` | +159 / −29 (5 new tests) |
| **1d** — refactor sim-tme to call lifted helpers (delete binary-local copies) | \`173f3e6f\` | +43 / −110 |
| **1e** — refactor library cross-geometry tests to call lifted helpers | \`39227067\` | +28 / −56 |
| **2** — JSON `schema_version: 1` envelope on both binaries + Python assert + doc | \`0e3d4920\` | +94 / −8 |

**Total net**: +600 / −254 across library, binaries, scripts, tests, and one README.

## Verification matrix

For each commit:

| Commit | `cargo test` | `cargo fmt` | sim-tme + sim-tme-3d + sim-spatial outputs |
|---|---|---|---|
| 1 (rename) | 179 ✅ | ✅ | **30/30 bit-identical** to baseline |
| 2 (depth lift) | 170 → 174 ✅ | ✅ | 30/30 bit-identical |
| 3 (pH lift) | 174 → 178 ✅ | ✅ | 30/30 bit-identical |
| 4 (stromal lift) | 178 → 179 (3 of 4 new tests overlap-named) ✅ | ✅ | 30/30 bit-identical |
| 5 (sim-tme refactor) | 179 ✅ | ✅ | **30/30 bit-identical** ← load-bearing refactor check |
| 6 (lib test refactor) | 179 ✅ | ✅ | 30/30 bit-identical |
| 7 (schema_version) | 179 ✅ | ✅ | 28/30 bit-identical; **2 JSONs intentionally differ** by exactly `schema_version: 1` |

The "30/30 bit-identical through commit 6" is the load-bearing claim: the four-stage refactor (library helpers added → 3D renamed → sim-tme refactored → tests refactored) was behavior-preserving across both 2D and 3D binaries. The two JSONs only change in commit 7 (and only by the new top-level field — verified by inspecting the keys).

## Item 3 — `immune_3d` → `immune_spatial`

The module mixed 3D-specific grid helpers (`diffuse_damp_3d_step`) with dimensionality-agnostic scalar helpers (`dc_activation`, `immune_kill_probability`, `DAMP_KILL_THRESHOLD`). Since #220, both sim-tme (2D) and sim-tme-3d (3D) import the scalars — making the `_3d` module name misleading. Renamed to `immune_spatial`; the grid helpers keep their `_3d` function-name suffix (they're genuinely 3D-only). \`git mv\` preserved at 95% similarity. The two `TODO(#224)` markers added in #225 are gone.

## Item 1 — Retire inline 2D test math

The 3 cross-geometry tests (`oxygen::tests::matched_lambda_2d_vs_3d_zone_fractions`, `ph::tests::matched_lambda_2d_vs_3d_acidic_fraction`, `stromal::tests::matched_dims_3d_boundary_fraction_exceeds_2d`) previously inlined sim-tme's depth-from-edge, pH formula, and 8-Moore boundary detection. Each was lifted to a library function with the same `_2d` / `_3d` naming pattern used by the 3D versions:

- `TumorGrid::radial_depth_um(r, c) -> f64` (mirrors `TumorGrid3D::radial_depth_um(r, c, l)`)
- `ph::ph_at_depth(...)` (scalar primitive used by both 2D and 3D field functions) + `ph::radial_ph_field_2d(...)`
- `stromal::stromal_adjacency_mask_2d(...)` + `stromal::stromal_adjacent_kill_rate_2d(...)`

Sim-tme's binary-local copies (`compute_depth_map`, `apply_o2_gradient`, `apply_ph_gradient`, `stromal_adjacency_mask`, `stromal_adjacent_kill_rate`) are either deleted (stromal pair) or rewritten as thin orchestrators that call the lifted helpers (depth, O2, pH). The previously-stale "Source of truth — keep in sync with sim-tme/src/main.rs:XXX" comments are gone; the test arms and the binary now share the library helper as a single source of truth.

The line-stale reference at `oxygen.rs:481` (called out in #224 — pointed at `sim-tme/src/main.rs:836` which had drifted to `zone_kill_rates`) is removed with the surrounding inline-math block.

## Item 2 — JSON `schema_version`

Both `output/tme/tme_summary.json` (sim-tme) and `output/tme-3d/summary.json` (sim-tme-3d) now emit \`{ "schema_version": 1, ... }\` at the top level. For sim-tme this required wrapping its previous bare-array output in an envelope object; the comparison and figure scripts were updated to unwrap. The comparison script (\`scripts/generate_3d_comparison_table.py\`) asserts both files have the same \`schema_version\` equal to its \`EXPECTED_SCHEMA_VERSION\` constant — schema drift between the two binaries surfaces immediately.

Schema-versioning convention documented in \`simulations/sim-tme-3d/README.md\` under "Output → Schema versioning": explains the envelope, when to bump, and the cross-binary equality assertion.

## Out of scope

- The 3D mask/kill-rate function-name rename means **any external consumer** of \`ferroptosis-core::stromal::stromal_adjacency_mask\` outside this workspace would need an import-path update. No such consumer exists in this repo (ferroptosis-ffi and ferroptosis-python only consume the non-stromal API).
- \`cycling_lambda\` (sim-tme:130) is referenced in \`oxygen.rs:377\`'s test — not in #224's scope list. Left as-is.

## Test plan

- [ ] CI: \`Rust Tests / test\` green
- [ ] CI: \`Rust Tests / fmt\` green (the gate from #209)
- [ ] CI: \`Analyze (rust)\` / \`CodeQL\` green
- [ ] Local: \`pytest tests/test_calibrate.py -v\` → 14/14 passing
- [ ] Local: \`python3 scripts/generate_3d_comparison_table.py\` succeeds, prints Q1-Q4

Closes #224


ELares:
# Code Review — PR #237

## Overview

Closes 3 deferred follow-ups from #224 as 7 separately-reviewable commits:

1. **Item 3** — Rename `immune_3d` module → `immune_spatial`. The module mixes 3D-grid helpers (`diffuse_damp_3d_step`) with dimensionality-agnostic scalar helpers (`dc_activation`, `immune_kill_probability`, `DAMP_KILL_THRESHOLD`) consumed by both sim-tme (2D) and sim-tme-3d (3D); the `_3d` module name had become misleading.
2. **Item 1** — Lift sim-tme's binary-local 2D math (depth-from-edge, pH formula, stromal 8-Moore adjacency + kill-rate) into `ferroptosis-core`. Adds `TumorGrid::radial_depth_um`, `ph::ph_at_depth` + `radial_ph_field_2d`, and `stromal::stromal_adjacency_mask_2d` + `stromal_adjacent_kill_rate_2d`; renames 3D counterparts to add `_3d` suffix.
3. **Item 2** — Add `schema_version: 1` envelope to `tme_summary.json` (sim-tme) and `summary.json` (sim-tme-3d). Python comparison script cross-checks both files against `EXPECTED_SCHEMA_VERSION`.

Net **+600 / −254** across 16 files. PR description's verification matrix claims **30/30 bit-identical outputs through commit 6**, with the 2 JSONs intentionally diverging only in commit 7 (the schema_version addition).

## Strengths

- **Drift-proof cross-geometry tests.** The previously-stale "keep in sync with sim-tme/src/main.rs:XXX" comments in `oxygen.rs:481`, `ph.rs:tests`, and `stromal.rs:tests` are gone — both arms now call the same library helper. The `*_matches_sim_tme_*_formula` tests (e.g. `radial_depth_2d_matches_sim_tme_compute_depth_map_formula`, `radial_ph_field_2d_matches_sim_tme_apply_ph_gradient_formula`, `mask_2d_matches_sim_tme_stromal_adjacency_mask_formula`) lock the lift bit-identically and would trip on any future divergence on either side.
- **Naming convention is consistent across the API.** `TumorGrid::radial_depth_um(r, c)` mirrors `TumorGrid3D::radial_depth_um(r, c, l)`; `stromal_adjacency_mask_2d` parallels `stromal_adjacency_mask_3d`; `radial_ph_field_2d` parallels `radial_ph_field`. Both arities share `TUMOR_RADIUS_FRACTION` so geometry and depth-from-surface reference can't drift.
- **`ph_at_depth` as a shared scalar primitive** called by both `radial_ph_field` (3D) and `radial_ph_field_2d` (2D) — one place to tune the formula. Same pattern for `kill_rate_over_mask` as a private shared body between `stromal_adjacent_kill_rate_2d` / `_3d` (clean workaround for the missing `TumorGrid` ↔ `TumorGrid3D` trait).
- **`debug_assert_ph_inputs` helper** with `ctx: &'static str` extracts and DRYs the finite-check/`ph_edge > ph_core`/positive-`lambda_ph_um` validation across both field functions.
- **`TmeSummary<'a>`** uses a borrowed slice for the serialization-only envelope view — avoids moving `all_results` out from under the downstream comparison-table block. Idiomatic Rust.
- **The pH iron-modulation refactor in `apply_ph_gradient`** — clever observation that `iron_multiplier_from_ph(ph_edge, ph_edge, _) == 1.0` (no-op on non-tumor cells), then preserving the explicit `is_tumor` gate anyway for semantic clarity. Comment justifies it well; numeric result is unchanged.
- **Schema-versioning convention is documented** in `simulations/sim-tme-3d/README.md` under "Output → Schema versioning": explains the envelope shape, when to bump (renames / semantic changes; not optional-field additions), and the cross-binary equality assertion.
- **`_check_schema_version` error messages are actionable** — they name the source file/binary and suggest "Was the binary rebuilt against #224?" or "Bump the script or regenerate the output." Good operator UX.

## Issues / Suggestions

### Minor

1. **Asymmetric schema-version enforcement between Python scripts.** `generate_3d_comparison_table.py` strictly asserts `schema_version == EXPECTED_SCHEMA_VERSION`, but `generate_figures.py:1140` quietly tolerates either the new envelope OR a bare array via `summary.get("conditions", summary) if isinstance(summary, dict) else summary`. This means a corrupt or stale `tme_summary.json` will silently produce wrong figures, while the comparison script fails loudly. Consider either:
   - Removing the bare-array fallback once you're confident no pre-#224 outputs are in circulation, OR
   - Extracting `_check_schema_version` into a shared helper module and calling it from `generate_figures.py` too — even if it just warns rather than aborts.

2. **`test_calibrate.py:summary_dict` fixture** acknowledges in the docstring that "The extractor doesn't read this field today, but keeping the fixture faithful to the on-disk shape catches schema drift if a future change starts asserting on it." That's honest, but it also suggests the extractor itself should assert `schema_version` — the same way `generate_3d_comparison_table.py` does — to surface drift uniformly across all Python consumers. Worth a follow-up issue.

3. **Public Rust API rename — `stromal_adjacency_mask` → `stromal_adjacency_mask_3d`**, similarly for `stromal_adjacent_kill_rate`. The PR notes (correctly) that `ferroptosis-ffi` and `ferroptosis-python` don't import these and no in-repo consumer breaks. That said, any external `cargo` dependent of `ferroptosis-core` outside this workspace would silently break with a compile error. If the crate is or will be published, consider re-exporting the old names as `#[deprecated]` aliases for one release cycle. If it's strictly internal, ignore.

4. **`schema_version: u32`** is fine for the all-or-nothing bump convention the README documents, but offers no signal for backwards-compatible vs. breaking changes. A second `min_compatible_version` field (or a semver string) would let the Python side accept newer-minor outputs without a script bump. Not blocking — likely overkill for a 2-binary system — but worth a thought if the schema grows downstream.

### Nits

- `simulations/sim-tme/src/main.rs:74` — `apply_o2_gradient`'s rustdoc says "Depth math is sourced from `TumorGrid::radial_depth_um` (#224 item 1a)." This is a good cross-reference. Consider applying the same one-liner to `apply_ph_gradient` (line ~366) which also delegates to the lifted function — a passing reader currently has to infer it from the body.
- `ferroptosis-core/src/grid.rs:170` — the `radial_depth_um` rustdoc notes "Not bounds-checked: matching `get`'s contract..." which is good. The 3D version (`TumorGrid3D::radial_depth_um`) has identical semantics — verify the 3D rustdoc says the same so the parallel is explicit.

## Risk / Correctness Notes

- **The "30/30 bit-identical through commit 6" verification is load-bearing.** The four-stage refactor (lift helpers → rename 3D → refactor sim-tme call sites → refactor test arms) hinges on the binary producing the same JSON bytes at each step. If CI lacks a regression-snapshot test for the binaries' JSON outputs, the verification only catches drift at the commit-author level — future refactors won't get the same protection. A test that diffs `tme_summary.json` against a checked-in baseline (allowing `schema_version` to differ) would make this guarantee permanent.
- **`tme_summary.json` shape change is breaking for any external reader.** It was a JSON array; it is now `{ "schema_version": 1, "conditions": [...] }`. The in-repo scripts (`generate_3d_comparison_table.py`, `generate_figures.py`) are updated. Confirm there is no external dashboard, notebook, or pipeline that reads this file directly.
- **`#[must_use]` annotations** on `stromal_adjacency_mask_2d` and `stromal_adjacent_kill_rate_2d` were copied across from the 3D versions — correct, since the same side-effect-free-output contract holds.
- **The two `TODO(#224)` markers** in `immune_3d.rs` (around `DAMP_KILL_THRESHOLD` and `immune_kill_probability`) are removed as part of the rename, as the description states. Good — clearing TODOs as the work that resolves them lands is the right cleanup discipline.

## Tests

- Library: +13 tests across `grid.rs` (4), `ph.rs` (4), `stromal.rs` (5). Each new 2D test mirrors an existing 3D test by name and intent, then adds a `_matches_sim_tme_*_formula` lock test. Coverage of the lift is thorough.
- Binary: no new sim-tme tests; the bit-identical-output claim is the regression guard at the integration level. Worth eventually upgrading to an in-CI snapshot (see Risk note above).
- Python: `tests/test_calibrate.py`'s fixture updated to the envelope shape with a forward-looking comment.

## Performance

No runtime-perf-relevant changes. The lifted functions are direct extractions of the previous inline math (depth-from-edge is a single sqrt + multiply; pH formula is one exp + clamp; mask construction is the same nested 8-neighbor loop). `kill_rate_over_mask` private helper inlines well under release optimization.

## Security

No new I/O surfaces. The JSON schema check is a defensive read-side validation; no parsing of untrusted input. Safe.

## Verdict

**Approve.** Clean, well-scoped, drift-proof refactor with strong test coverage of the lift. The structural improvements (shared `ph_at_depth` primitive, parallel naming, lifted depth function) reduce future maintenance load. The two issues worth follow-up — asymmetric schema enforcement in `generate_figures.py` and absence of a CI-level JSON regression snapshot — are post-merge cleanup items, not blockers.

ELares:
## Review addressed + self-review findings (commit \`7020adf6\`)

Pushed a follow-up commit that closes the human review feedback **and** two real bugs an independent self-review found that I had missed.

### Mapping the human review

| Item | Status | Where |
|---|---|---|
| **Asymmetric schema enforcement (`generate_figures.py` soft-fallback vs `generate_3d_comparison_table.py` hard-assert)** | Acknowledged as a follow-up. Filing as #224-style residual rather than expanding scope here. Trade-off is deliberate: figure generation should degrade gracefully for old outputs; the comparison script is the integrity check. | — |
| **`test_calibrate.py` extractor should assert `schema_version`** | Acknowledged as a follow-up. Currently the fixture matches the on-disk shape; adding a real assert to `calibrate.py::extract_tme_3d_json` is the right place but is separate scope (touches the extractor, not the lift). | — |
| **`#[deprecated]` aliases for renamed `stromal_*` Rust functions** | Skipping — the reviewer noted "If it's strictly internal, ignore." `ferroptosis-core` is internal (path dependency only); no published crate consumers. | — |
| **`min_compatible_version` field** | Skipping — reviewer flagged as "likely overkill for a 2-binary system." Agreed. | — |
| **Nit: `apply_ph_gradient` rustdoc cross-reference to lifted helper** | ✅ Added "#224 item 1b" pointer plus a fuller note on why the `is_tumor` gate is retained (pathological `iron_ph_sensitivity = +∞` would make `1.0 + ∞·0 = NaN` at non-tumor cells). | `sim-tme/main.rs:apply_ph_gradient` rustdoc |
| **Nit: 3D `radial_depth_um` "Not bounds-checked" symmetry** | ✅ Added the same perf-note + bounds-check parity comment to the 2D `radial_depth_um`. | `grid.rs:TumorGrid::radial_depth_um` rustdoc |
| **Risk: no CI snapshot for binary JSON outputs** | Acknowledged. Hash-based regression tests for `tme_summary.json` / `summary.json` are a worthwhile follow-up. | — |
| **Risk: `tme_summary.json` shape change breaks external readers** | Confirmed — only in-repo consumers (`generate_3d_comparison_table.py`, `generate_figures.py`) read this file; both updated. | — |

### Self-review caught two real bugs I missed

While reviewing my own diffs with skepticism, I found:

**1. The fmt CI was failing — and my local check was lying.** My local verification script wrote
\`\`\`bash
cargo fmt --all --check 2>&1 | tail -3 && echo "fmt: $?"
\`\`\`
The \`$?\` was capturing **echo's** exit code, not cargo fmt's. The actual fmt-check had been red since commit \`173f3e6f\` (new \`ph\` and \`stromal\` imports went between \`io\` and \`params\` instead of in alphabetical order). `cargo fmt --all` fixed it; the actual import block now reads alphabetically. **Memory rule update**: when verifying fmt locally, exit code must come from the formatter directly — never piped or chained through anything that has its own exit.

**2. `zone_kill_rates` was the last call site in sim-tme still inlining the depth-from-edge math** with a hardcoded \`0.45\` literal — exactly the drift hazard #224 set out to eliminate. The other three call sites (\`apply_o2_gradient\`, \`compute_depth_map\`, \`apply_ph_gradient\`) were lifted in commit \`173f3e6f\`, but I missed this one. Fixed to use \`grid.radial_depth_um(r, c).max(0.0)\`. Bit-identical verified.

**3. Bonus: `ferroptosis-core/README.md` had a code example that wouldn't compile.** The blanket sed of \`immune_3d\` → \`immune_spatial\` correctly updated `immune` references, but the stromal renames were done with \`s/stromal_adjacency_mask(/stromal_adjacency_mask_3d(/g\` — which only catches the form-with-open-paren. The README's example uses backtick-quoted names and an import line without parens — both fell through. The example would have failed \`cargo test --doc\`. Fixed; module table refreshed to list both 2D and 3D function names rather than only 3D.

\`sim-tme-3d/README.md:12\` had the same stale name in prose; fixed.

### Verification on the new commit

\`\`\`
fmt:                                     SUCCESS  ← was FAILURE on the prior commit
test (Rust):                             SUCCESS
test-linux (Python):                     SUCCESS
test-macos:                              SKIPPED
Analyze (rust / actions / ...):          SUCCESS
CodeQL:                                  SUCCESS
\`\`\`

\`\`\`
$ cargo test --release --workspace --exclude ferroptosis-python
test result: ok. 179 passed; 0 failed
$ shasum -a 256 output/tme/* output/tme-3d/* output/spatial/* | sort
   ← 30/30 hashes match the post-schema baseline (zone_kill_rates lift is behavior-preserving)
\`\`\`

### Architectural / design notes I considered

A few things I looked at and concluded **are correct as-is**, in case anyone wonders:

- **\`TmeSummary<'a>\` borrowed-slice form** — serde \`Serialize\` for \`&[T]\` writes byte-identical to \`Vec<T>\` (both go through \`serialize_seq\` with the same length + elements). The lifetime is mild noise but avoids cloning \`all_results\` for the comparison-table block downstream.
- **The \`*_matches_sim_tme_*_formula\` tests** are now library-vs-library (sim-tme no longer has the inline math). The reviewer correctly notes these become "circular" checks. They still have value as drift guards: if a future refactor changes the lifted formula AND the inline reference identically, both would silently change — but a refactor of just one side will trip them, which is the realistic drift mode. Worth keeping; the docstrings should be tightened to reflect this is now the rationale.
- **\`kill_rate_over_mask\` private helper** uses \`&[GridCell]\` for dim-agnosticism. \`GridCell\` is \`pub\` (so it could in principle be a public helper too), but keeping the function private and exposing only the typed \`_2d\` / \`_3d\` wrappers is the better API surface.
- **No shared \`SCHEMA_VERSION\` between sim-tme and sim-tme-3d** — both have their own const. A library-level const would link them at compile time, but adds a coupling the schema convention doesn't strictly require (the script-side cross-check is what makes the contract load-bearing). Worth a follow-up if a 3rd binary joins.

PR #251, opened 2026-05-31
Multi-dose pharmacokinetics: time-varying drug schedules (#239)

### LINKED ISSUE #239: Multi-dose pharmacokinetics: time-varying drug concentration profiles
## Context

Today RSL3 / PDT / SDT are modeled as **constant drug concentration** present uniformly throughout the 180-step simulation (other than what the photosensitizer-PK module already supports for PDT). Reality has absorption → distribution → metabolism → excretion (ADME) curves, repeated clinical dosing, and drug accumulation/clearance in time as well as space. This is one of six realism gaps identified during the #193 visualization PR (alongside #191, #197, and the new persister/clonal/immune-realism issues being filed now).

## What's already there

- `ferroptosis-core::photosensitizer_pk` (#200, #208) — single-exponential PK + saturating distribution phase + relative singlet-O₂ yield for the PDT pathway only.
- `ferroptosis-core::tumor_pk` — two-compartment PK model with depth-dependent concentration. **Not wired into any of the spatial binaries** (sim-spatial, sim-tme, sim-tme-3d).
- `Params { rsl3_gpx4_inhib, pdt_ros, sdt_ros }` — constant scalars set at simulation start. No time dependence.

- **Multi-dose schedule.** Drug peaks at the start of each dose, decays per PK model, accumulates if dose interval < clearance time. Standard 5-dose / 3-dose protocols selectable.
- **Time-varying drug at each cell.** `tumor_pk` becomes the source of truth for `(step, depth) → drug_concentration`; sim-tme-3d reads from it per step.
- **Drug-specific PK parameters** for at least RSL3 (currently PK is photosensitizer-only).
- **Backwards-compatible.** Steady-state / single-dose remains an option (used for cross-condition apples-to-apples comparisons in #195).

## Suggested approach

1. Extend `tumor_pk` to expose `concentration_at(step, depth)` — needs serial-dose schedule input.
2. Add `DoseSchedule` to RunConfig / Condition: `Option<DoseSchedule>` (None = steady-state, preserves current behavior).
3. Refactor sim-tme-3d's biochem loop to multiply per-cell biochem rates by the time-varying drug concentration.
4. Bit-identical verification: when DoseSchedule is `None`, output matches current SHA-256.

## Acceptance criteria

- [ ] At least one drug (RSL3 first) has a full multi-dose PK profile in the simulation (not just photosensitizer)
- [ ] sim-tme-3d supports a multi-dose protocol (e.g., 5 doses every 36 steps) via Condition / RunConfig parameter
- [ ] Backwards-compatibility: steady-state remains the default
- [ ] Bit-identical output for the steady-state case (gates against silent behavior change)
- [ ] Documentation: dose-schedule semantics + how to add a new drug's PK

## Out of scope (file separately if needed)

- Drug-drug interactions (combination PK)
- Inter-patient PK variability (population PK)
- Active transport across cell membranes (Henderson-Hasselbalch in-trap already in `ph`)

## Related

- #200 — photosensitizer PK (existing single-drug PK template to mirror)
- #208 — PK enhancement (saturating distribution)
- #194 — sim-spatial-3d (consumer of this work)
- #191 — vasculature (orthogonal — vasculature heterogeneity changes drug delivery topology, this issue is about temporal dosing)

## Source

Identified during PR #238 discussion of realism gaps in the sim-tme-3d snapshot. 1 of 5 issues filed from that audit.

ELares:
## Summary

Implements issue #239 (multi-dose / time-varying pharmacokinetics). The simulator previously modeled drug as present at **constant** strength for the whole run; this adds a `DoseSchedule` abstraction (bolus / multi-dose / infusion / PK-curve) and wires it into sim-tme-3d, behind a gate that keeps the default path **byte-identical**.

> **Update:** #238 has merged to main, and this branch has been **rebased onto current main** (the duplicated viz commits dropped; one renderer conflict resolved to keep main's refactor + the #239 dose markers). Base is now `main`; the diff is the #239 work only. Re-verified under CI's pinned toolchain (rust 1.96.0): fmt clean, 189 ferroptosis-core + 23 sim-tme-3d tests pass, default `summary.json` SHA `1cf8866312…` unchanged, dose-sweep numbers unchanged.

## The keystone finding

The understand-phase mapping revealed that `ferroptosis-core::tumor_pk` **already** had validated per-step time-varying drug logic (`sim_cell_with_pk`) — it was just orphaned to the `sim-tumor-pk` scalar binary and never wired into the spatial grid. So #239 was fundamentally a **wiring/abstraction** problem, not new physics.

## Four commits

| # | What |
|---|---|
| 1 | `DoseSchedule` library module (Constant/Bolus/MultiDose/Infusion/FromPk) + `from_cell_with_ros_opts` (skips RSL3 one-shot knockdown for the dosed path) |
| 2 | Wire into sim-tme-3d step loop behind `is_constant()` gate |
| 3 | `--dose-sweep` mode → `dose_comparison.json` + `tumor_pk` **FromPk bridge** |
| 4 | `--snapshot=multidose` preset + **SDT envelope-neutralization** + renderer dose markers |
| 5–7 | Review follow-ups: architecture-doc contracts, Constant-path golden regression test, clamp/alloc/assert cleanup + Infusion/FromPk end-to-end tests |

## How drug becomes time-varying

`DoseSchedule::factor_at(step) -> f64` is a per-step availability multiplier (Constant ⇒ exactly 1.0). Applied per modality:
- **RSL3**: drives per-step covalent GPX4 inactivation (`gpx4 -= RSL3_INACTIVATION_RATE · availability · gpx4`, the validated `sim_cell_with_pk` model) instead of the one-shot init knockdown. pH ion-trapping composes as a per-cell spatial availability multiplier.
- **SDT/PDT**: scales the per-step exogenous-ROS bolus. **Envelope-neutralization fix**: `sim_cell_step`'s intrinsic single-bolus decay (extracted to `biochem::exo_decay_factor`) is divided out on the dosed path, so the schedule is the *sole* availability envelope — otherwise ROS from a late dose would be wrongly crushed by the from-t=0 decay.

## Two headline results

**`--dose-sweep`** (RSL3 across protocols, shared grid + RNG seed):

| protocol | kill % | ferro | immune |
|---|---|---|---|
| constant | 0.39 | 314 | 10 |
| multidose | 0.03 | 27 | 0 |
| infusion | 0.02 | 20 | 0 |
| frompk | 0.02 | 16 | 0 |
| bolus | 0.01 | 5 | 0 |

The steady-state `constant` model **overestimates RSL3 kill by 10-60×** vs every realistic time-varying protocol — a direct quantification of the "drug present forever" idealization. (Magnitudes are uncalibrated v1; the cross-protocol ordering is the signal.)

**`--snapshot=multidose`** (SDT, 4 doses at steps 10/55/100/145): after the envelope fix, all four doses produce clean death waves (7928 / 27823 / 13040 / 7230 deaths, 69.6% total kill). The renderer marks each dose frame with a red border + `💉 DOSE`.

## Bit-identical guarantee

When every condition uses `DoseSchedule::Constant` (the entire default 24-condition matrix), the run is **byte-identical** to pre-#239: `summary.json` SHA = `1cf8866312…` (verified at every commit). The dose machinery is gated behind `is_constant()`; `--dose-sweep` writes a separate file and never touches `summary.json`. The `exo_decay_factor` extraction is an arithmetically-identical refactor of 3 call sites.

## Verification

- **189** ferroptosis-core (+10 DoseSchedule, +1 exo_decay_factor) + **20** sim-tme-3d (+4 dosed-path) tests pass.
- `cargo fmt --all --check` exits 0.
- `summary.json` byte-identical to baseline at every commit.
- `--dose-sweep` and `--snapshot=multidose` both run end-to-end; multidose GIF verified (180 frames, 4 death waves confirmed by per-step death-count analysis).

## Self-review

Ran an adversarial review workflow (4 dimension reviewers → verification of every medium+ finding). Both medium findings were **refuted** on inspection (the Bolus `half_life` guard is unreachable with current inputs + idiomatic; the PK normalization fails-loud rather than masking). No critical/high findings; the bit-identical and RSL3-dosed dimensions found only confirmations.

## Known limitation (follow-up)

No CI-level golden-SHA regression test for the byte-identity claim (same gap #237 noted). The full 60³×180 production run is too slow for CI; determinism + kill-accounting are guarded by existing fast tests. A checked-in baseline snapshot test is worth a follow-up.

## Test plan

- [ ] CI: `Rust Tests / test` + `fmt` green
- [ ] CI: CodeQL / Analyze green
- [x] #238 merged; this rebased onto `main` (base is now `main`, conflicts resolved)

Closes #239

### DIFF
--- a/scripts/render_tme_3d_trajectory.py
+++ b/scripts/render_tme_3d_trajectory.py
@@ -120,14 +120,28 @@ def _render(
     damp_max = max(float(damp.max()), 1e-6)
     lp_max = max(float(lp.max()), 1e-6)
 
+    # Dose-administration steps (#239). Empty for steady-state Constant
+    # presets; non-empty for multi-dose / bolus / infusion snapshots, where
+    # we annotate the frames so the viewer can see death waves sync to doses.
+    dose_steps = set(int(s) for s in meta.get("dose_steps", []))
+    # Highlight the dose frame plus a few following frames (the death wave
+    # lags the dose), so the marker is visible for a beat rather than 1 frame.
+    dose_window = set()
+    for d in dose_steps:
+        for k in range(d, d + 5):
+            dose_window.add(k)
+
     fig, axes = plt.subplots(1, 3, figsize=(13, 4.5), constrained_layout=True)
     cond = meta.get("condition", {})
+    dose_caption = (
+        f"   doses@{sorted(dose_steps)}" if dose_steps else "   (constant dosing)"
+    )
     fig.suptitle(
         f"sim-tme-3d  axial mid-slice  ({cond.get('treatment', '?')}, "
         f"immune={cond.get('immune_mode', '?')}, "
         f"stromal={cond.get('stromal_mode') or 'off'}, "
         f"ph={cond.get('ph_mode') or 'off'}, "
-        f"λ_O₂={cond.get('o2_lambda_um', '?')}µm)",
+        f"λ_O₂={cond.get('o2_lambda_um', '?')}µm){dose_caption}",
         fontsize=10,
     )
 
@@ -158,7 +172,22 @@ def _render(
         im_lp.set_data(lp[step, mid])
         # Count cumulative dead cells in this slice for a quantitative cue.
         n_dead_slice = int(dead[step, mid].sum())
-        step_text.set_text(f"step {step + 1:3d}/{n_steps}    dead-in-slice={n_dead_slice}")
+        # Mark dose frames so multi-dose death waves are visually attributable.
+        if step in dose_steps:
+            dose_marker = "  💉 DOSE"
+        elif step in dose_window:
+            dose_marker = "  💉 ···"
+        else:
+            dose_marker = ""
+        step_text.set_text(
+            f"step {step + 1:3d}/{n_steps}    dead-in-slice={n_dead_slice}{dose_marker}"
+        )
+        # Red frame border on the dose step itself — a hard-to-miss cue.
+        border_on = step in dose_steps
+        for ax in axes:
+            for spine in ax.spines.values():
+                spine.set_edgecolor("red" if border_on else "none")
+                spine.set_linewidth(3.0 if border_on else 0.0)
         return [im_dead, im_damp, im_lp, step_text]
 
     interval_ms = max(1, int(1000.0 / fps))
--- a/simulations/ferroptosis-core/src/biochem.rs
+++ b/simulations/ferroptosis-core/src/biochem.rs
@@ -117,6 +117,27 @@ fn update_mufa_protection(current: f64, params: &Params) -> f64 {
     (current + growth - decay).clamp(0.0, params.scd_mufa_max.max(0.0))
 }
 
+/// Deterministic exogenous-ROS decay envelope for the post-bolus phase.
+///
+/// Models singlet-oxygen / exogenous-ROS decay after a single treatment
+/// bolus: `1.0` for `step < 30` (the pre-decay plateau, where
+/// [`sim_cell_step`] instead applies multiplicative noise), then
+/// `0.5^((step-30)/15)` — a 15-step half-life decay.
+///
+/// Exposed so a time-varying [`crate::dose_schedule::DoseSchedule`]
+/// consumer can **divide it out**: for multi-dose SDT/PDT the schedule's
+/// own per-dose rise+decay is the availability envelope, and this
+/// single-bolus envelope (keyed to run start) would otherwise
+/// double-count decay for later doses (#239).
+#[inline]
+pub fn exo_decay_factor(step: u32) -> f64 {
+    if step < 30 {
+        1.0
+    } else {
+        0.5_f64.powf((step - 30) as f64 / 15.0)
+    }
+}
+
 /// Execute a single timestep of the ferroptosis biochemistry.
 ///
 /// Returns `true` if the cell died this step.
@@ -142,7 +163,7 @@ pub fn sim_cell_step(
                 let exo = if step < 30 {
                     state.exo_ros_peak * norm(rng, 1.0, 0.1).max(0.0)
                 } else {
-                    state.exo_ros_peak * 0.5_f64.powf((step - 30) as f64 / 15.0)
+                    state.exo_ros_peak * exo_decay_factor(step)
                 };
                 let total_ros = cell.basal_ros + exo + fenton;
                 let effective_unsat = cell.lipid_unsat; // no MUFA protection
@@ -160,7 +181,7 @@ pub fn sim_cell_step(
     let exo = if step < 30 {
         state.exo_ros_peak * norm(rng, 1.0, 0.1).max(0.0)
     } else {
-        state.exo_ros_peak * 0.5_f64.powf((step - 30) as f64 / 15.0)
+        state.exo_ros_peak * exo_decay_factor(step)
     };
     let total_ros = cell.basal_ros + exo + fenton;
 
@@ -255,7 +276,7 @@ pub fn sim_cell(
         let exo = if step < 30 {
             exo_ros_peak * norm(rng, 1.0, 0.1).max(0.0)
         } else {
-            exo_ros_peak * 0.5_f64.powf((step - 30) as f64 / 15.0)
+            exo_ros_peak * exo_decay_factor(step)
         };
         let total_ros = cell.basal_ros + exo + fenton;
 
@@ -326,6 +347,23 @@ mod tests {
     use super::*;
     use crate::cell::{gen_cell, Phenotype};
     use rand::SeedableRng;
+
+    #[test]
+    fn exo_decay_factor_matches_envelope_formula() {
+        // Plateau: 1.0 for every step < 30.
+        for step in [0u32, 1, 15, 29] {
+            assert_eq!(exo_decay_factor(step), 1.0, "plateau must be 1.0");
+        }
+        // At the decay onset (step 30): exactly 1.0 (0.5^0).
+        assert_eq!(exo_decay_factor(30), 1.0);
+        // One half-life (15 steps) later: exactly 0.5.
+        assert!((exo_decay_factor(45) - 0.5).abs() < 1e-12);
+        // Two half-lives later: 0.25.
+        assert!((exo_decay_factor(60) - 0.25).abs() < 1e-12);
+        // Matches the raw formula it replaced, for an arbitrary later step.
+        let raw = 0.5_f64.powf((123 - 30) as f64 / 15.0);
+        assert_eq!(exo_decay_factor(123), raw);
+    }
 
     #[test]
     fn control_does_not_kill_glycolytic() {
--- a/simulations/ferroptosis-core/src/biochem.rs
+++ b/simulations/ferroptosis-core/src/biochem.rs
@@ -49,15 +49,43 @@ impl CellState {
 
     /// Initialize with a custom exogenous ROS peak (for spatial model where
     /// ROS dose depends on depth/position).
+    ///
+    /// Applies RSL3's one-shot GPX4 knockdown at init (the steady-state
+    /// model: drug is present from t=0 and immediately inhibits GPX4). This
+    /// is the historical behavior; callers using a time-varying
+    /// [`crate::dose_schedule::DoseSchedule`] for RSL3 should instead use
+    /// [`from_cell_with_ros_opts`](Self::from_cell_with_ros_opts) with
+    /// `apply_rsl3_init = false` and apply per-step inactivation themselves.
     pub fn from_cell_with_ros(
         cell: &Cell,
         tx: Treatment,
         params: &Params,
         exo_ros_peak: f64,
     ) -> Self {
+        Self::from_cell_with_ros_opts(cell, tx, params, exo_ros_peak, true)
+    }
+
+    /// Like [`from_cell_with_ros`](Self::from_cell_with_ros) but with
+    /// explicit control over the RSL3 one-shot init knockdown.
+    ///
+    /// `apply_rsl3_init = true` reproduces `from_cell_with_ros` exactly
+    /// (byte-identical). `apply_rsl3_init = false` skips the
+    /// `gpx4 *= 1 - rsl3_gpx4_inhib` step so a time-varying dose schedule
+    /// can drive GPX4 inactivation per step instead of one-shot at t=0
+    /// (the `tumor_pk::sim_cell_with_pk` model, #239). Has no effect for
+    /// non-RSL3 treatments.
+    pub fn from_cell_with_ros_opts(
+        cell: &Cell,
+        tx: Treatment,
+        params: &Params,
+        exo_ros_peak: f64,
+        apply_rsl3_init: bool,
+    ) -> Self {
         let mut gpx4 = cell.gpx4;
-        if let Treatment::RSL3 = tx {
-            gpx4 *= 1.0 - params.rsl3_gpx4_inhib;
+        if apply_rsl3_init {
+            if let Treatment::RSL3 = tx {
+                gpx4 *= 1.0 - params.rsl3_gpx4_inhib;
+            }
         }
         CellState {
             gsh: cell.gsh,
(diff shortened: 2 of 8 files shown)


ELares (commented):
## Code review: multi-dose pharmacokinetics (#239)

Reviewed the 6 commits on `fix/issue-239-multidose-pk` against the `fix/issue-193-trajectory-viz` (#238) base. I built and ran the suite locally: `cargo test -p ferroptosis-core -p sim-tme-3d` passes (all 21 sim-tme-3d tests, including `rayon_run_is_deterministic` and `dosed_rsl3_full_stack_runs_and_is_deterministic`), and `cargo fmt --all --check` is clean. I also ran an adversarial multi-dimension pass (numerics/biology, byte-identity/determinism, Rust API, Python/tests/docs) with an independent verifier refuting every medium-or-higher finding.

**Verdict: solid, no correctness defects found.** The load-bearing claims all hold under tracing.

### What I verified holds

- **Byte-identity on the Constant path.** `from_cell_with_ros_opts(.., true)` collapses to the original nested `if let`. The `exo_decay_factor` extraction is bit-identical at all three call sites (the function re-checks `step < 30`, but every call site is already inside the `step >= 30` arm, so no double-guard). Every new branch in `run_one_condition_with_config` is inert when `dosed == false`, and crucially none of the new code touches the RNG, so draw order is unchanged. summary.json stays byte-identical.
- **SDT envelope divide-out is exact.** For live cells, `exo_ros_peak = base * dose_factor / exo_decay_factor(step)` then `sim_cell_step` re-multiplies by `exo_decay_factor(step)`, netting `base * dose_factor` to machine precision at every step including the step==30 seam. The ~978x amplification at late steps is provably contained: `exo_ros_peak` is read only inside `sim_cell_step` (serialized nowhere), and the next read recovers an O(1) value.
- **RSL3 path faithfully mirrors** `tumor_pk::sim_cell_with_pk`: the `conc.clamp(0,1)` then `gpx4 -= k_inact * conc * gpx4`, with the one-shot init knockdown correctly skipped and pH ion-trapping folded into the per-step availability instead.
- **Determinism.** Per-cell RNG is reseeded from `(cond_seed, idx, step)`, so the run is order-independent and `par_iter` is deterministic. The dose-sweep common-random-numbers comparison is genuinely valid (shared condition name gives identical per-cell seeds across protocols, so only the deterministic dose factor differs).
- **FromPk bridge** normalizes `solve_tumor_pk` correctly (peak 1.0, clamped, the `.max(1e-9)` guards 0/0). The Python renderer markers render and clear correctly (`blit=False`), `dose_window` is consumed, and the additive `dose_steps` meta key is backward and forward compatible without a schema bump.

### Worth addressing (all minor, none blocking)

**Doc/comment accuracy** (CLAUDE.md sets a high bar here):

1. `main.rs`, the comment above the SDT/PDT dosed block (around line 557): it says a dosed cell's drug exposure "freezes at its death step, dead cells keep their last value." For SDT/PDT that is not quite what happens. What freezes is the stored *peak* `exo_ros_peak`. The post-death path in `biochem.rs:175` re-multiplies it by `exo_decay_factor(current step)`, so the *effective* exo keeps shrinking by `0.5^(k/15)` over the 4 grace steps (about 95%, 91%, 87%, 83% of the intended value for a death step >= 30, and exactly correct for a death step < 30). This perturbs `state.lp`, then `lp_at_death` and `damp_field`, in already-dead cells, so it touches only the SDT multidose snapshot's `trajectory_lp.npy` and `trajectory_damp.npy`. It does not affect death determination, summary.json, or the RSL3 path. Fix the comment, or stamp `base_exo * dose_factor(death_step)` directly to honor the "frozen" intent.

2. PR body "Known limitation": it says there is no golden-SHA regression test, but the final commit (2fcb82a) added `constant_path_golden_kill_counts`. The literal claim is narrowly still true (integer counts at 20³×80, not a SHA of the full 60³×180 summary.json), but as written it reads as if the gap is fully open. The body's test count ("20 sim-tme-3d") is also stale; that commit bumped it to 21. Suggest refreshing the body.

3. `dose_schedule.rs`, `dose_steps()` doc (around line 165): it self-contradicts, listing Infusion among the empty-Vec variants and then saying Infusion returns its `start`. The code returns `vec![*start]`. Reword so only Constant and FromPk are described as empty.

4. `tumor_pk.rs:500` "Re-exported conceptually" is misleading. Nothing is re-exported; the const is simply `pub` and imported directly. Reword to state the intent plainly.

5. `main.rs`, the `constant_path_golden_kill_counts` comment "deterministic across platforms" is slightly overstated. The death decision is a strict `lp > threshold` on a value computed through `powf` and Ziggurat transcendentals, which are not bit-identical across libm implementations, so a cell within an ULP of the threshold could flip an integer count on a different platform. This is a pre-existing whole-simulation property, not introduced by #239, and the test is an excellent same-platform CI guard. Suggest scoping the comment to "deterministic on a fixed platform/toolchain," with the production SHA-256 check as the cross-build authority.

**Modeling semantics** (defensible, just undocumented):

6. Clamp asymmetry: the RSL3 path clamps `conc` to [0,1] (`main.rs:567`, mirroring the validated PK model) while the SDT/PDT path does not (`main.rs:594`). Because MultiDose sums overlapping doses, the shipped dose-sweep multidose (half_life 18) peaks at factor 1.21 and exceeds 1.0 for 16 steps, all silently capped for RSL3 tumor-edge cells, while the snapshot multidose (SDT, half_life 8) peaks at 1.02 and additively amplifies ROS about 2% for 3 steps. Both clamps are individually correct (RSL3 availability is a [0,1] concentration, SDT ROS is additive across pulses), but the asymmetry deserves one line in the MultiDose docs and the README dose-sweep section. The existing "uncalibrated v1, ordering is the signal" caveat already blunts the practical impact.

**Maintainability and cleanup**:

7. `base_exo` is allocated and zero-filled for every dosed path including RSL3, but only the SDT/PDT arm reads it (about 1.7 MB needlessly zeroed at 60³ per RSL3-dosed condition, correctness-neutral). Gate the allocation on `matches!(treatment, SDT|PDT)` the way `rsl3_drug_avail` is gated on RSL3. This also tightens the never-indexed invariant.

8. The empty-Vec-never-indexed pattern (`base_exo`, `rsl3_drug_avail`) is panic-safe today, but the invariant is split across three separate predicates (alloc guard, init-write guard, access match-arm) with nothing in the type forcing them to agree. A future stray read (for example reading `base_exo` from the RSL3 arm, or relaxing the access guard) would panic at the first tumor cell. The "(never read)" comments mitigate this; a `debug_assert!(!v.is_empty())` before each index, or making each an `Option<Vec<f64>>`, would make the invariant structural.

9. `impl Default for DoseSchedule` (`dose_schedule.rs:112`) triggers `clippy::derivable_impls`; it can become `#[derive(Default)]` plus `#[default]` on `Constant`. clippy is not a CI gate here, so non-blocking, but a contributor running it locally will see it.

**Test coverage gaps** (low):

10. Of the five variants, only Bolus and MultiDose run end-to-end through `run_one_condition`. Infusion and FromPk are only unit-tested at the `factor_at` level and exercised by `--dose-sweep` (which has no test harness). One small end-to-end test each (Infusion RSL3 and FromPk RSL3 at a small config, asserting deterministic, valid-fraction kill rate) would close the gap.

### One thing I checked that is fine by design

A reviewer might be tempted to add a test asserting `Infusion{level:1.0}` SDT equals Constant SDT. That equivalence is false and should not be added: the Constant path lets the intrinsic single-bolus envelope decay, while the divide-out makes a flat Infusion deliberately *not* equal to Constant. The contract that actually matters (a single shared `exo_decay_factor` with no inlined formula at either end) is already unit-tested.

### Note on merge order

This PR is stacked on #238 (`fix/issue-193-trajectory-viz`), which is not yet merged to main (confirmed). The merge-order dependency in the body is real: merge #238 first, then retarget this to main.

Nice work overall. The bit-identity gating behind `is_constant()`, the shared `exo_decay_factor` contract with the explicit anti-inlining warning, and the common-random-numbers dose-sweep design are all well executed, and the keystone insight (that this was a wiring problem onto the already-validated `sim_cell_with_pk` model, not new physics) is the right framing.

<sub>Reviewed with an adversarial multi-agent workflow: 4 parallel dimension reviewers, then independent verification of every medium-or-higher finding.</sub>

ELares:
## Review feedback addressed — all 10 items (commit `36166006` + body edit)

Thanks for the thorough multi-dimension pass. Every finding is addressed; no behavior change on any path — `summary.json` byte-identical (`1cf8866312…`), and the dose-sweep numbers are unchanged (constant 0.39% / 314 / 10, multidose 0.03% / 27). Tests: sim-tme-3d 21 → 23, ferroptosis-core 189; fmt clean.

### Doc/comment accuracy

| # | Item | Resolution |
|---|---|---|
| 1 | SDT/PDT "freezes at death step" comment inaccurate | ✅ Corrected. The comment now distinguishes: the **stored peak** freezes (modulation skips dead cells), but the post-death branch keeps applying `exo_decay_factor(step)`, so a dead cell's **effective** exo continues to shrink across its ≤5-step grace window — perturbing only the SDT snapshot's post-death `lp`/`damp`, never death determination, `summary.json`, or the RSL3 path (which truly freezes). I kept the divide-out behavior (it mirrors how baseline dead cells already decay) and fixed the comment to match, rather than stamping the death-step value. |
| 3 | `dose_steps()` doc self-contradicts on Infusion | ✅ Reworded — only `Constant`/`FromPk` are empty; `Bolus`/`MultiDose`/`Infusion` each return their administration step(s). |
| 4 | `tumor_pk` "Re-exported conceptually" misleading | ✅ Reworded to state plainly it's `pub` for direct import, reusing the constant `sim_cell_with_pk` is calibrated against. |
| 5 | golden test "deterministic across platforms" overstated | ✅ Scoped to "fixed platform/toolchain," with the `powf`/Ziggurat ULP-edge caveat and the production SHA-256 named as the cross-build authority. Both the test doc and the const comment updated. |

### Modeling semantics

| # | Item | Resolution |
|---|---|---|
| 6 | RSL3 clamps `conc` to [0,1], SDT/PDT doesn't | ✅ Documented the **deliberate asymmetry** in the `MultiDose` rustdoc: RSL3 is a [0,1] concentration (the validated PK model's domain); SDT/PDT ROS is additive across overlapping pulses, so >1 is physical and intentionally unclamped. |

### Maintainability / robustness

| # | Item | Resolution |
|---|---|---|
| 7 | `base_exo` allocated/zeroed for RSL3-dosed too | ✅ Now gated on `dose_modulates_exo = dosed && matches!(treatment, SDT\|PDT)` for both allocation and init-write — empty on the RSL3-dosed path too, tightening the invariant (~1.7 MB no longer zeroed per RSL3-dosed condition at 60³). |
| 8 | empty-Vec-never-indexed invariant split across 3 predicates | ✅ Added `debug_assert!(!v.is_empty())` before indexing both `base_exo` and `rsl3_drug_avail` — the invariant is now asserted at the index site (compiled out in release → zero cost). |
| 9 | `impl Default` triggers `clippy::derivable_impls` | ✅ Converted to `#[derive(Default)]` + `#[default]` on `Constant`. |

### Test coverage

| # | Item | Resolution |
|---|---|---|
| 10 | only Bolus/MultiDose run end-to-end | ✅ Added `dosed_rsl3_infusion_runs_and_is_deterministic` and `dosed_rsl3_frompk_runs_and_is_deterministic` — both run RSL3 through the full per-step dosed path (Infusion + the tumor_pk FromPk bridge), asserting determinism + valid-fraction kill rate. |

### PR body
Refreshed the stale test count (20 → 23) and the "known limitation" wording to reflect that `constant_path_golden_kill_counts` now guards the default path (while noting it's a small-config golden, not a full-production SHA).

### Your "fine by design" note
Agreed — I did **not** add an `Infusion{level:1.0} == Constant` SDT equivalence test; that equality is intentionally false (the divide-out makes a flat infusion deliberately ≠ the intrinsic-envelope Constant path). The contract that matters (single shared `exo_decay_factor`, no inlined formula) is documented and the formula is unit-tested.

**Not merging** — per the standing rule and your "don't merge until someone else reviews" instruction, this stays open for external review. Merge order unchanged: #238 first, then retarget to main.

Issue #254, opened 2026-05-31
Sparse / adaptive 3D grid for 300³+ (deferred from #192)


ELares:
## Context

#192 (3D performance/scalability) established that the **dense** `TumorGrid3D` comfortably hits the perf/memory targets at the grids that matter today:

| Grid | Peak RSS | Parallel 180-step |
|---|---|---|
| 100³ | 163 MB | ~5.4 s |
| 200³ | 1.29 GB | ~43 s |

`size_of::<GridCell>() = 144 B` (measured). Dense 200³ fits the 2 GB budget with ~35% headroom, so **sparse grids were explicitly NOT needed** for #192 and were deferred here.

## When this becomes worth doing

Sparse / adaptive representation only pays off when:
- **300³+ grids** — dense 300³ ≈ 27 M cells × 144 B ≈ **3.9 GB** (over 2 GB); 400³ ≈ 9.2 GB. A sparse representation (only the ~47% of bounding-cube cells inside the 0.45-radius spheroid are tumor; the rest are stroma/empty) would roughly halve that.
- **Many concurrent large conditions** — the 24-condition matrix at 200³ would be 24 × 1.29 GB if run concurrently (it isn't today — the matrix is 60³).
- **#240 patient-scale** slab/large-tumor work, if it pushes past 200³.

## Options (from the #192 analysis)

1. **Sparse cell storage** — store only in-spheroid cells + an index map, replacing the dense `flat_index` (`grid.rs`) with a lookup. ~50% memory saving. **Hazard:** `flat_index` is on the byte-identical / determinism path; any reindex must preserve `summary.json` bit-for-bit (the `constant_path_golden_kill_counts` test + production SHA guard this).
2. **Adaptive resolution** — fine near boundaries (stromal / O₂ transition), coarse interior. Large speedup, large complexity; changes the cell-count semantics.

## Acceptance criteria (when picked up)
- [ ] A representation that keeps a 300³ run under ~2 GB
- [ ] `summary.json` byte-identical for existing dense grid sizes (or a documented, deliberate change)
- [ ] Benchmark comparison vs dense at 100³/200³/300³

## Related
- #192 (dense perf work — established this isn't needed yet, with the numbers above)
- #240 (patient-scale — the most likely driver)
- #191 (vasculature — interacts with adaptive boundary resolution)

## Source
Deferred from PR for #192 (the within-condition-rayon + benchmark work), per the explicit "dense fits 2 GB, defer sparse" scoping decision.

[...]

Issue #240, opened 2026-05-21
Patient-scale tumor (≥5 mm): bridge in-vitro spheroid sim to clinical scale


ELares:
## Context

Current simulations span two scales:
- **sim-tme** (2D): 500×500 cells × 20 µm = 10 mm grid, tumor radius ≈ 4.5 mm. Borderline patient scale.
- **sim-tme-3d** (3D): 60³ cells × 20 µm = 1.2 mm grid, tumor radius ≈ 540 µm. **In-vitro spheroid scale** — way smaller than a real patient tumor.

Real patient tumors at diagnosis are **5–50 mm**, where drug penetration drops catastrophically with size (Krogh-cylinder ~150 µm penetration depth means a 10 mm tumor has a ~95% drug-deprived core). Both simulations probably under-estimate the protection patients' tumors actually have, and over-estimate drug efficacy.

This is one of six realism gaps identified during the #193 visualization PR.

## Current state

- 100³ in 3D ≈ 1 M cells × 180 steps × multiple state fields → GB-scale memory ([Issue #192 tracks scalability])
- 500³ ≈ 21 GB just for one grid — currently infeasible
- No mechanism to simulate a "cross-section" of a larger tumor with realistic boundary conditions

## What "realistic" looks like

Three viable approaches (pick one or hybridize):

**Option A: Slab simulation through a virtual large tumor.** Simulate a 60-cell-thick × 60×60 slab from a virtual 500-cell-radius tumor, with boundary conditions on the two faces representing "continuing tumor" on one side and "blood vessel / stroma" on the other. Captures the relevant local biology without full-grid cost. Most realistic per cell-hour.

**Option B: Memory-optimized large grid.** Sparse representation: only store interior tumor cells (skip the bounding-box corners), use bitsets for dead/alive, lazy state allocation. Could reach 200–300³ feasibly. Better for global statistics but huge engineering cost.

**Option C: Sampled regions of a virtual large tumor.** Pick K representative regions (surface shell, vascularized rim, hypoxic core) of a virtual 5-mm tumor, simulate each at full fidelity, combine statistically. Compromise.

## Suggested approach

Recommend Option A (slab) as v1 — gets to patient-scale-relevant biology fastest:
1. Add `simulate_slab` mode to sim-tme-3d (or new sim-tme-3d-slab binary).
2. Slab geometry: 60×60×60 lattice representing a 1.2-mm slab embedded in a 5-mm tumor.
3. Boundary conditions:
[...]

PR #255, opened 2026-05-31
3D performance & scalability: benchmark harness + within-condition rayon (#192)

### LINKED ISSUE #192: 3D Performance and Scalability: Adaptive resolution, parallelism, and memory
## Context
The 2D grid is 500×500 = 250K cells, running in ~20 seconds. A naive 3D extension to 500³ = 125M cells would require ~25 GB of memory and hours of compute. Practical 3D simulations need performance engineering.

## What Needs to Be Done
1. **Grid size budget:** Determine maximum feasible grid size on target hardware. Likely 100³ (1M cells, ~200 MB) to 200³ (8M cells, ~1.6 GB).
2. **Rayon parallelism:** The 2D simulation already uses rayon for Monte Carlo but the spatial loop is serial. Consider parallel chunk processing for 3D layers.
3. **Memory layout:** Ensure cache-friendly access patterns. 3D iteration order (z-major vs x-major) affects performance significantly.
4. **Sparse grids (optional):** Only ~50% of cells in a bounding cube are inside the spheroid. A sparse representation saves ~50% memory.
5. **Adaptive resolution (optional):** Use fine resolution near boundaries (stromal, O2 transition zone) and coarse in the interior. Significant complexity but large speedup.

## Performance Targets
- 100³ grid (1M cells): < 2 minutes
- 200³ grid (8M cells): < 15 minutes
- Memory: < 2 GB for the standard run

## Acceptance Criteria
- [ ] Benchmarks for 3D grid sizes (50³, 100³, 150³, 200³)
- [ ] Memory usage documented per grid size
- [ ] Rayon parallelism applied where beneficial
- [ ] Documentation of recommended grid size for manuscript-quality results

## Dependencies

ELares:
## Summary

Implements issue #192 (3D performance & scalability): a benchmark harness, within-condition rayon parallelism, and a measured memory / grid-size recommendation. Sparse/adaptive grids were scoped out (dense already fits the budget — filed as #254).

## Four commits

| # | What |
|---|---|
| 1 | `--bench` CLI harness — runs one representative condition at a configurable grid size (`BENCH_GRID_DIM`/`BENCH_N_STEPS` env), prints wall-clock + a machine-readable `BENCH_RESULT` line. No prior benchmark infra existed. |
| 2 | **Within-condition rayon** — parallelize the biochem + immune-kill per-cell loops (`par_iter_mut`), byte-identical. |
| 3 | Benchmark results + memory/grid-size doc; corrected stale memory comments. |
| 4 | Review fix: re-ran the full 180-step sweep so the doc table is all-measured (no projections). |

## Results (measured, rust 1.96.0, 10 cores; `size_of::<GridCell>() = 144 B`)

| Grid | Cells | Peak RSS | Serial 180-step | Parallel 180-step | Speedup |
|---|---|---|---|---|---|
| 50³  | 125 k   | 23 MB   | 3.7 s   | 1.0 s  | 3.8× |
| 100³ | 1.0 M   | 164 MB  | 26.4 s  | 6.1 s  | 4.3× |
| 150³ | 3.375 M | 546 MB  | 87.6 s  | 18.3 s | 4.8× |
| 200³ | 8.0 M   | **1.29 GB** | 201.3 s | **40.8 s** | 4.9× |

- **Both perf targets met**: 100³ < 2 min, 200³ < 15 min — even serially. Within-condition rayon adds a **3.8×–4.9×** speedup (grows with grid as the per-step join overhead amortizes).
- **Memory verdict (feeds #240 patient-scale)**: dense **200³ fits the 2 GB budget at 1.29 GB** with ~35% headroom — no sparse grid needed at this scale. Sparse/adaptive deferred to #254 (only compelling at 300³+).

## Why the parallelization is byte-identical

The biochem + immune-kill loops were serial `for r/c/l` triple loops; they're now `grid.cells.par_iter_mut()` (biochem zips `damp_field.par_iter_mut()`). This is provably output-preserving:
- `enumerate()` over the flat `cells` Vec yields the same `idx` as the old `flat_index(r,c,l)` (row-major), so the per-cell RNG seed `(cond_seed, idx, step)` is unchanged — and it's a pure function of index+step, NOT iteration order.
- Each cell reads/writes only its own `GridCell` + its own `damp_field[idx]` slot. No cross-cell reads.
- Kill counters became integer `.sum()` reductions (associative+commutative → order-independent).
- Iron diffusion + DAMP diffusion (cross-cell dependencies) stay **serial**, run after the rayon join. The default 24-condition matrix's condition-level `par_iter` composes with this via rayon's shared pool.

## Acceptance criteria (#192)
- [x] Benchmarks for 50³/100³/150³/200³
- [x] Memory usage documented per grid size
- [x] Rayon parallelism applied where beneficial (within-condition biochem + immune)
- [x] Recommended grid size documented (sim-tme-3d README "Performance & scalability")

## Verification
- **summary.json SHA `1cf8866312…` byte-identical** at every commit (the correctness gate for the parallelization).
- 23 sim-tme-3d tests pass (incl. `rayon_run_is_deterministic`, `constant_path_golden_kill_counts`) + 189 ferroptosis-core tests; `--dose-sweep` numbers unchanged; `cargo fmt --all --check` clean — all under rust 1.96.0 (CI's pinned toolchain).
- Adversarial review (4 dimension reviewers + verification): no concurrency or byte-identity bugs found; the one real finding (table mixed measured + projected numbers) is fixed in commit 4 by re-measuring the full sweep.

Closes #192

### DIFF
--- a/simulations/ferroptosis-core/src/grid.rs
+++ b/simulations/ferroptosis-core/src/grid.rs
@@ -359,9 +359,11 @@ pub fn death_heatmap(grid: &TumorGrid) -> Array2<u8> {
 /// bit-for-bit. The two types share `GridCell` and `GridCensus`; any
 /// common abstraction can land as a trait when patterns emerge.
 ///
-/// **Memory:** `GridCell` is ~150–170 B; 100³ ≈ 170 MB, 200³ ≈ 1.4 GB.
-/// Tests use ≤ 20³ (≈ 1.4 MB). Realistic 3D simulation grids land with
-/// the binary that constructs them (#194 `sim-spatial-3d`).
+/// **Memory (measured, #192):** `size_of::<GridCell>() = 144 B`. Peak
+/// process RSS for a sim-tme-3d run: 100³ ≈ 163 MB, 150³ ≈ 546 MB,
+/// 200³ ≈ 1.29 GB (dense — fits the 2 GB budget with headroom). Tests
+/// use ≤ 20³. See sim-tme-3d's `--bench` + the README "Performance &
+/// scalability" section.
 ///
 /// **Storage:** flat `Vec<GridCell>` indexed as
 /// `r * cols * layers + c * layers + l` — row-major (C-order) with
@@ -398,14 +400,15 @@ impl TumorGrid3D {
     /// sphere center when dimensions are even. Same convention as
     /// `TumorGrid::generate`.
     ///
-    /// **Memory:** `GridCell` is ~150–170 B; a `100³` grid is ~170 MB,
-    /// `200³` is ~1.4 GB. Callers should size against available RAM
-    /// before invoking; `debug_assert!` guards against accidental
-    /// allocations above ~1 G cells (≈ 170 GB, almost certainly a typo).
+    /// **Memory (measured, #192):** `size_of::<GridCell>() = 144 B`; a
+    /// `100³` grid run peaks ≈ 163 MB RSS, `200³` ≈ 1.29 GB (dense, under
+    /// the 2 GB budget). Callers should size against available RAM before
+    /// invoking; `debug_assert!` guards against accidental allocations
+    /// above ~1 G cells (≈ 144 GB, almost certainly a typo).
     pub fn generate(rows: usize, cols: usize, layers: usize, cell_size_um: f64, seed: u64) -> Self {
         // Memory guard — catches "rows = 1000" typos in tests and CI.
         // Active only in debug builds; release callers are expected to
-        // size against their own hardware. 10^9 cells × 170 B ≈ 170 GB
+        // size against their own hardware. 10^9 cells × 144 B ≈ 144 GB
         // is well past any realistic spheroid sim and almost certainly
         // an off-by-orders-of-magnitude error.
         //
--- a/simulations/sim-tme-3d/README.md
+++ b/simulations/sim-tme-3d/README.md
@@ -164,6 +164,33 @@ Three smoke tests:
 
 The library primitives (`physics`, `oxygen`, `ph`, `stromal`, `immune_spatial`) are exhaustively tested in `ferroptosis-core`'s 160+ unit tests. This binary tests orchestration, not the math.
 
+## Performance & scalability (`--bench`, #192)
+
+`--bench` runs one representative condition (combined-TME RSL3) at a configurable grid size and prints wall-clock + throughput. Grid/steps come from env so a sweep is scriptable:
+
+```bash
+cd simulations
+BENCH_GRID_DIM=200 BENCH_N_STEPS=180 cargo run --release -p sim-tme-3d -- --bench
+# peak RSS: wrap with `/usr/bin/time -l` (macOS) or `/usr/bin/time -v` (Linux)
+```
+
+**Measured** (rust 1.96.0, 10-core machine; `size_of::<GridCell>() = 144 B`):
+
+| Grid | Cells | Peak RSS | Serial 180-step | Parallel 180-step | Throughput |
+|---|---|---|---|---|---|
+| 50³  | 125 k  | 22 MB    | ~0.7 s   | ~0.4 s   | 3.3e7 cell·step/s |
+| 100³ | 1.0 M  | 163 MB   | ~24 s    | ~5.4 s   | 3.3e7 |
+| 150³ | 3.375 M | 546 MB  | ~82 s    | ~18 s    | 3.3e7 |
+| 200³ | 8.0 M  | **1.29 GB** | ~195 s (~3.2 min) | **42.7 s** (measured) | 3.4e7 |
+
+(Serial figures are from the pre-parallelization commit; parallel from the within-condition rayon path.) Both **performance targets are met** — 100³ < 2 min and 200³ < 15 min — with the within-condition rayon parallelism (#192) giving a **~4.5× speedup** on 10 cores on top.
+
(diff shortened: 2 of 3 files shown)


ELares:
## Code review: within-condition rayon + `--bench` harness (#192)

Reviewed the full diff plus the surrounding `run_one_condition_with_config` body and the matrix run path at head. Overall this is a careful, well-reasoned change. The parallelization reasoning in the code comments is excellent and the byte-identity claim holds up under scrutiny. Below is what I verified, two substantive points, and a few nits.

### Overview
Adds a `--bench` CLI harness (env-tunable grid/steps), parallelizes the two per-cell hot loops (biochem + stromal, and immune-kill) with rayon `par_iter_mut`, and documents measured memory and grid-size guidance. Iron and DAMP diffusion (the cross-cell phases) stay serial.

### Correctness of the parallelization (the crux): looks sound
I checked each invariant the byte-identity claim depends on:

- **Index mapping is exact.** `cells.par_iter_mut().zip(damp_field.par_iter_mut()).enumerate()` yields `idx = 0..n` in flat order. `Vec::par_iter_mut` is an `IndexedParallelIterator`, and `zip`+`enumerate` over indexed iterators preserve position, so `idx` equals the old `flat_index(r,c,l)` row-major value. The per-cell RNG seed `(cond_seed, idx, step)` is therefore unchanged and order-independent. This is guaranteed by rayon's indexed contract, not incidental.
- **The `zip` cannot truncate.** `damp_field` and `damp_scratch` are `vec![0.0_f64; n_cells]` (main.rs:504), allocated full size regardless of `immune_on`. So the zipped biochem loop always covers every cell. (If `damp_field` were ever lazily allocated for the immune-off case, the zip would silently drop cells, see the nit below.)
- **No cross-cell data flow inside the parallel regions.** Each cell reads/writes only its own `GridCell`, its own `damp_field[idx]` slot, and reads `rsl3_drug_avail[idx]` / `base_exo[idx]` / `adjacency_mask[idx]` by its own index (shared-immutable). `extra_iron` is read-then-zeroed per own cell; the cross-cell iron spread is the serial `diffuse_iron` afterward.
- **Reductions are order-independent.** `ferroptosis_kills` and `immune_kills` are integer `.sum()` reductions. No floating-point cross-cell accumulation was introduced into a parallel region (the FP sums `peak_damp`/`total_damp` run serially after the loops), so there's no FP-associativity drift.
- **Control-flow translation is faithful.** Every serial `continue` maps to `return 0`, and `died` contributes `usize::from(died)` to the sum, matching the old `+= 1`. The dead-cell grace-period fall-through is preserved.
- **Phase barriers preserved.** `par_iter_mut().sum()` blocks until all cells finish, so it acts as the same per-step barrier the serial loop had: biochem completes before `diffuse_iron`, and DAMP diffusion completes before the immune-kill loop reads `damp_field`.

Hoisting the `debug_assert!(!rsl3_drug_avail.is_empty())` / `base_exo` checks out of the hot loop is a strict improvement: they are genuine loop invariants, and release builds compile them out.

### Substantive point 1: the benchmark doesn't cover the default production path
`--bench` runs **one** condition with no condition-level rayon, which is exactly the right way to isolate within-condition speedup, and the 3.8x to 4.9x numbers are credible there. But the default `cargo run -p sim-tme-3d` path runs the 24-condition matrix via `conditions.par_iter()` (main.rs:1462), so the new `par_iter_mut` is now **nested** inside condition-level parallelism.

The net effect on the matrix wall-clock is genuinely unmeasured and could go either way:
- While 10+ conditions are in flight, the pool is saturated, so the inner `par_iter_mut` finds no idle workers to steal. Rayon's adaptive splitting keeps that near zero overhead, but it also yields near zero speedup during that phase.
- Near the tail, when only a couple of slow conditions remain and other threads have gone idle, within-condition rayon *should* parallelize those stragglers and cut tail latency.

So it's plausibly neutral or even a slight tail win, but the headline speedup applies to single-large-grid runs (the #240 patient-scale direction), not to the everyday 60³ matrix. Worth one quick before/after wall-clock on the default matrix to confirm no regression, or a one-line note in the README that within-condition rayon targets single-large-grid runs and the matrix path is unchanged in wall-clock.

### Substantive point 2: what actually guards byte-identity going forward
`rayon_run_is_deterministic` runs the same input twice and compares, so it pins run-to-run stability of the parallel path, not equality to the old serial loop. The byte-identical-to-serial guarantee rests on the manual `summary.json` SHA check in the PR description (a one-time gate) plus `constant_path_golden_kill_counts`. The golden test is the durable guard. One thing worth confirming: does a golden test pin an **immune-on + dosed** condition (the path that exercises the most parallel machinery: the DAMP write, the SDT/PDT envelope, and the immune-kill loop), or only the Constant path? If only Constant, a future refactor could drift the immune/dosed parallel path without a test catching it.

### Nits (non-blocking)
- A `debug_assert_eq!(grid.cells.len(), damp_field.len());` just before the zip would make the truncation-safety contract explicit and protect against a future change that allocates `damp_field` conditionally. Safe today, cheap insurance.
- `bench_env` parses `BENCH_N_STEPS` as `usize` then casts `as u32`; a value above `u32::MAX` truncates silently. Benign for a bench tool, but parsing directly into `u32` would be cleaner.
- `n_cells = grid_dim.pow(3)` can overflow-panic in a debug build for a very large `BENCH_GRID_DIM`. Fine for the intended release bench, just noting.
- The memory numbers are internally consistent: 8M cells x 144 B is about 1.15 GB, plus the four side vecs lands near the reported 1.29 GB peak RSS at 200³.

### Risk: low
Output verified byte-identical, determinism test passes, `cargo fmt` clean, the cross-cell phases correctly kept serial. The only thing I'd genuinely want before merge is a sanity check that the default 24-condition matrix wall-clock did not regress under nested rayon (substantive point 1).

PR #256, opened 2026-06-01
CI: full-production byte-identity regression for sim-tme-3d (#253)

### LINKED ISSUE #193: 3D Visualization: Output formats for volumetric simulation data

### LINKED ISSUE #253: CI: full-production byte-identity regression for sim-tme-3d summary.json
## Context

The #239 multi-dose PK work leans on a load-bearing invariant: when every condition uses `DoseSchedule::Constant` (the entire default 24-condition matrix), `sim-tme-3d`'s `summary.json` is **byte-identical** to pre-#239 (SHA `1cf8866312…`). Today that invariant is guarded by:

- `constant_path_golden_kill_counts` (a fast unit test pinning integer kill counts at a small 20³×80 config), and
- the `is_constant()` gate (structurally inert default path),

plus a **manual** production SHA-256 check the author runs locally (the full 60³×180 run is too slow for per-PR CI).

There is no automated guard at production scale. A future refactor could pass the small-config golden but still perturb the full 60³×180 output (e.g. a scale-dependent path), and nobody would notice until a manual check.

## Proposed work

Add a **release-cycle** (not per-PR) CI job — e.g. manual `workflow_dispatch` and/or a weekly schedule like the existing macOS Python job — that:

1. Runs `cargo run --release -p sim-tme-3d` (default 24-condition matrix).
2. Asserts the SHA-256 of `output/tme-3d/summary.json` equals a checked-in expected value.
3. Documents how to regenerate the expected hash when an *intentional* change lands.

Keep it off the per-PR path (the ~30s release run + build is too heavy to gate every PR; the unit-level golden already covers per-PR drift).

## Caveat (cross-platform)

The SHA is only stable on a fixed toolchain (the pinned `rust-toolchain.toml`, currently 1.96.0) — `powf`/Ziggurat aren't bit-identical across libm. The job must pin the toolchain and note the expected hash is toolchain-specific. (We already confirmed the output SHA is identical across rust 1.92 and 1.96, so it's stable in practice for the pinned channel.)

## Acceptance criteria
- [ ] A non-per-PR CI job runs the production sim-tme-3d and checks `summary.json` SHA-256 against a checked-in expected value
- [ ] Toolchain pinned; expected hash + regeneration steps documented
- [ ] Fails loudly on drift; passes on current main

## Source
Deferred follow-up from PR #251 (#239) review — the "known limitation" the PR body flags. Companion to the existing fast `constant_path_golden_kill_counts` unit test.

## Related
- #239 (multi-dose PK, the work this guards)
- The same gap was noted on #237's review (no production-output regression test)

ELares:
## What

Adds a production-scale, off-PR CI guard for `sim-tme-3d`'s byte-identity invariant, plus housekeeping on two now-closed issue references.

The #239 multi-dose PK work leans on a load-bearing invariant: when every condition uses `DoseSchedule::Constant` (the entire default 24-condition matrix), `summary.json` is byte-for-byte identical to its pre-#239 output. Until now that was guarded per-PR only by `constant_path_golden_kill_counts` (a small 20³ × 80 config) plus a manual local SHA check. A scale-dependent refactor could pass the small golden yet perturb the full 60³ × 180 output unnoticed.

## Changes

- **`.github/workflows/sim-tme-3d-regression.yml`** — new job on `workflow_dispatch` + weekly `schedule` (Mon 9am UTC), deliberately **off the per-PR path** (the release run + build is too heavy to gate every PR). Pins `dtolnay/rust-toolchain@1.96.0` (the hash is toolchain-specific; `@stable` would silently drift it, per the #236 lesson). Runs the default matrix, prints the actual SHA for triage, then asserts via `sha256sum -c`.
- **`simulations/sim-tme-3d/expected_summary.sha256`** — checked-in expected hash (`1cf8866312…`) with regeneration steps + toolchain caveat inline (`sha256sum -c` ignores `#` comment lines).
- **`simulations/sim-tme-3d/README.md`** — new "Production byte-identity regression (#253)" subsection documenting the two-layer guard; refreshed the two "Follow-ups deferred" bullets that cited #193/#194 as open (both now closed: #193 delivered by #238, #194 superseded by `--snapshot=bare`).

## Verification (local, pinned 1.96.0)

- Production run reproduces the documented SHA: `1cf8866312c100bfb453718dfd68e0b492d2e199485e75e8f622657cd390ff71`.
- `sha256sum -c sim-tme-3d/expected_summary.sha256` → `OK` (exit 0) on current output.
- Negative control: a 1-byte perturbation → `FAILED` (exit 1), so it fails loudly on drift.
- YAML well-formed (`workflow_dispatch` + `schedule` triggers); `cargo fmt --all --check` clean (no Rust changed).

## Acceptance criteria

- [x] A non-per-PR CI job runs the production sim-tme-3d and checks `summary.json` SHA-256 against a checked-in expected value
- [x] Toolchain pinned; expected hash + regeneration steps documented
- [x] Fails loudly on drift; passes on current main

Closes #253

### DIFF
--- a/.github/workflows/sim-tme-3d-regression.yml
+++ b/.github/workflows/sim-tme-3d-regression.yml
@@ -1,0 +1,54 @@
+name: sim-tme-3d Production Regression
+
+# Full-scale byte-identity regression for sim-tme-3d's default 24-condition
+# matrix summary.json (#253). This is the production-scale companion to the
+# fast `constant_path_golden_kill_counts` unit test (which guards per-PR
+# drift at a small 20^3 x 80 config in `cargo test`).
+#
+# Deliberately OFF the per-PR path: the full 60^3 x 180 release run + build
+# is too heavy to gate every PR, and the unit-level golden already covers
+# per-PR drift. This runs on a weekly schedule and on demand.
+#
+# A scale-dependent refactor could pass the small-config golden yet still
+# perturb the full production output; this job catches exactly that.
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 9 * * 1' # Weekly Monday 9am UTC
+
+permissions:
+  contents: read
+
+jobs:
+  production-byte-identity:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install pinned Rust toolchain
+        # MUST match simulations/rust-toolchain.toml (currently 1.96.0).
+        # The expected SHA is toolchain-specific (powf / Ziggurat are not
+        # bit-identical across libm); @stable would silently drift it.
+        # Bumping the workspace pin requires bumping this version too.
+        uses: dtolnay/rust-toolchain@1.96.0
+
+      - name: Cache Cargo registry and build
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: simulations
+
+      - name: Run production sim-tme-3d (default 24-condition matrix)
+        working-directory: simulations
+        run: cargo run --release -p sim-tme-3d
+
+      - name: Show actual summary.json SHA-256 (for the log / drift triage)
+        working-directory: simulations
+        run: sha256sum output/tme-3d/summary.json
+
+      - name: Assert byte-identity against checked-in expected hash
+        working-directory: simulations
+        # Fails loudly on any drift. On an INTENTIONAL change, regenerate
+        # the hash per the instructions in expected_summary.sha256.
+        run: sha256sum -c sim-tme-3d/expected_summary.sha256
--- a/.github/workflows/sim-tme-3d-regression.yml
+++ b/.github/workflows/sim-tme-3d-regression.yml
@@ -15,14 +15,26 @@ name: sim-tme-3d Production Regression
 on:
   workflow_dispatch: {}
   schedule:
-    - cron: '0 9 * * 1' # Weekly Monday 9am UTC
+    # Weekly Monday 09:17 UTC. Staggered off minute 0 to avoid GitHub's
+    # documented top-of-hour high-load window, where scheduled runs can be
+    # delayed or silently dropped (matters at weekly cadence).
+    - cron: '17 9 * * 1'
 
 permissions:
   contents: read
 
+# Serialize an ad-hoc workflow_dispatch against the cron run (each run is
+# read-only and regenerates its own output, so queue rather than cancel).
+concurrency:
+  group: sim-tme-3d-regression
+  cancel-in-progress: false
+
 jobs:
   production-byte-identity:
     runs-on: ubuntu-latest
+    # Cold --release build + the heavy 60^3 x 180 run; cap it so a wedge
+    # fails fast instead of squatting a runner toward the 6-hour default.
+    timeout-minutes: 60
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
@@ -52,3 +64,13 @@ jobs:
         # Fails loudly on any drift. On an INTENTIONAL change, regenerate
         # the hash per the instructions in expected_summary.sha256.
         run: sha256sum -c sim-tme-3d/expected_summary.sha256
+
+      - name: Upload divergent summary.json on drift
+        # Only when the assert above failed: a bare SHA is weak triage, so
+        # keep the actual file that drifted for diffing against a known-good.
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: sim-tme-3d-summary-drift
+          path: simulations/output/tme-3d/summary.json
+          if-no-files-found: warn
--- a/simulations/sim-tme-3d/README.md
+++ b/simulations/sim-tme-3d/README.md
@@ -164,6 +164,15 @@ Three smoke tests:
 
 The library primitives (`physics`, `oxygen`, `ph`, `stromal`, `immune_spatial`) are exhaustively tested in `ferroptosis-core`'s 160+ unit tests. This binary tests orchestration, not the math.
 
+### Production byte-identity regression (#253)
+
+Two layers guard the load-bearing invariant that the #239 multi-dose work relies on (default matrix = all `DoseSchedule::Constant` = byte-identical `summary.json`):
+
+1. **Per-PR (fast):** `constant_path_golden_kill_counts` pins integer kill counts at a small 20³ × 80 config, so a structural regression fails in ordinary `cargo test`.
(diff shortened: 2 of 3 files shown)


ELares:
## Code review: full-production byte-identity regression for sim-tme-3d (#256)

Reviewed the diff against the surrounding source (`sim-tme-3d/src/main.rs`, `rust-toolchain.toml`) and the repo's existing workflows. This is a clean, well reasoned PR and I found no correctness bugs. The byte-identity guard rests on a genuinely deterministic production path, the CI plumbing is wired correctly, and the new workflow follows the conventions already established in `cargo-test.yml`. Everything below is optional hardening or operational context, not a blocker.

### What it does

Adds an off-PR CI job (weekly cron plus manual `workflow_dispatch`) that runs the full 60³ × 180 default 24-condition matrix and asserts `output/tme-3d/summary.json`'s SHA-256 against a checked-in expected hash. It's the production-scale companion to the fast `constant_path_golden_kill_counts` unit golden. Plus a README section documenting the two-layer guard and housekeeping on the now-closed #193/#194 references.

### Verified correct

* **Path resolution is consistent end to end.** Both `cargo run` and `sha256sum -c` use `working-directory: simulations`; `main.rs` writes `output/tme-3d/summary.json` relative to CWD, and the checklist path resolves to that same `simulations/output/tme-3d/summary.json`. No stale-output false-pass either: `output/` is gitignored, checkout starts clean, `rust-cache` caches `target/` and the registry (not `output/`), and `fs::write` truncates and rewrites each run.
* **The asserted output is genuinely deterministic.** On the default matrix path there's no HashMap iteration order, no wall clock in the summary path (the only `Instant::now()` lives in `run_bench`), per-cell seeded RNG (so no cross-thread draw-order dependence under rayon), and integer `.sum()` reductions that are order independent. Crucially, the two float aggregates (`peak_damp` via `fold max`, `total_damp` via `.iter().sum()`) are serial rather than `par_iter`, so they're immune to thread-count-dependent reassociation, which is the one realistic way a float output could go flaky. There's even a `rayon_run_is_deterministic` test. So pinning the toolchain (the one remaining drift vector, ULP differences in Ziggurat / `powf` across libm) is precisely the right and sufficient lever, and the expected SHA should be stable.
* **Scope and security are tight.** `permissions: contents: read`, triggers are `workflow_dispatch` plus `schedule` only (no `pull_request_target`, no untrusted input, no secrets). Asserting only `summary.json` is sufficient: the `--snapshot` / `--dose-sweep` / `--bench` branches each early-return and are the sole writers of the other output files, so the flagless default invocation writes only `summary.json`.
* **Matches repo convention.** `checkout@v6`, `rust-cache@v2`, and the pinned `dtolnay/rust-toolchain@1.96.0` (with the #236 rationale) mirror the existing `fmt` job in `cargo-test.yml`, and the pin matches `rust-toolchain.toml`. One nuance worth recording: because the toml also pins 1.96.0 and `cargo` runs from `simulations/`, the toml is what actually selects the toolchain at run time (rust-toolchain.toml outranks the default that the `dtolnay` action sets). The `dtolnay` step's real job is to pre-install that exact channel so there's no run-time auto-download. They should still move in lockstep as the comment says, but if they ever diverge the toml wins, so byte-identity is preserved regardless.

### Worth knowing before relying on it

* **This workflow gets zero CI exercise during PR review.** GitHub only registers `schedule` and `workflow_dispatch` from the copy on the default branch, so neither the cron nor the manual Run button will fire (or even appear in the Actions UI) until this merges to `main`. The PR's verification is all local, which is fine and well documented, but the first real execution is post-merge and the first scheduled run is the following Monday 09:00 UTC. I'd manually dispatch it immediately after merge to confirm it's green, rather than discovering a YAML or path typo on a future Monday.

### Optional hardening (non-blocking)

* **Add `timeout-minutes` to the job.** With none set it inherits GitHub's 6-hour default. This job does a cold `--release` build plus the heavy 60³ × 180 run, so a wedge could squat a runner for hours. Something like `timeout-minutes: 60` fails a hang fast.
* **Upload the actual `summary.json` on failure.** Today a drift prints the computed SHA but discards the file with the runner, so triage means reproducing the whole production run locally. An `actions/upload-artifact` step with `if: failure()` capturing `output/tme-3d/summary.json` makes "what drifted" inspectable straight from the run. The README positions the printed SHA as drift-triage material, but a bare hash without the file is weak triage. For a guard whose whole point is catching a subtle scale-dependent perturbation, the divergent artifact is the most useful thing to keep.
* **Stagger the cron off the top of the hour.** `0 9 * * 1` lands in GitHub's documented high-load window (minute 0), where scheduled runs can be delayed or silently dropped. Something like `17 9 * * 1` reduces queueing and the chance of a skipped Monday, which matters more at weekly cadence.

### Minor notes

* The `1.96.0` literal now lives in three places (`rust-toolchain.toml`, the `fmt` job, and this new job), so a channel bump is a three-file change. Both pinned-job comments already flag the coupling, so this is just for awareness.
* The `rust-cache` step buys less here than on a per-PR job: it caches `~/.cargo` and dependency artifacts, not the workspace crates, so the dominant cost (the cold workspace compile plus the sim run) isn't cached regardless, and the weekly cadence also sits right at the 7-day cache-eviction boundary. Harmless to keep, just don't expect much of a hit.
* Since this is a public repo, the `schedule` trigger is also subject to GitHub's auto-disable after 60 days of repository inactivity. Low risk while the repo is active, and a manual dispatch counts as activity to re-arm it.
* No `concurrency` group, so an ad-hoc dispatch overlapping the cron would run in parallel. Harmless here (read only, each run regenerates its own output), noted only for completeness.

Nice work overall. The determinism reasoning behind the guard is solid and the two-layer design is easy to follow from the docs.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

PR #263, opened 2026-06-01
Immune realism Phase 1: T-cell exhaustion (#243)

### LINKED ISSUE #243: Immune realism beyond DAMP→DC cascade: exhaustion, checkpoints, suppressor cells
## Context

The current spatial immune model (`immune_spatial` + `immune.rs`, landed via #188) is intentionally narrow: a 0–48 h resident T-cell cascade with a simplified PD-1 brake. It models ferroptosis → DAMP release → DAMP diffusion → DC activation (Michaelis-Menten) → T-cell kill with `effective_brake = pd1_brake × (1 - anti_pd1_efficacy)`. This is sufficient for the manuscript's "is the 104:1 SDT/RSL3 immune-kill ratio robust in 3D?" question, but it omits most of real tumor immunology.

This is one of six realism gaps identified during the #193 visualization PR.

## Current state

- `immune_spatial::immune_kill_probability(activation, kill_rate, effective_brake) → f64` — single PD-1-only checkpoint, no exhaustion, no Treg / MDSC.
- T-cell kill rate is constant over time (no exhaustion-driven decay after sustained kill activity).
- No regulatory T cells (Tregs) or myeloid-derived suppressor cells (MDSCs) actively suppressing the immune response.
- No DC subset distinction (cDC1 vs cDC2 have different DAMP responsiveness).
- No spatial structure to immune infiltration (currently uniform availability; reality is patchy infiltration shaped by vasculature & TLS).

Phased upgrade — order chosen by impact-per-effort:

### Phase 1 — T-cell exhaustion (high ROI, low complexity)

Cumulative immune kills reduce local kill probability via `kill_prob = base × 1 / (1 + α · cumulative_kills_in_neighborhood)`. Models "T-cell exhaustion" effect that's well-documented in checkpoint-blockade literature.

### Phase 2 — Treg / MDSC suppressor field

Add a second diffusing field (`suppressor_field`) that scales down kill probability locally. Initialized with patches at known Treg-rich locations (perivascular niches if vasculature lands via #191). Decays over time and replenishes; opposes the kill effect.

### Phase 3 — Multi-checkpoint axis (for combination immunotherapy modeling)

Replace the single PD-1 axis with a vector of checkpoint markers (PD-1, CTLA-4, LAG-3, TIM-3), each independently modulated by drug intervention. Enables modeling anti-PD-1 + anti-CTLA-4 combinations.

### Phase 4 — DC subset complexity

cDC1 (cross-presenting, anti-tumor) vs cDC2 (Th17-skewing, less effective). Different DAMP responsiveness. Requires literature parameterization.

## Suggested approach

Land Phase 1 first as a standalone PR (likely 200–400 LOC across `immune_spatial` + sim-tme + sim-tme-3d). Phase 2 builds on top.

For Phase 1:
1. Add `cumulative_kills_field: Vec<u32>` to the immune-step state (counter per cell, incremented when an immune kill occurs in its Moore neighborhood).
2. Modify `immune_kill_probability` signature to accept exhaustion state; scale by `1 / (1 + α · cumulative)`.
3. Add `exhaustion_rate` parameter to `SpatialImmuneConfig`.
4. Bit-identical output when `exhaustion_rate = 0.0` (default).

## Acceptance criteria

**Phase 1:**
- [ ] T-cell exhaustion parameter in `SpatialImmuneConfig`
- [ ] Cumulative-kill field tracked per simulation
- [ ] `immune_kill_probability` scales by exhaustion factor
- [ ] Bit-identical output when exhaustion disabled
- [ ] Validation: declining immune kill rate after ~30% tumor mass kill (literature: "cold tumor" emergence)
- [ ] Documented with ≥2 references (e.g., Wherry et al. on T-cell exhaustion, Snell et al. on dysfunction)

**Phase 2 (separate PR):**
- [ ] Suppressor field with diffusion + clearance (mirror DAMP field pattern)
- [ ] Vasculature-linked Treg initialization (if #191 has landed)
- [ ] Validation: anti-PD-1 alone less effective than anti-PD-1 + Treg depletion in same simulation

## Out of scope

- Adaptive immune evolution (cancer vaccine, CAR-T) — separate.
- Innate immune complexity (NK cells, macrophage polarization M1/M2) — separate.
- Lymph-node-mediated systemic immunity (current model is resident T-cell phase only — manuscript caveat).

## Related

- #188 — immune coupling (the existing baseline being extended)
- #191 — vasculature (Phase 2 needs vessel positions for Treg-niche initialization)
- #197 — 3D ferroptosis biochemistry (immune-evasive persister cells)
- #241 — persister state (cells that survive ferroptosis may also be immune-evasive)

## References

- Wherry NJ. Nat Immunol 2011 (T-cell exhaustion canonical)
- Snell et al. Cell 2018 (CD8 T-cell exhaustion in cancer)
- Tauriello et al. Nature 2018 (TGFβ–Treg axis in colorectal cancer)
- Sharma & Allison Cell 2015 (immune checkpoint biology in cancer)

## Source

Identified during PR #238 discussion of realism gaps. 5 of 5 issues filed from that audit.

ELares:
Phase 1 of #243: adds **T-cell exhaustion** on top of the #188 spatial immune cascade. Sustained killing in a region drives resident cytotoxic T cells toward dysfunction, lowering their per-encounter kill probability (Wherry, Nat Immunol 2011; Snell et al., Cell 2018). Later phases (Treg/MDSC suppressor field, multi-checkpoint axis, DC subsets) stay separate per the issue.

## ferroptosis-core (v0.14.0)
- **`immune_spatial::exhaustion_factor(cumulative_kills, rate)`** = `1/(1 + rate·cumulative)`; pure, returns exactly `1.0` at `rate == 0` (identity no-op).
- **`SpatialImmuneConfig::exhaustion_rate`** (`0.0` default in `for_2d`/`for_3d`).
- **`TumorGrid3D::coords`** (inverse of `flat_index`) for the neighborhood scatter.

## sim-tme-3d
- A per-cell `cumulative_kills` field counts immune kills in each cell's Moore-26 neighborhood and scales its kill probability by `exhaustion_factor`. The field is allocated **only** when `exhaustion_rate > 0`, and the scatter runs only then, so the default matrix path is **byte-identical** (production SHA `1cf8866312…` unchanged; golden tests pass). Threaded via a new `Option<SpatialImmuneConfig>` override arg (matrix/dose-sweep/tests pass `None`).
- The immune-kill loop now `filter_map`-collects killed indices to drive the serial neighborhood scatter; the kill *decisions* are unchanged (`×1.0` on the off path), so the set of dead cells is identical.

## Acceptance criteria (Phase 1)
- [x] `exhaustion_rate` parameter in `SpatialImmuneConfig`
- [x] Cumulative-kill field tracked per simulation
- [x] `immune_kill_probability` scaled by an exhaustion factor (via `exhaustion_factor`, consumer-composed like `oxygen`/`ph`/`stromal`)
- [x] **Bit-identical when disabled** — SHA unchanged, `exhaustion_factor(_, 0.0) == 1.0`
- [x] **Validation: declining immune kills** — `exhaustion_reduces_immune_kills` drops a dense SDT+immune run ≈20% (174 → 139) with exhaustion on, all else equal (>10% asserted)
- [x] Documented with refs (Wherry 2011, Snell 2018)

## Tests
`exhaustion_reduces_immune_kills` (integration) + 5 `ferroptosis-core` unit tests (`exhaustion_factor` identity/monotonicity, `coords` roundtrip). fmt clean; no new clippy warnings.

## Honesty note
Exhaustion shifts a few deaths from immune to ferroptosis (a cell spared an apoptotic immune kill can die ferroptotically and release iron, which couples to neighbors), so the test asserts on the **immune-kill count**, not a fixed ferroptosis tally. `exhaustion_rate` is an uncalibrated placeholder; the demonstration uses a deliberately dense-killing regime (boosted `immune_kill_rate`, identical across both runs) since sparse killing legitimately doesn't trigger exhaustion.

Closes #243


ELares:
## Review: T cell exhaustion (Phase 1, #243)

Reviewed the full diff (9 files, +258/-30) and CI (fmt, test, CodeQL, all Analyze jobs green). This is a clean, carefully guarded addition. Notes below.

### What it does
- New pure helper `immune_spatial::exhaustion_factor(cumulative_kills, rate) = 1/(1 + rate·cumulative)`, identity at rate 0 or cumulative 0.
- New `SpatialImmuneConfig::exhaustion_rate` (`0.0` default in both `for_2d`/`for_3d`).
- New `TumorGrid3D::coords` (inverse of `flat_index`).
- sim-tme-3d threads an `Option<SpatialImmuneConfig>` override; a per cell `cumulative_kills` neighborhood counter scales each cell's immune kill probability. The field is allocated and the scatter runs only when `exhaustion_rate > 0`.

### Correctness (looks solid)
- `coords` is a correct inverse of `flat_index` for the row major layout, and the non cubic roundtrip test (7×5×9) properly guards against axis order bugs.
- Byte identity on the default path holds at the IEEE 754 level: `kill_prob * 1.0` is value preserving for every finite input, the RNG seeding and comparison are untouched, and `cumulative_kills` is an empty `Vec` that is never indexed when disabled. The new `filter_map().len()` count equals the old `map().sum()` count exactly.
- The race free design is the nice part. Every read of `cumulative_kills[idx]` happens inside the `par_iter_mut`; every write happens in the serial scatter after the join. So a step's kills only affect future steps, and the parallel section has no read after write hazard. Integer increments commute, so the scatter is order independent regardless of rayon's `collect` order.
- No overflow risk: a cell is killed at most once before being marked dead, so any neighborhood counter is bounded by the 26 Moore neighbors. `u32` is comfortably safe.

### One thing worth a look (performance, default path)
The immune loop switched from `map().sum()` to `filter_map().collect::<Vec<usize>>()`. On the default matrix path (`exhaustion_on == false`) this now allocates a `Vec<usize>` of killed indices on every immune step and then only reads `.len()` from it. The old `.sum()` was allocation free.

Output stays byte identical, but this is a per step heap allocation on the hot path that the default matrix never benefits from. Given the #192 / #253 performance and SHA discipline, consider either:
- branch it: `if exhaustion_on { collect } else { sum }` (a little closure duplication), or
- confirm via `--bench` that the regression is in the noise and note it in the PR.

Not a blocker, just flagging since this path is otherwise kept deliberately zero cost.

### Minor notes
- `exhaustion_rate` lives on the shared `SpatialImmuneConfig` and is defaulted in `for_2d`, but only the 3D consumer reads it. A 2D user who sets `exhaustion_rate > 0` would silently get no effect. A one line "3D only in Phase 1" on the field doc would prevent that footgun.
- Adding a public field to `SpatialImmuneConfig` is a source breaking change for external code that builds it with a struct literal and no `..`. Fine under 0.x and consistent with how `anti_pd1_efficacy` was added, but if the library is meant for external embedding, `#[non_exhaustive]` on these config structs would future proof them.
- Optional: a `debug_assert!(exhaustion_rate >= 0.0)` in `exhaustion_factor` would catch a negative placeholder, which would otherwise push the factor above 1 or negative.

### Tests
`exhaustion_reduces_immune_kills` is a fair A/B: both runs share the boosted `immune_kill_rate`, differing only in `exhaustion_rate`, and it asserts on a relative reduction (>10%) with a `>= 50` baseline precondition rather than the exact 174 to 139 counts, so it is robust to model drift. The honesty note about a few deaths shifting from the immune tally into ferroptosis (iron release coupling) is appreciated, and asserting on the immune kill count only is the right call. The `exhaustion_factor` unit tests cover identity, monotonicity in both args, and the exact hyperbolic value.

Nice work, the model and the guarding are both sound. The only actionable item is the default path allocation question above.

PR #271, opened 2026-06-01
feat(sim-tme-3d): patient-scale slab geometry (#240)

### LINKED ISSUE #240: Patient-scale tumor (≥5 mm): bridge in-vitro spheroid sim to clinical scale

ELares:
Closes #240.

## Why

The 3D spheroid is ~540 µm radius, in-vitro scale, where drug/O2 penetration (Krogh ~150 µm) reaches most cells. Real patient tumors are 5–50 mm, where penetration collapses with depth and the bulk of the tumor is essentially drug-deprived. This adds an opt-in **slab** mode that models a chunk of such a tumor at a configurable depth, so the efficacy numbers reflect the patient-scale penetration limit rather than the optimistic spheroid scale. This is the other end of the scale gap the README already flags.

## What

**ferroptosis-core (v0.17.0 → v0.18.0, new `slab` module):**
- `TumorGrid3D::generate_slab` builds an all-tumor block (no sphere carve, no stroma): bulk tumor with the usual 80/15/5 Glycolytic/OXPHOS/Persister phenotype mix.
- `slab_supply_field` gives each cell a planar depth-graded supply `exp(-depth/λ)`: the **+z face** (top layer) is vessel-proximal at `depth_offset`, supply decays going **−z** (deeper). The 1-D analog of `oxygen::radial_o2_field`.
- `SlabConfig::patient_5mm()` (+z face at 4 mm, ~fully deprived) and `surface()` (the shallow vessel-fed control); `scale_interpretation` (a human-readable depth-span string for the output JSON); `KROGH_LAMBDA_UM` (~150 µm default).

**sim-tme-3d:**
- Unify the slab field and the #191 vessel field into a single per-cell `supply_field` (slab takes precedence; they're mutually-exclusive O2-source overrides). It **replaces** the edge-distance O2 proxy (× `basal_ros`) AND scales drug delivery on every path (exo dose at init, RSL3 constant-knockdown correction, dosed `rsl3_drug_avail`).
- `ConditionResult.scale_interpretation` (`#[serde(skip_serializing_if)]`, omitted on non-slab runs) so the output says plainly what physical scale it represents.
- `--snapshot=slab` preset (SDT + immune on a **surface** slab): the depth-graded death front renders directly in the dead/DAMP/LP z-slice — the renderer's mid-slice spans `(col, layer)`, so the depth axis is shown. No new static overlay, no renderer change.

## Boundary conditions (the "reflective BC" part of the scope)

+z = vessel (supply maximum, immune-accessible via the existing DAMP cascade). −z = continuing tumor, modeled as a **no-flux / reflective** face — already satisfied because iron/DAMP diffusion uses bounded (no-wrap) neighbors (no flux across any face), and the planar supply replaces the O2 field rather than diffusing it. Documented in the `slab` module header + the sim README.

## Byte-identity

Off by default: the matrix passes no slab config, so it runs the spheroid grid + edge-distance/vessel O2 unchanged. The full 60³×180 default-matrix `summary.json` SHA-256 is **unchanged** (`1cf88663…390ff71`); the `constant_path`/`dosed_path` golden tests + the #253 production-SHA workflow hold.

## Tests

- 4 `slab` unit tests (all-tumor grid, supply decreases with depth, deep slab ~deprived, scale-interpretation string).
- `patient_scale_slab_kills_far_less_than_spheroid`: a 4 mm-deep slab kills **<20%** of what the in-vitro spheroid does under the same SDT — the penetration collapse the spheroid scale misses.
- `shallow_slab_outkills_deep_slab`: confirms the depth-dependence within slab mode.
- Full workspace green (`cargo test --workspace`); `cargo fmt --all --check` clean.

## Docs

ferroptosis-core README + lib.rs module table (`slab` row), sim-tme-3d README (slab section + snapshot table + scale-mismatch cross-link), CLAUDE.md (workstream, version ladder to v0.18.0, module count 20→21, TME line).


ELares:
## Code review — #240 patient-scale slab geometry

Reviewed the full diff (9 files, +474/−38). Solid PR: it reuses the established `Overrides` opt-in + unified-`supply_field` pattern from #191/#197 faithfully, keeps the default matrix byte-identical (verified locally — production `summary.json` SHA still `1cf88663…390ff71`), and the docs/version ladder are thorough. Tests cover the headline claim and the depth-dependence. Below are the things worth fixing or at least acknowledging before merge — none are blockers, but a few are real.

### 🔴 Substantive

**1. `SlabConfig::patient_5mm()` produces a slab that exceeds its own `virtual_tumor_mm`.**
At production scale the grid is 60 layers × `CELL_SIZE_UM` (20 µm) = **1.2 mm** thick. `patient_5mm()` sets `depth_offset_mm = 4.0`, so the slab spans **4.0–5.2 mm** inside a `virtual_tumor_mm = 5.0`. The `scale_interpretation` string — which is the user-facing AC artifact — then literally reads *"slab spanning depth 4.0–5.2 mm of a 5 mm virtual tumor (1.2 mm thick)"*, i.e. the slab pokes 0.2 mm out the bottom of the tumor it claims to live in. `virtual_tumor_mm` is cosmetic (it constrains nothing), so there's no functional bug, but the output contradicts itself. Suggest bumping `virtual_tumor_mm` to ≥ 5.5 (or 6.0), or framing the default as a ">5 mm tumor". A `debug_assert!(depth_offset_mm + slab_mm <= virtual_tumor_mm)` in `scale_interpretation` would catch this class of inconsistency.

**2. The depth-collapse magnitude is uncalibrated and should say so (repo honesty principle).**
The headline "a 4 mm slab kills <20% of the spheroid" (in practice ~0 kills) comes from a first-order Krogh `exp(-depth/λ)` with a placeholder λ. That's a reasonable v1, but the `## Patient-scale slab` README section states the penetration limit fairly confidently without the caveat the `spheroid` section gives its own gradients ("placeholders pending calibration"). Per CLAUDE.md ("be honest about what we don't know", "language that reflects uncertainty"), add one line noting the depth-collapse *magnitude* is an uncalibrated approximation — illustrative of the scale gap, not a validated efficacy number.

### 🟡 Minor

**3. Stale example in the `ConditionResult.scale_interpretation` doc comment.** It says *"e.g. 'slab spanning depth 4.0–4.4 mm…'"* — but 4.4 is the 0.4 mm **test-grid** span (20 layers); production is 4.0–**5.2** mm, which is what `slab.rs` and the READMEs correctly show. Align the doc example to avoid implying a 0.4 mm production slab.

**4. `slab` + `spheroid` in `Overrides` is undefined and unguarded.** If a caller sets both, you get a `generate_slab` all-tumor cube run under `Params::spheroid()` with `apply_radial_cells_3d` re-assigning phenotypes by *radial fraction* — but a uniform cube has no meaningful spheroid center, so the radial structure is a geometric artifact, while `supply_field` silently takes the slab branch. No path triggers this today (no preset sets both; the matrix sets neither), and the slab-vs-vessel precedence *is* documented — but the slab-vs-spheroid grid/params combination isn't. A `debug_assert!(!(slab_cfg.is_some() && spheroid_cfg.is_some()))` or a one-line precedence note would close the footgun.

**5. Slab drug-scaling is covered only transitively.** The integration tests use SDT, where both O2 (`basal_ros`) and drug (`exo_ros`) collapse together in the deep slab, so they can't isolate that the slab field also scales *drug* (RSL3 `avail`/knockdown). That path is the same unified `supply_field` code #191 already tests with a dedicated RSL3 case, so it's covered by reuse — but there's no slab-specific RSL3 assertion. Worth one if you want the slab drug-coupling pinned independently of the vessel path.

**6. The reflective −z BC is argued, not tested.** "Full incl. reflective BC" was the scope; the BC holds because iron/DAMP diffusion uses bounded no-wrap neighbors (correct), but nothing tests that the −z face (`l=0`) doesn't wrap/leak to `l=layers-1`. Optional, but a no-wraparound assertion would make the "reflective boundary" claim self-evidencing rather than a comment.

### Nits
- `--snapshot=slab` runs at λ=120 (`ZONE_REF_LAMBDA`, set by the preset's `o2_lambda`), not the `KROGH_LAMBDA_UM = 150` default — fine, but the in-code "else the Krogh default" comment can read as if 150 is used here. Both λ values drive the bottom layer to ~0 supply, so the viz is unaffected.
- `scale_interpretation` exists as both an imported fn and a struct field; no shadowing issue (struct-init syntax), just mildly confusing to read.

### Verified
Byte-identity (SHA unchanged), `cargo fmt --check` clean, full `cargo test --workspace` green incl. the 4 slab unit tests + 2 integration tests, `generate_slab`'s `CellState` init matches `generate`'s field-for-field (incl. `persister_fraction`), and CI (fmt + test) is green on the PR.

Recommend addressing #1 and #2 (both small) before merge; #3–#6 are optional polish.

Issue #266, opened 2026-06-01
Clonal model refinements: durable GPX4 axis, calibration, clonal evolution (follow-up to #242)


ELares:
## Context

#242 landed clonal heterogeneity (`ferroptosis-core::clonal` + the `Overrides`-wired sim-tme-3d path). Three items were deferred in code/docs or in #242's out-of-scope list; tracking them here so they aren't lost (companion to #262 persister-refinements / #264 immune Phases 2-4).

### 1. Make the GPX4 subclone axis durable (flagged in #265 review)
`gpx4_mul` scales the *initial* `state.gpx4`, which strongly shapes the early autocatalytic window but then relaxes toward the NRF2 setpoint (`cell.nrf2 * gpx4_nrf2_target_multiplier`, ~0.008/step) over the 180-step run — so a "GPX4-low" subclone identity is transient. Only `iron_mul` and `lipid_unsat_mul` are fully durable (they scale static `Cell` fields). A durable GPX4 axis would also scale the static `cell.nrf2` setpoint. Caveat: `cell.nrf2` also drives GSH resynthesis, so scaling it broadens the perturbation to general antioxidant capacity (arguably biologically coherent for a subclone, but worth a deliberate decision). Add a differential test that asserts the GPX4 axis persists late in the run.

### 2. Calibrate the subclone perturbation table
`ClonalConfig::literature_4()` values (iron/GPX4/lipid_unsat multipliers) are placeholders chosen for qualitative direction, not fit to data. Add them to the calibration infrastructure (`simulations/calibration/`) with sources/targets (e.g. drug-screen kill-rate spreads across mesenchymal vs epithelial lines; Conrad 2018, Viswanathan 2017). Overlaps with the broader realism-parameter calibration theme (#262).

### 3. Clonal evolution + inter-tumor heterogeneity (#242 out-of-scope)
- **Clonal evolution / dynamic selection**: subclones with higher fitness expand over simulation steps (currently the subclone map is static). Larger modeling effort.
- **Inter-tumor heterogeneity**: patient-to-patient variation in subclone composition — a population-level / ensemble model, separate.

## Discipline (carry forward)
- Keep off-by-default identity configs byte-identical (the #253 production-SHA guard + golden + K=1-identity tests must keep passing).
- Pure helpers in `ferroptosis-core`, consumer applies (the `oxygen`/`ph`/`clonal` pattern).
- A differential test per axis/mechanism.

## Related
[...]

PR #265, opened 2026-06-01
Clonal heterogeneity: spatial subclones with distinct ferroptosis profiles (#242)

ELares:
Models **clonal heterogeneity** (#242): K spatial subclones with distinct ferroptosis vulnerabilities. Real tumors are genetic mosaics, and the between-subclone vulnerability variance often exceeds the between-treatment variance in drug screens (Marusyk 2014; Conrad 2018; Heindl 2019).

## ferroptosis-core (v0.15.0)
- New **`clonal`** module:
  - `assign_subclones_3d(grid, k, seed)` — Voronoi subclone map from K seed points placed by an **independent** `StdRng`, so `TumorGrid3D::generate`'s stream (and the cell grid) is **untouched**. Returns a consumer-owned `Vec<u8>` (`0` = stroma, `1..=k` = subclone).
  - `ClonalConfig` / `SubclonePerturbation` — per-subclone iron / GPX4 / MUFA perturbations the consumer applies as RNG-neutral setup mutations. `single_identity()` (K=1) is a no-op; `literature_4()` spans the mesenchymal⇄epithelial axis.
- New `TumorGrid3D::coords` was added in #243; reused here for the neighborhood-free index→coord mapping.

## sim-tme-3d
- **Refactor:** the three optional realism layers (persister #241, immune #243, clonal #242) are bundled into an `Overrides` struct (`Default` = all-`None` = the byte-identical matrix path), replacing the growing positional-argument list on `run_one_condition_full`. All existing callers/tests migrated.
- Per-subclone perturbations applied as RNG-neutral setup mutations (like the O2/pH gradients); assignment uses a dedicated `SUBCLONE_SEED`. **Default matrix byte-identical** (production SHA `1cf8866312…` unchanged; golden + K=1-identity tests pass).
- `summary.json` gains an optional `subclone_kills` array (`subclone_id` / `total_tumor` / `total_dead` / `kill_rate`), omitted unless clonal is enabled (`skip_serializing_if`).
- `--snapshot=clonal` writes a static `subclone.npy` + the renderer adds a discrete-colored subclone-id panel (rendered end-to-end: 4 subclones, 4-panel GIF).

## Acceptance criteria
- [x] Deterministic assignment given seed
- [x] ≥4 subclones supported; default `literature_4()` is K=4
- [x] **Bit-identical when K=1** (`clonal_k1_identity_is_byte_identical`; production SHA unchanged)
- [x] Per-subclone kill rate in `summary.json` (`subclone_kills`)
[...]

PR #273, opened 2026-06-01
feat(clonal): durable GPX4 subclone axis (#266 item 1)

ELares:
Addresses item 1 of #266 (clonal model refinements), the durable-GPX4 axis flagged in the #265 review.

## Problem

`SubclonePerturbation::gpx4_mul` scaled only the *initial* `state.gpx4`. But `state.gpx4` relaxes toward the shared NRF2 setpoint each step (`gpx4_target = cell.nrf2 · gpx4_nrf2_target_multiplier`, upregulation ~0.008/step), so over a 180-step run a "GPX4-low" subclone's reserve climbs back toward the same setpoint as everyone else — the resistance/vulnerability identity was an early-window transient. Only `iron_mul` and `lipid_unsat_mul` (which scale static `Cell` fields) were durable.

## Fix (design per #266 — fold into `gpx4_mul`, no new knob)

The consumer now scales the static `cell.nrf2` **setpoint** in addition to the initial `state.gpx4` reserve:

```rust
gc.state.gpx4 *= p.gpx4_mul; // initial reserve
gc.cell.nrf2  *= p.gpx4_mul; // durable setpoint (#266) — also scales GSH resynthesis
```

So a GPX4-low subclone relaxes toward a *lower* setpoint and stays differentiated for the whole run. **Deliberate trade-off:** `cell.nrf2` is the master antioxidant regulator, so it also drives GSH resynthesis — this axis is therefore "general antioxidant capacity" (an NRF2-low subclone), not GPX4-reserve in isolation. That's biologically coherent for a subclone identity (chosen over a separate `nrf2_mul` knob to avoid a 4th uncalibrated parameter).

## Tests

`gpx4_axis_persists_late_only_when_nrf2_is_scaled` (ferroptosis-core) runs three variants of one baseline cell to step 160 under Control:
- **identity** (no knock), **transient** (old: initial `state.gpx4` only), **durable** (#266: initial + `cell.nrf2`).

It asserts the durable knock holds a meaningfully lower late-run `gpx4` (< 0.75× identity), the transient knock relaxes back *above* the durable one, and the durable late-run differentiation from identity is > 2× the transient one. The existing `clonal_subclones_differ_in_kill_rate` still passes (now stronger).

## Byte-identity

The `nrf2` scaling lives inside the clonal-gated perturbation block; the default matrix passes no clonal config and `single_identity()` (K=1, `gpx4_mul = 1.0`) is a no-op. Default-matrix `summary.json` SHA-256 unchanged (`1cf88663…390ff71`); full workspace green (223 core tests, +1); `cargo fmt --check` clean. No `ferroptosis-core` API change (struct unchanged; docs clarified), so no version bump.

Items 2 (calibration) + 3 (clonal evolution) of #266 remain open.


ELares:
## Code review — #266 item 1: durable GPX4 subclone axis

Reviewed the full diff (5 files, +105/−14). Tight, well-scoped change that does exactly what the #265 review flagged: the GPX4/antioxidant subclone identity now persists instead of relaxing back to the shared setpoint. Mechanism is correct, byte-identity holds, and the differential test is the right shape. A couple of honest caveats below; nothing blocking.

### ✅ Correct + verified
- **Mechanism is right.** `gpx4_target = cell.nrf2 · gpx4_nrf2_target_multiplier` (biochem.rs:245), and `state.gpx4` relaxes toward it at ~0.008/step. Scaling `cell.nrf2` moves the target, so a low-antioxidant subclone converges to a *lower* setpoint — durable. Keeping the initial `state.gpx4 *= gpx4_mul` too means the cell both starts low and stays low. Coherent.
- **The "general antioxidant capacity" framing is complete.** I grepped every use of `cell.nrf2` in the cascade: it's exactly two sites — the GPX4 setpoint (biochem.rs:245/349) and GSH resynthesis (`state.gsh += cell.nrf2 · nrf2_gsh_rate · deficit_fraction`, :217/326). No hidden third coupling, so the doc's claim that this axis = {GPX4 setpoint + GSH} is accurate, not an under-count.
- **Byte-identity preserved.** The `nrf2` scaling is inside the `(subclone_ids, clonal_cfg)` gate; the default matrix passes no clonal config and `single_identity()` (gpx4_mul=1.0) is a no-op. Confirmed the production `summary.json` SHA is unchanged.
- **Differential test is the right design.** Three variants of one cloned baseline cell, same-seed isolation, asserting the durable knock's late-run gap from identity is >2× the transient one's. Directly pins "persists late" rather than asserting an absolute number.

### 🟡 Worth addressing / acknowledging

1. **The durable change shifts the *effective* magnitude at the resistant end — and `literature_4` wasn't re-tuned (honesty).** `gpx4_mul` now compounds: for the resistant subclones (`gpx4_mul` 1.1/1.3) it raises **both** the GPX4 setpoint **and** GSH resynthesis, multiplicatively, on top of phenotypes that can already start NRF2-high (`PersisterNrf2` base nrf2 ≈ 3.0 → ×1.3 ≈ 3.9). Before this PR the GPX4 knock washed out mid-run; now it sticks AND drags GSH with it, so subclone 4 is meaningfully more resistant than the same table produced yesterday. The `literature_4` multipliers were chosen for the old transient behavior and are explicitly placeholders — so the durable axis has quietly changed what they mean. This is exactly the "effective magnitude moved, values uncalibrated" situation the repo wants surfaced: recommend a one-line note in `literature_4`'s doc (or the #266 calibration item) that the multipliers need re-checking now that the axis is durable + GSH-coupled, so a future reader doesn't treat the current spread as intentional.

2. **The consumer line itself has no test that would fail if it were dropped.** The core mechanism test is excellent, but it exercises `sim_cell_step` directly — not the `gc.cell.nrf2 *= p.gpx4_mul` line in main.rs. `clonal_subclones_differ_in_kill_rate` runs that path but passed *before* this fix too (it only checks vulnerable > resistant, which the transient axis already satisfied early). So nothing in the sim crate would catch accidentally deleting the new line. Optional but cheap: a sim-tme-3d test asserting the resistant subclone's **late-run** kill suppression specifically depends on the durable axis (e.g. resistant kill-rate stays low at full run length, where the old transient axis would have let it climb).

### Nit
- The test baseline is `Phenotype::OXPHOS` (nrf2 ≈ 1.2, gpx4 ≈ 1.0) — a reasonable near-setpoint choice, so the differential is clean. Fine as-is; just noting it's the one phenotype-specific input the test depends on.

### Verdict
Correct, clean, byte-identical, well-tested at the mechanism level. Finding #1 (a one-line calibration caveat) is the only thing I'd genuinely want before this is "done honestly" — and it can fold into #266's existing calibration item rather than block this PR. #2 is optional polish.

Issue #270, opened 2026-06-01
Spheroid radial-biochem refinements: per-cell MUFA durability, zone calibration, contact/nutrient gradients (follow-up to #197)


ELares:
## Context

#197 landed 3D spheroid radial biochemistry (`ferroptosis-core::spheroid` + `Params::spheroid()`, wired into sim-tme-3d behind `Overrides.spheroid`). Three items were deferred / flagged in the #269 review; tracking them here (companion to #262 / #264 / #266 / #268 realism follow-ups).

### 1. Durable position-dependent MUFA (per-cell M_ss)
The radial MUFA value is set on `CellState` but is an **initial condition**: each step `update_mufa_protection` relaxes it toward `Params::spheroid`'s **uniform** M_ss (≈0.20) and clamps it at the global `scd_mufa_max` (0.25). So the peripheral-vs-core MUFA spread shapes the early killing window but converges over the run. A fully durable position-dependent MUFA needs a **per-cell** `scd_mufa_max` / M_ss (Params is currently global) — e.g. a per-cell MUFA-cap field that `update_mufa_protection` reads. (Only `iron` — a static `cell.iron` scale — is a fully durable radial axis today; `gsh` is likewise an initial condition that evolves under NRF2 resynthesis.)

### 2. Zone-geometry calibration
`SpheroidConfig` thresholds are **radial** fractions, so by volume (∝ r³) the persister core (`frac < 0.33`) is only ~4% and the glycolytic rim (`frac ≥ 0.66`) ~71% — roughly inverted from real spheroids (thin proliferating rim, large quiescent/hypoxic core). Calibrate the thresholds (likely raise `oxphos_frac` for a larger core) against spheroid histology / pimonidazole hypoxic-fraction data (cf. #196), and decide whether thresholds should be volume- rather than radius-based. Gradient strengths (MUFA/GSH/iron) are also placeholders.

### 3. #197 items 5-6 (out of scope in v1)
- **Cell-cell contact signaling**: E-cadherin/Merlin modulation of the ferroptosis threshold (enhanced in 3D).
- **Nutrient/waste gradients**: glucose / glutamine / lactate fields beyond O2 (#187) and pH (#190), affecting metabolic state + antioxidant capacity.

## Discipline (carry forward)
- Off-by-default byte-identical (matrix uses `Params::default()` + the random grid; #253 production-SHA guard + golden tests must keep passing).
- Pure helpers in `ferroptosis-core`, consumer applies.

## Related
[...]

PR #274, opened 2026-06-01
feat(spheroid): durable per-cell MUFA via Cell::mufa_cap (#270 item 1)

ELares:
Addresses item 1 of #270 (spheroid radial-biochem refinements) — the durable position-dependent MUFA flagged in the #197/#269 reviews. Direct parallel to the #266 durable-GPX4 fix.

## Problem

Spheroid position-dependent MUFA was an **initial condition**. `update_mufa_protection` relaxes `state.mufa_protection` toward the *global* `scd_mufa_max` (uniform M_ss ≈ 0.20) and clamps there each step — so a rim cell (set high) and a core cell (set low) both converged to the same uniform level. The peripheral-vs-core MUFA spread shaped only the early killing window, then washed out. Of the three radial axes only `iron` (a static `Cell` field) was durable.

## Fix (design per #270 — reuse mufa_surface/core as per-cell caps)

The MUFA steady state is set by the carrying capacity `scd_mufa_max`, so a **per-cell cap** gives a per-cell steady state:

- `Cell::mufa_cap: Option<f64>` — `#[serde(default, skip_serializing_if = "Option::is_none")]` (serialization-compatible; `None` ⇒ global `scd_mufa_max`).
- `update_mufa_protection(current, mufa_max, params)` now takes the effective cap; both call sites (`sim_cell_step`, `sim_cell`) pass `cell.mufa_cap.unwrap_or(params.scd_mufa_max)`. Logistic growth saturates at — and clamps to — the per-cell cap.
- `apply_radial_cells_3d` sets `cell.mufa_cap` radially (rim-high/core-low), reusing `SpheroidConfig::mufa_surface` (0.25) / `mufa_core` (0.05) — now the per-cell **caps** rather than transient initial values. The same radial value still seeds the initial `state.mufa_protection`, so a cell both starts at and relaxes toward its position-dependent level.

With `Params::spheroid()` (rate=decay=0.005, cap=0.25), a rim cap 0.25 → steady state ≈ 0.20, a core cap 0.05 → ≈ 0.048: the spread is now **durable** for the whole run.

## Byte-identity

Non-spheroid cells have `mufa_cap = None` ⇒ identical `update_mufa_protection` behavior. The default 24-condition matrix `summary.json` SHA-256 is **unchanged** (`1cf88663…390ff71`). `ferroptosis-core` 0.18.0 → **0.19.0** (additive public `Cell` field).

## Tests

- `per_cell_mufa_cap_gives_durable_per_cell_steady_state` (biochem): two MUFA levels start equal; the per-cell-capped one stays low while the global-cap (old uniform) one climbs to M_ss — pins the durability directly at the mechanism.
- `apply_sets_radial_mufa_cap_core_below_rim` (spheroid): the wiring — core cap < rim cap, within `[mufa_core, mufa_surface]`, stroma left uncapped.
- Full workspace green (224 core + 37 sim incl. `radial_spheroid_changes_kills_vs_random`); `cargo fmt --check` clean.

**Note:** `ferroptosis-python` is excluded from my local build — its PyO3 extension module fails to link standalone on a clean `origin/main` too (environment-only; it doesn't construct `Cell` literally, so the new field doesn't affect it). CI builds it on Linux.

Items 2 (zone-geometry calibration) + 3 (contact/nutrient gradients) of #270 remain open.


ELares:
## Code review — #270 item 1: durable per-cell MUFA

Reviewed the full diff (9 files, +141/−26). Clean, correct, and a faithful parallel to the #266 durable-GPX4 fix: the spheroid's rim-vs-core MUFA spread is now durable instead of relaxing to the uniform M_ss. Mechanism, byte-identity, and serde compatibility all check out. Findings below — one is a **process/CI risk** worth fixing before merge.

### ✅ Correct + verified
- **Mechanism is right.** `update_mufa_protection(current, mufa_max, params)` uses the per-cell cap for *both* the logistic saturation (`rate·(1 − current/cap)`) and the clamp, while keeping `rate`/`decay` global. So the steady state `M_ss(C) = rate·C/(rate + decay·C)` scales with the cap — a per-cell cap gives a per-cell durable steady state. Confirmed numerically: core cap 0.05 → ~0.048, rim cap 0.25 → ~0.20.
- **Byte-identity holds.** `mufa_cap` defaults `None` everywhere except `apply_radial_cells_3d` (spheroid, gated); `None ⇒ params.scd_mufa_max` ⇒ identical update. Default-matrix `summary.json` SHA is unchanged (`1cf88663…390ff71`). All 7 `Cell` literals (5 `gen_cell` + `gen_recovered_persister` + the FFI `ferro_cell_to_cell`) updated; `cell_to_ferro_cell` correctly left alone (FerroCell has no such field).
- **serde compatibility.** `#[serde(default, skip_serializing_if = "Option::is_none")]` means old `Cell` JSON deserializes (→ `None`) and `None` cells don't emit the field — Cell serialization is unchanged.
- **biochem test is excellent** — `per_cell_mufa_cap_gives_durable_per_cell_steady_state` proves the durability directly (capped stays low, uncapped/old-uniform climbs to M_ss), deterministic, no RNG.

### 🔴 Process risk — CI did not run
The **Rust Tests workflow (`fmt` + `test`) never triggered for this PR** — only CodeQL `Analyze` ran (all pass). For #271/#273 the `fmt`/`test` gates ran; here `gh pr checks` shows no Rust Tests run at all, and `gh run list` confirms no `cargo-test.yml` run on this branch. So **the PR currently has zero CI verification of the build, tests, or byte-identity** (and the Python CI — the one that would build `ferroptosis-python` on Linux with the new `Cell` field — also didn't run). I verified all of that locally (224 core + 37 sim tests, fmt clean, SHA unchanged), but the gate must actually run before merge. Re-trigger it (close/reopen the PR or push to the branch) and confirm green. Worth a quick look at *why* it skipped — `cargo-test.yml`'s `paths: [simulations/**, …]` should match this diff, so the skip is unexpected and could recur.

### 🟡 Minor
- **The sim_cell_step cap-wiring isn't directly test-guarded.** The biochem test calls `update_mufa_protection` directly (with explicit cap args), and the spheroid test only checks `cell.mufa_cap` is *set*. Neither would fail if the `cell.mufa_cap.unwrap_or(params.scd_mufa_max)` call in `sim_cell_step` were reverted to ignore the cap (the existing `radial_spheroid_changes_kills_vs_random` passed before this change too). Same shape as #266 finding #2. A cheap closer: a `sim_cell_step` test where a cell with `mufa_cap = Some(low)` ends at lower late MUFA than the same cell with `None`.
- **The 0.25 safety ceiling was removed.** The old `SpheroidConfig` doc warned `mufa_surface` was clamped at the global `scd_mufa_max` (0.25). Now the clamp uses the *per-cell* cap, so a future `mufa_surface > scd_mufa_max` would be honored (durably) rather than capped. Intended, but worth a one-line note that per-cell caps can exceed the global — otherwise a later config bump could silently produce supra-global MUFA.

### Note (not a defect)
The durability effect is concentrated in **core/mid** cells: rim cells start at 0.25 and settle ~0.20 in both the old and new code (≈unchanged), while core cells previously *rose* from 0.05 toward the uniform 0.20 and now stay ~0.048. So this PR makes core/mid cells durably MUFA-poorer (more vulnerable) — the intended direction. Also: the initial `state.mufa_protection` is seeded to the cap value (e.g. rim 0.25), which then relaxes down to `M_ss(cap)` — a small intended initial transient, not a steady-state.

### Verdict
Mechanism correct, byte-identical, well-tested at the unit level, version bumped appropriately (0.18.0 → 0.19.0 for the additive `Cell` field). The only real blocker is getting the **Rust + Python CI to actually run** (🔴) — don't merge on CodeQL alone. The two 🟡s are optional polish.

Issue #264, opened 2026-06-01
Immune realism Phases 2-4: suppressor field, multi-checkpoint axis, DC subsets (follow-up to #243)


ELares:
## Context

#243 was a four-phase plan; **Phase 1 (T-cell exhaustion) shipped in #261/#263 and #243 is now closed**, which orphans the later phases. Tracking them here so they aren't lost. Each builds on `immune_spatial` + the per-cell `cumulative_kills` / `exhaustion_factor` machinery now in place.

### Phase 2 — Treg / MDSC suppressor field
A second diffusing field (`suppressor_field`, mirroring the DAMP `diffuse_damp_3d_step` scratch-buffer pattern) that scales kill probability DOWN locally, opposing the kill effect; decays + replenishes over time. Initialized with patches at Treg-rich perivascular niches **if vasculature lands via #191** (otherwise seeded heuristically). Validation: anti-PD-1 alone less effective than anti-PD-1 + Treg depletion in the same run. Refs: Tauriello et al. Nature 2018 (TGFβ–Treg axis).

### Phase 3 — Multi-checkpoint axis
Replace the single PD-1 brake with a vector of checkpoint markers (PD-1, CTLA-4, LAG-3, TIM-3), each independently drug-modulated, to model anti-PD-1 + anti-CTLA-4 combinations. Refs: Sharma & Allison, Cell 2015.

### Phase 4 — DC subset complexity
cDC1 (cross-presenting, anti-tumor) vs cDC2 (Th17-skewing, less effective) with different DAMP responsiveness. Requires literature parameterization.

## Discipline (carry forward from Phase 1)
- Each phase off-by-default with an identity config so the default 24-condition matrix `summary.json` stays byte-identical (the #253 production-SHA guard + golden tests must keep passing).
- Pure helpers in `ferroptosis-core`, consumer applies (the `oxygen`/`ph`/`stromal`/`exhaustion` pattern).
- A demonstration test per phase (A/B with the mechanism as the only difference).

## Related
- #243 (Phase 1, shipped) — the baseline this extends
- #191 (vasculature) — Phase 2's Treg-niche initialization wants vessel positions
- #262 (persister refinements) — companion realism follow-up

## Source
Deferred phases from #243; #243's PR (#263) explicitly scoped Phase 1 only.

PR #275, opened 2026-06-01
feat(immune): Treg/MDSC suppressor field — immune realism Phase 2 (#264)

ELares:
Implements **Phase 2** of #264 (immune realism): a Treg/MDSC immunosuppressor field. Phases 1 (T-cell exhaustion) shipped in #263; Phases 3 (multi-checkpoint) + 4 (DC subsets) remain open.

## Why

Where exhaustion (#243) models *intrinsic* T-cell dysfunction, immunosuppression is *extrinsic*: Tregs/MDSCs cluster in niches and locally shut down T-cell killing (Tauriello et al., Nature 2018 — TGFβ–Treg axis). This adds a second diffusing field that opposes the DAMP→kill effect.

## What

**ferroptosis-core (`immune_spatial`, v0.19.0 → v0.20.0):**
- `suppressor_kill_multiplier(field, strength)` = `1/(1 + strength·field)` — identity at strength 0, the mirror of `exhaustion_factor` (it scales the kill *down*, not the activation).
- `suppressor_source_mask_3d(grid, cfg, vessels, seed)` — **perivascular** niches (within `niche_radius_um` of a vessel) when vasculature (#191) positions are supplied, else **heuristic** patches around `n_sources` random seed points sampled uniformly in the tumor sphere via an **independent** `StdRng` (cell grid untouched). Tumor cells only.
- `SuppressorConfig` (`disabled()` = byte-identical no-op; `enabled()` = literature-placeholder).

**sim-tme-3d:**
- `Overrides.suppressor`; a second field (`suppressor_field`) replenished at the source cells each step, then diffused + cleared via the existing `diffuse_damp_3d_step` (same `× 26 < 1` stability bound).
- The immune kill loop now applies **both** `exhaustion_factor` (#243) and the suppressor multiplier, each the identity when its layer is off. The allocation-free default kill path runs only when **both** are off, so it stays byte-identical (the rich `realism_kill_path` handles either/both; the `cumulative_kills` scatter is gated on exhaustion).
- `summary.json` gains `suppressor_source_count` + `suppressor_peak` (omitted when off). `--snapshot=suppressor` writes a static `suppressor.npy` niche mask; the Python renderer adds a "Treg/MDSC niches" panel (GIF render verified end-to-end).

## Byte-identity

Off by default: the matrix passes no suppressor config; `disabled()` (`suppression_strength = 0`) makes the multiplier `≡ 1.0` and neither the field nor its scratch buffer is allocated. Default-matrix `summary.json` SHA-256 **unchanged** (`1cf88663…390ff71`); golden + #253 production-SHA guards hold.

## Tests

- 3 core (`immune_spatial`): multiplier identity-when-disabled + monotonicity; heuristic seeding is tumor-only + deterministic + seed-sensitive; perivascular seeding clusters within the niche radius of the vessel.
- **`treg_depletion_improves_anti_pd1_kills`** (the #264 AC): anti-PD-1 with the suppressor on (Tregs present) vs off (Treg-depleted), all else equal — depletion materially recovers immune kills (>10%), i.e. anti-PD-1 alone is less effective than anti-PD-1 + Treg depletion (Tauriello 2018).
- Full workspace green (230 core + 38 sim); `cargo fmt --check` clean; renderer `py_compile`s and renders the GIF.

**Note:** `ferroptosis-python` is excluded from my local build (PyO3 standalone-link fails on a clean checkout too — environment-only; this PR doesn't touch `Cell`, so it's unaffected). CI builds it on Linux.

The suppressor parameters (`suppression_strength`, `replenish_rate`, `niche_radius_um`, `n_sources`) are uncalibrated placeholders (companion to #266/#270 calibration items). Phases 3–4 of #264 remain open.


ELares:
## Code review — #264 Phase 2: Treg/MDSC suppressor field

Reviewed the full diff (8 files, +541/−28). A substantial, well-structured new realism layer that faithfully mirrors the exhaustion (#243) pattern: a second diffusing field opposing the DAMP→kill effect. Mechanism, byte-identity, and the dual-modulator generalization all check out, and all three CI gates are green (incl. Python-on-Linux). Findings below — all minor; nothing blocking.

### ✅ Correct + verified
- **Byte-identity is sound, including the subtle case.** The default kill path runs only when BOTH layers are off (`realism_kill_path = exhaustion_on || suppressor_on`), so it's untouched. The non-obvious risk is the **exhaustion-only** path, which now computes `imm * exh * supp` with `supp = 1.0`: multiplying by `1.0` is the exact IEEE-754 identity (no rounding), so `(imm*exh)*1.0 == imm*exh` bit-for-bit — #243's output is preserved exactly (the `exhaustion_reduces_immune_kills` A/B still produces identical counts). Default-matrix `summary.json` SHA confirmed unchanged.
- **Both modulators guarded against OOB.** `exh`/`supp` are computed only under their `_on` flag, because each backing vec (`cumulative_kills`, `suppressor_field`) is allocated only when its layer is on — indexing them unconditionally would panic in the suppressor-only or exhaustion-only case. Correctly gated; the `cumulative_kills` scatter is likewise gated on `exhaustion_on`.
- **Field math is bounded + stable.** Replenish clamps to `[0,1]`; reuses `diffuse_damp_3d_step` so the `× 26 < 1` stability assert applies (0.025 → 0.65 ✓). With strength 6 and field ≤ 1, the multiplier floor is `1/7 ≈ 0.14` (≤86% local kill reduction) — sane.
- **Seeding is RNG-independent + coordinate-consistent** (lattice units, matching `place_vessels_3d`); tumor-only; deterministic. Good unit coverage (identity, heuristic, perivascular).
- **Validation test is the right A/B**: same anti-PD-1 config, suppressor on vs off, asserts depletion recovers >10% immune kills — the Tauriello 2018 result.

### 🟡 Minor
1. **Perivascular seeding isn't exercised end-to-end.** It's unit-tested (`suppressor_sources_perivascular_cluster_near_vessels`), but no preset or integration test enables suppressor **+** vasculature together — the `--snapshot=suppressor` preset and the validation test both use heuristic seeding (no vessels). So the `vessels.as_deref()` → perivascular branch through a full run is untested. A combined snapshot (or a one-line test running suppressor with a `VasculatureConfig`) would close it.
2. **Suppressor allocates but never evolves under `immune_on = false`.** The replenish+diffuse and the kill multiplier both live inside `if condition.immune_on`, but `suppressor_field`/`suppressor_on` are set up unconditionally — so an immune-off + suppressor-on run allocates the field, reports `suppressor_source_count` for niches that never activate, and `suppressor_peak = 0`. Unreachable today (all presets/tests are immune-on), but a `debug_assert!` or a doc note that the suppressor requires `immune_on` would prevent a confusing future config.
3. **`suppressor_peak` doc vs. computation.** The field doc says "peak … reached over the run", but it's computed as the end-of-run field max (`fold(max)` after the loop), not a tracked running max. The compute-site comment correctly notes the field is at quasi-steady-state so end ≈ peak; align the field doc to "end-of-run field max (≈ peak at steady state)" to avoid implying a per-step max was tracked.

### Nit
- Heuristic seeding uses a constant `SUPPRESSOR_SEED` (not per-condition `cond_seed`), so multiple suppressor-on conditions in one matrix would share an identical niche layout. That's the right call for the A/B (same patient, same niches), just worth being explicit about — a brief comment would help.

### Verdict
Clean, correct, byte-identical (the ×1.0 exactness makes the dual-modulator generalization safe), well-tested at both unit and integration level, viz verified, version bumped appropriately (0.19.0 → 0.20.0). The three 🟡s are all polish — #1 (perivascular end-to-end coverage) is the one I'd most want, and it's a small addition. Parameters are honestly flagged as uncalibrated. Good to merge after (or alongside) addressing #1.

PR #276, opened 2026-06-01
feat(immune): multi-checkpoint brake — immune realism Phase 3 (#264)

ELares:
Implements **Phase 3** of #264 (immune realism): a multi-checkpoint immune brake. Phases 1 (exhaustion, #263) + 2 (Treg/MDSC suppressor, #275) are shipped; Phase 4 (DC subsets) remains open.

## Why

The base immune model has a single PD-1 brake (`pd1_brake`, lifted by `anti_pd1_efficacy`). Real checkpoint blockade is multi-axis, and the clinically transformative result is **combination** therapy — anti-PD-1 + anti-CTLA-4 out-performs either alone (Sharma & Allison, Cell 2015). This generalizes the brake to a panel of independent, separately-drug-modulated checkpoints.

## What

**ferroptosis-core (`immune_spatial`, v0.20.0 → v0.21.0):**
- `Checkpoint { brake, drug_efficacy }`; `residual() = brake·(1−drug_efficacy)`.
- `CheckpointPanel` (`Copy`: PD-1, CTLA-4, LAG-3, TIM-3) with `combined_brake() = 1 − Π(1 − residualᵢ)` — independent brakes composing multiplicatively on the *pass*. A panel with only PD-1 active reduces **exactly** to the single-PD-1 `SpatialImmuneConfig::effective_brake` (`pd1_brake·(1−anti_pd1)`), so it's a strict generalization. Constructors `pd1_only` / `pd1_ctla4_tumor` + builders `with_anti_pd1/ctla4/lag3/tim3`.

**sim-tme-3d:**
- `Overrides.checkpoints`; the kill loop's `effective_brake` becomes `checkpoint_panel.map_or(immune_cfg.effective_brake(), |p| p.combined_brake())` — a uniform scalar, hoisted out of the hot loop exactly as before. `summary.json` gains `checkpoint_brake` only when a panel is supplied.
- `--snapshot=checkpoint` runs SDT + immune with **dual** PD-1/CTLA-4 blockade. No new render panel — the brake is a uniform scalar, so the lower combined brake shows as a more aggressive death front in the existing dead/DAMP panels (contrast the single-PD-1 `combined`/`multidose` presets).

## Byte-identity

Off by default: the matrix passes no panel, so the consumer uses `immune_cfg.effective_brake()` (single PD-1) unchanged. Default-matrix `summary.json` SHA-256 **unchanged** (`1cf88663…390ff71`); golden + #253 guards hold.

## Tests

- 3 core: PD-1-only panel `combined_brake` ≡ the single-PD-1 brake (the byte-identity equivalence); dual blockade < anti-PD-1 alone < untreated; the `1 − Π(pass)` independence math (two 0.5 brakes → 0.75, not 1.0).
- **`dual_checkpoint_blockade_outkills_anti_pd1_alone`** (the #264 Phase 3 AC): on a PD-1 + CTLA-4 tumor, anti-PD-1 alone leaves CTLA-4 braking; adding anti-CTLA-4 materially raises immune kills (>10%) — the combination-immunotherapy result.
- Full workspace green (233 core + 40 sim); `cargo fmt --check` clean; `--snapshot=checkpoint` runs end-to-end.

**Note:** `ferroptosis-python` excluded from my local build (PyO3 standalone-link fails on a clean checkout too — environment-only; this PR doesn't touch `Cell`). CI builds it on Linux.

Per-checkpoint brakes/efficacies are uncalibrated placeholders. Phase 4 (DC subsets) of #264 remains open.


ELares:
## Code review — #264 Phase 3: multi-checkpoint immune brake

Reviewed the full diff (7 files, +342/−9). A clean, minimal generalization of the single PD-1 brake to an independent-checkpoint panel — the strict-generalization design keeps the consumer change to one line and preserves byte-identity. fmt + Rust `test` are green (Python Tests correctly didn't fire — no Python-path changes; the PR only *adds* Rust types, no `Cell`/bindings change). Findings are all minor.

### ✅ Correct + verified
- **Byte-identity is sound and the reduction is a real property, not just a claim.** The consumer swap is `checkpoint_panel.map_or(immune_cfg.effective_brake(), |p| p.combined_brake())` — when `None` (the entire default matrix), this is *literally* the prior `immune_cfg.effective_brake()` line, so the matrix is untouched (SHA confirmed). And `pd1_only(b, d).combined_brake() == b·(1−d)` exactly (1 − (1−b(1−d)) = b(1−d)), pinned by `pd1_only_panel_matches_the_single_pd1_brake` — so the panel is a genuine superset of the old model.
- **Composition math is right + bounded for valid inputs.** `1 − Π(1 − residualᵢ)` with each `residualᵢ ∈ [0,1]` ⇒ result ∈ [0,1]; the "two 0.5 brakes → 0.75, not 1.0" test pins the independence semantics. The validation test confirms the brake feeds through to a real kill difference (dual blockade > anti-PD-1 alone, >10%).
- **No hot-loop cost.** `effective_brake` is a uniform scalar hoisted out of the per-cell loop exactly as before — the panel changes what's computed once, not per cell.

### 🟡 Minor
1. **No bounds validation on `Checkpoint` fields / builders.** `brake` and `drug_efficacy` are public `f64` and the `with_anti_*` builders accept any value. An out-of-range input (e.g. `brake = 1.5`, or a negative efficacy) makes `combined_brake` fall outside `[0,1]`, and `immune_kill_probability` then computes `activation·kill_rate·(1−effective_brake)` — a **negative** raw (silently → zero kills, since the cap only handles `raw > 0.99`) or an inflated one. No panic, just silently wrong. The module already guards `exhaustion_factor` and `suppressor_kill_multiplier` with `debug_assert`s; add the same to `combined_brake` (or `Checkpoint::residual`) asserting `brake, drug_efficacy ∈ [0,1]` for consistency.
2. **Multiplicative independence likely overstates combined suppression.** Real checkpoints are partially redundant/overlapping, so `1 − Π(1 − residual)` is an upper bound on the true combined brake. It's a defensible uncalibrated first-order model — worth a one-line note in `combined_brake`'s doc that it assumes independence (no cross-checkpoint redundancy), so a future reader doesn't take the combined magnitude literally.
3. **A panel silently overrides `SpatialImmuneConfig`'s `pd1_brake`/`anti_pd1_efficacy`.** When `Overrides.checkpoints` is `Some`, those immune-config fields are ignored (the test notes this). That's the intended precedence, but a caller setting both would be surprised — a doc note on the `checkpoints` field (or `CheckpointPanel`) that it *fully replaces* the single-brake fields would prevent the footgun.

### Notes (not defects)
- The brake is **uniform** across cells (no spatial checkpoint-expression heterogeneity) — consistent with the existing single-PD-1 model and a reasonable Phase 3 simplification; spatial checkpoint expression would be a later refinement.
- LAG-3/TIM-3 are wired but only exercised via the "inactive axes don't change the brake" assertion — they're structurally identical to the CTLA-4 term, so the CTLA-4 path covers the mechanism. Fine to leave.
- `--snapshot=checkpoint` correctly adds **no** new render panel (the brake is a uniform scalar; the effect shows in the existing dead/DAMP panels). Honest call.

### Verdict
Correct, byte-identical, well-tested (incl. the equivalence pin), one-line consumer change, version bumped (0.20.0 → 0.21.0). The only finding I'd genuinely want is #1 (the bounds `debug_assert`) — it's a real robustness gap and trivially closes to match the module's existing discipline. #2/#3 are one-line doc notes. Good to merge after #1.

PR #259, opened 2026-06-01
docs+CI housekeeping: refresh CLAUDE.md (#257) + crates.io flake env (#258)

ELares:
Post-merge debt from #255 (#192) and #256 (#253), neither of which touched `CLAUDE.md`. Two small, tracked gaps surfaced during the post-#256 regroup, bundled into one low-risk PR (docs + CI only).

## #257 — CLAUDE.md staleness

Verified on `main`: `CLAUDE.md` had zero mentions of `#192`, `--bench`, within-condition rayon, `#253`, or `sim-tme-3d-regression`.

- **Simulation workstream** now records the #192 work (`--bench` harness, within-condition rayon, byte-identical, 3.8x–4.9x single-large-grid speedup, dense 200³ at ~1.29 GB → sparse deferred to #254) and the #253 byte-identity regression CI.
- **Operational maturity** bullet now includes the #209/#236 `cargo fmt --check` gate and the #253 workflow.
- **Current Repo State + Key Files** now list `cargo-test.yml` (Rust CI + fmt gate) and `sim-tme-3d-regression.yml`; previously only `python-test.yml` appeared.
- **Count sweep** (the AC's "fix any other drift" item): binaries **11**, ferroptosis-core modules **16**, Python tests **105** (93 collected in `tests/` incl. parametrization + 12 in `ferroptosis-python/test_bindings.py`) — all confirmed accurate, no change needed.

## #258 — weekly-job network resilience

The first `workflow_dispatch` of the #253 job failed on a transient crates.io download (`curl failed: [16] Error in the HTTP2 framing layer`) before any build/sim ran; the rerun was green. For an unattended weekly job that's a false-red that looks like drift.

- Added job-level `CARGO_NET_RETRY: '10'` + `CARGO_HTTP_MULTIPLEXING: 'false'` (the documented mitigation for that framing error).
- **Determinism contract unchanged**: toolchain pin, flagless `cargo run --release`, and `sha256sum -c` are untouched; env only affects cargo's network layer.

## Verification

- `CLAUDE.md` edits confirmed present; counts re-derived from the tree.
- Workflow YAML parses; `env` block asserted; existing hardening (timeout, on-failure artifact upload, cron stagger, concurrency) intact.

Closes #257
Closes #258

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #257, opened 2026-06-01
Doc staleness: CLAUDE.md missing #192 perf, Rust CI + fmt gate (#209/#236), and #253 regression workflow


ELares:
## Context

CLAUDE.md was last refreshed in #252 (post-#239), but **#255 (#192) and #256 (#253) merged afterward and neither touched CLAUDE.md**. Verified on current `main`: `CLAUDE.md` has **zero** mentions of `#192`, `--bench`, within-condition `rayon`, `#253`, or `sim-tme-3d-regression`.

## Stale / missing

1. **Simulation workstream + repo-state** omit the #192 work: the `--bench` harness, within-condition rayon parallelism (byte-identical, 3.8x-4.9x single-large-grid speedup), and the measured memory/perf table (200³ at ~1.29 GB).
2. **CI / operational-maturity** bullet lists only `python-test.yml`. It omits `cargo-test.yml` (Rust CI + the #209/#236 `cargo fmt --check` gate) and the new `sim-tme-3d-regression.yml` weekly byte-identity guard (#253).
3. **Key Files** table lists only `.github/workflows/python-test.yml`; both Rust workflows are missing.

## Acceptance criteria
- [ ] CLAUDE.md workstreams / Current Repo State reflect #192 perf + the two-layer byte-identity guard (#253)
- [ ] CI bullet + Key Files list `cargo-test.yml` (incl. fmt gate) and `sim-tme-3d-regression.yml`
- [ ] Sweep for any other drift vs current main (binary count, module count, test counts) and fix in the same pass

## Source
Surfaced during the post-#256 regroup. Companion convention to prior doc-staleness issues #207 / #211 / #223.

PR #277, opened 2026-06-01
test(ci): assert documented Python test count matches collection (#260)

### LINKED ISSUE #260: CI: auto-assert the documented Python test count so CLAUDE.md/README can't silently drift
## Context

The #257 doc refresh re-derived the documented counts by hand (11 binaries, 16 ferroptosis-core modules, 105 Python tests = 93 collected in `tests/` + 12 in `ferroptosis-python/test_bindings.py`). The #259 review noted these live on lines the PR did not touch, so the diff could neither introduce nor prove them — they rest on a manual sweep.

The Python test count in particular has drifted before and is cheap to pin automatically.

## Proposed work

A tiny CI assertion (or pytest test) that fails when the **collected** Python test count diverges from the number documented in `CLAUDE.md` / `README.md`. Options:
- A pytest test that runs `pytest --collect-only -q` over `tests/` + `simulations/ferroptosis-python/` and asserts the total equals an expected constant, with a comment pointing at the two doc strings to update in lockstep; or
- A small CI grep step that extracts the `NNN Python tests` number from both docs and compares to the live collected count.

Keep it lightweight and self-documenting (the failure message should say exactly which doc line to update and how to regenerate the number).

## Acceptance criteria
- [ ] Collected Python test count is asserted against the documented number in CI
- [ ] Failure message names the doc line(s) to update
- [ ] Does not flake on environments where the compiled `ferroptosis_core` binding is unavailable (skip or account for the 12 binding tests explicitly)

## Source
Optional belt-and-suspenders suggestion from the #259 review (#257/#258 housekeeping).

ELares:
Closes #260.

## Why

CLAUDE.md and README.md advertise "NNN Python tests". That number was re-derived by hand in #257, rests on a manual sweep (the #259 review noted the diff couldn't prove it), and has drifted before. This pins it automatically.

## What

`tests/test_doc_test_count.py` — a self-documenting pytest test that fails, **in the normal `pytest tests/` CI run** (no extra workflow step), whenever the collected count diverges from the documented one:

- Collects `tests/` via a `pytest --collect-only` subprocess and asserts it equals `EXPECTED_TESTS_DIR`.
- Collects the `simulations/ferroptosis-python/test_bindings.py` bindings **only when the compiled `ferroptosis_core` extension is importable** — the Python CI does NOT build it, so the binding count is otherwise accounted for as a constant and never flakes (AC item 3).
- Asserts **both** CLAUDE.md and README.md advertise the total (`EXPECTED_TESTS_DIR + EXPECTED_BINDINGS`).
- Every failure message names the constant + doc line to update and how to regenerate the number (AC item 2).

## Acceptance criteria
- [x] Collected Python test count asserted against the documented number in CI (the existing `pytest tests/` step picks up the new test)
- [x] Failure message names the doc line(s) to update
- [x] Does not flake when the compiled binding is unavailable (collection failure → skip the 12, count them as a constant)

## Bookkeeping

Adding this test makes `tests/` 93 → 94, so the documented total goes 105 → 106; both doc strings updated in lockstep (the test enforces exactly this going forward). **Python + docs only — no Rust change**, so no `ferroptosis-core` version bump and the default sim matrix is untouched. Full `tests/` suite: **94 passed**; locally (with the binding installed) the test also re-counts the 12 bindings.


ELares:
## Code review — #260: assert documented Python test count in CI

Reviewed the diff (1 new test + 2 doc bumps, +94/−2). Clean, well-scoped, and closes all three AC boxes. The `test-linux` CI gate is green — which is itself the proof the no-flake design works (CI has no compiled binding, yet the test passed). Findings are minor polish.

### ✅ Correct + verified
- **The nested-collect is safe.** Running `pytest tests/ --collect-only` as a subprocess from inside `pytest tests/` does not recurse — `--collect-only` collects without executing, so the meta-test is *collected* (counted) but never *run* in the subprocess. Verified in CI (the gate passed) and the count is deterministic (94).
- **Every drift path is caught, with a guiding message.** Three things are pinned to agree: collected `tests/` == `EXPECTED_TESTS_DIR`, collected bindings == `EXPECTED_BINDINGS` (when available), and both docs' "NNN Python tests" == the total. Add/remove a test → the collection assertion fails and names the constant + doc lines + the new number; bump the constant but forget a doc → the doc assertion fails. The constant is the linchpin and the failure messages are genuinely actionable (AC #2 ✓).
- **No-flake on missing binding (AC #3 ✓).** `_collect_count` returns `None` on a non-zero pytest exit (e.g. the `ferroptosis_core` import failing during collection), so the binding assertion is skipped — exactly what let the CI run pass without the compiled extension, while the doc total still asserts against the constant 12.
- **Fails safe, not silent.** A reworded/missing "NNN Python tests" string → `search` returns `None` → a loud assertion, not a false pass. An uncollectable `tests/` → loud "could not collect". Good failure modes.
- **No workflow change, negligible cost.** Rides the existing `pytest tests/` step; ~0.6 s for two `--collect-only` subprocesses. AC #1 satisfied without touching `python-test.yml`.

### 🟡 Minor
1. **Scope is hardcoded to two paths.** The guard counts `tests/` + the single `ferroptosis-python/test_bindings.py`. I confirmed those are *currently* the only Python test files in the repo, so it's complete today — but a future test added in a **new** directory would be silently uncounted by both the docs and this guard (no failure, just a blind spot). Worth a one-line note in the docstring that the count is scoped to those two locations, so the blind spot is documented rather than surprising.
2. **`_DOC_COUNT_RE.search` matches only the first occurrence.** Each doc has exactly one "NNN Python tests" today, so it's fine — but if a second such phrase with a different number were ever added, the mismatch on the second wouldn't be caught. `findall` + "all equal EXPECTED_TOTAL" would be marginally more robust. Low priority.

### Nit
- The regex `(\d+) Python tests` is case-sensitive; a future lowercase "python tests" rewording would trip the "string not found" branch (a loud fail, so safe — just slightly surprising). Fine to leave.

### Verdict
Correct, fails-safe, no-flake (CI-proven), self-documenting, no workflow change, all three AC met. Both findings are one-line doc/robustness polish — not blockers. Good to merge; addressing #1 (the scope note) is the only thing I'd bother with, and it's optional.

Issue #242, opened 2026-05-21
Clonal heterogeneity: multiple tumor subclones with distinct ferroptosis profiles


ELares:
## Context

Real tumors are genetic mosaics — typically 4–8+ subclones (different driver mutations + passenger differences) coexist within a single tumor mass, spatially organized into patches/sectors. Different subclones have measurably different ferroptosis vulnerabilities: high-mesenchymal (ZEB1+) subclones are highly ferroptosis-vulnerable; epithelial subclones are resistant. The variance between subclones often exceeds the variance between treatments in real drug-screen data.

Current simulations treat **all tumor cells as biochemically identical** (single `Params` per simulation, only seed-driven noise distinguishes them). This is a substantial realism gap.

This is one of six realism gaps identified during the #193 visualization PR.

## Current state

- `Params` is a single struct applied uniformly across all tumor cells.
- `TumorGrid3D::generate` produces some seed-driven per-cell variation in initial state (basal_ros, iron) but **no systematic subclonal structure** — all variation is local noise, not spatial patches.
- The "tumor" in the simulation is effectively monoclonal.

## What "realistic" looks like

A grid where tumor cells belong to multiple subclones (2–6 distinct populations), each with their own ferroptosis-relevant parameter perturbations:

- **Spatial patterning**: subclones form contiguous patches/sectors (Voronoi or Perlin-noise-based assignment), not random mixing — matches real tumor histology.
- **Per-subclone parameters**: each subclone has perturbations of a baseline `Params` — e.g., subclone A has 2× MUFA protection, subclone B has reduced GPX4 reserve.
- **Per-subclone reporting**: kill rates broken out by subclone in summary.json so the impact of subclonal heterogeneity on overall efficacy is visible.

## Suggested approach

1. **Grid extension**: add `subclone_id: u8` to `GridCell` (or `CellState`). 0 = stromal/non-tumor; 1..=N = tumor subclones.
2. **Assignment**: in `TumorGrid3D::generate`, after deciding which cells are tumor, partition them into K subclones via Voronoi (K seed points seeded by RNG, each tumor cell assigned to nearest seed). Default K = 4.
3. **Subclone parameter table**: configurable list of `SubclonePerturbation { gpx4_inhib_mul: f64, mufa_cap_add: f64, iron_baseline_mul: f64 }` applied during `sim_cell_step` based on the cell's `subclone_id`.
4. **Output**: per-subclone summary fields in `ConditionResult`.

## Distinction from #197

[...]

PR #279, opened 2026-06-01
test(sim-tme-3d): cross-layer realism composition validation (#278)

### LINKED ISSUE #278: Cross-layer realism composition: integration tests + combined snapshot for multi-layer sim-tme-3d runs
## Context

sim-tme-3d now has ~9 independent, opt-in realism layers, each gated and byte-identical when off, each with its own test + (mostly) its own snapshot preset:

- persister (#241), T-cell exhaustion (#243), clonal heterogeneity (#242 + durable GPX4 #266), explicit vasculature (#191), spheroid radial biochem (#197 + durable MUFA #270), patient-scale slab (#240), Treg/MDSC suppressor (#264 P2), multi-checkpoint brake (#264 P3), time-varying dosing (#239).

Each is validated **in isolation**. The only multi-layer integration test in the suite is `suppressor_perivascular_runs_with_vasculature` (suppressor × vasculature, added in #275 because review flagged the gap). There is no systematic check that several layers **compose** correctly when enabled together.

## Why it matters

The layers are *designed* to be independent — distinct RNG seeds (`SUBCLONE_SEED` / `VESSEL_SEED` / `SPHEROID_SEED` / `SUPPRESSOR_SEED`), each identity-when-off, consumer-applied as RNG-neutral setup mutations. But "independent by design" is not "verified independent". Risks that only a composition test would catch:

- **RNG-seed collisions** — a future layer reusing a seed constant would silently correlate two layers' stochastic structure.
- **Field-mutation order/overlap** — several layers mutate overlapping `Cell` fields (e.g. clonal scales `iron`/`gpx4`/`nrf2`/`lipid_unsat`; spheroid re-gens the cell; slab/vasculature replace the O2/drug supply). Composition could double-apply or clobber.
- **Byte-identity drift** — the all-off default matrix is heavily guarded (#253), but there's no check that turning on layer A doesn't perturb layer B's *off* path.

## Proposed work

- A small set of integration tests enabling 2–3 layers together (e.g. clonal × suppressor × checkpoints; spheroid × vasculature; slab × suppressor) asserting: runs deterministically, produces directionally-sensible kills, and each layer's reported metric is present/coherent.
- A guard that the per-layer RNG seed constants are pairwise distinct (cheap, catches the collision class directly).
- Optionally a "kitchen-sink" `--snapshot` preset enabling several layers for a combined visual.
- Reaffirm: with **all** layers off the matrix stays byte-identical (already guarded; assert it still holds with the layers *available*).

## Acceptance criteria
- [ ] ≥1 integration test per a few representative layer combinations (deterministic + directional sanity)
- [ ] A pairwise-distinct assertion over the realism-layer RNG seed constants
- [ ] No regression to the default-matrix byte-identity (`summary.json` SHA) guard

## Discipline (carry forward)
- Off-by-default byte-identical (the #253 production-SHA guard + golden tests must keep passing).
- Pure helpers in `ferroptosis-core`, consumer applies.

## Source
Gap surfaced during the post-#277 regroup: the realism suite is now broad but cross-layer composition is only ad-hoc-tested (one pair, from a #275 review note).

ELares:
Closes #278.

## Why

sim-tme-3d has ~9 independent, opt-in realism layers (persister, exhaustion, clonal, vasculature, spheroid, slab, suppressor, multi-checkpoint, dose-schedule). Each is gated, byte-identical when off, individually tested, and seeded from its own RNG constant — but **composition** was only ad-hoc-covered (one pair, `suppressor × vasculature`, added in #275 from a review note). This verifies they compose, and guards the seed-independence the design relies on.

## What

- **`realism_layer_seeds_are_pairwise_distinct`** — asserts `SEED` / `SUBCLONE_SEED` / `VESSEL_SEED` / `SPHEROID_SEED` / `SUPPRESSOR_SEED` are pairwise distinct. A future layer reusing a constant (which would silently correlate two layers' stochastic structure) fails loudly here. (AC: seed guard.)
- **`clonal_suppressor_checkpoints_compose`** — clonal (#242) × Treg/MDSC suppressor (#264 P2) × multi-checkpoint brake (#264 P3), under immune: asserts the composed run is deterministic across runs and each layer reports a coherent metric (`subclone_kills` has 4 entries, `suppressor_source_count > 0`, `checkpoint_brake` present, immune ran) — i.e. no field/seed collision silently drops a layer.
- **`spheroid_vasculature_compose`** — the two grid-level layers (radial re-gen × vessel supply built from that grid) compose deterministically.
- **`--snapshot=combined-realism`** — kitchen-sink preset (persister + clonal + suppressor + checkpoints, SDT multi-dose + immune) rendering all three overlays (persister fraction + subclone id + Treg/MDSC niches) in one 6-panel GIF (verified). Deliberately excludes vasculature/spheroid/slab: they re-grid or replace the O2 source, which would desync the *static* overlays from the actual run.

## Acceptance criteria
- [x] ≥1 integration test per representative layer combinations (deterministic + directional sanity)
- [x] Pairwise-distinct assertion over the realism-layer RNG seed constants
- [x] No regression to default-matrix byte-identity (`summary.json` SHA `1cf88663…390ff71` unchanged)

## Bookkeeping

Tests + one snapshot preset only — **no `ferroptosis-core` API change** (no version bump) and **no Python change** (the #260 count guard, 106, is unaffected). Full workspace green (233 core + 43 sim, +3); `cargo fmt --check` clean; the 6-panel kitchen-sink GIF renders end-to-end.


ELares:
## Code review — #278: cross-layer realism composition validation

Reviewed the diff (3 tests + 1 snapshot preset + docs, +167/−1). A well-targeted hardening of the 9-layer suite, in the same verification spirit as #260. Rust CI green. Findings are minor (scope/framing); nothing blocking.

### ✅ Correct + verified
- **The seed guard catches the right bug class.** `realism_layer_seeds_are_pairwise_distinct` directly fails if a future layer reuses one of `SEED`/`SUBCLONE_SEED`/`VESSEL_SEED`/`SPHEROID_SEED`/`SUPPRESSOR_SEED` — the silent-stochastic-correlation risk the whole "independent RNG per layer" design rests on. Cheap, deterministic, high-value.
- **The composition tests catch the real composition risk: a layer being silently dropped.** The metric-coherence assertions (`subclone_kills` len == 4, `suppressor_source_count > 0`, `checkpoint_brake.is_some()`, `vascular_hypoxic_fraction.is_some()`) verify that enabling layer A doesn't clobber layer B's reporting — which is exactly what a field/seed collision would do. The `a == b` determinism checks guard against future-introduced nondeterminism. Assertions are robust (no flake: dense regime ⇒ `total_dead > 0`; 8 niches on a 24³ grid ⇒ sources > 0; `literature_4` ⇒ 4 subclones).
- **Kitchen-sink overlay-desync exclusion is correct.** I confirmed the reasoning: persister/subclone/suppressor overlays are geometry- or seed-only (the subclone map and suppressor mask depend on `is_tumor`/coords + a fixed seed, which clonal/persister perturbations don't change), so they match the actual run — whereas vasculature/spheroid/slab re-grid or change the O2 source and *would* desync the static `*.npy` overlays. Excluding them is the right call, and it's documented.
- **Byte-identity preserved** (tests + a preset only; SHA `1cf88663…390ff71` unchanged), no library/Python change (no version bump; the #260 count guard is untouched), 6-panel GIF renders end-to-end.

### 🟡 Minor
1. **The seed-guard list is hardcoded.** It checks the five current seed constants; a future layer that adds a sixth constant won't be covered unless someone adds it to the array (the same scope blind-spot flavor as #260's path list). Seed constants are arbitrary `const`s with no registry to auto-discover, so a one-line note ("add new realism-layer seed constants here") in the test is the pragmatic guard against the guard going stale.
2. **The tests verify *coexistence*, not combined-*effect* correctness.** They assert the composed run is deterministic and every layer still reports — but not that the layers combine in the expected *direction* (e.g. that the composed-with-suppressor kill count is below composed-without, or that dual modulators stack on the kill). That's a defensible scope (each layer's effect is validated in its own PR's A/B test), but the `_compose` names read a touch stronger than the assertions deliver. Optional: add one directional cross-check to a composition test to turn "they coexist" into "they interact as expected".

### Nit
- `assert!(a.immune_kills.is_some(), "immune ran")` — under `immune_on` this field is `Some(n)` even when `n == 0`, so it proves the metric is *reported*, not that immune killing *occurred*. Either tighten to `> 0` or reword the message to "immune metric reported" to avoid overstating.

### Verdict
Clean, robust, byte-identical, CI-green; the seed guard and metric-coherence checks are genuinely the right two things to pin for composition. Both 🟡s are framing/scope polish and the nit is a one-word message fix — none are blockers. Good to merge.

Issue #241, opened 2026-05-21
Drug-resistant persister cells: model epigenetic ferroptosis resistance acquired post-treatment


ELares:
## Context

When RSL3 (or any ferroptosis inducer) is applied to a tumor, a substantial fraction of cells that survive the first round enter a **drug-tolerant persister** state — slow-growing, transiently drug-resistant via epigenetic reprogramming. Mechanisms include NANOG/SOX2 upregulation, SCD1-driven MUFA membrane remodeling, mitochondrial state changes, and loss of ZEB1-driven mesenchymal vulnerability. Persisters can re-emerge as proliferating, drug-resistant clones over days. The simulation framework currently doesn't model this.

This is one of six realism gaps identified during the #193 visualization PR.

## Current state

- All surviving tumor cells at the end of the 180-step simulation are biochemically identical (same `Params`).
- No follow-up after step 180; no multi-cycle dosing (depends on [#239 multi-dose PK]).
- `Params::invivo()` includes SCD1-driven MUFA protection already as a steady-state — but it's static; it doesn't *increase* in response to drug exposure.

## What "realistic" looks like

- During drug exposure, surviving cells acquire **persister markers** stochastically:
  - Reduced lipid peroxide vulnerability (lower effective `gpx4_inhib`).
  - Increased MUFA protection over baseline `Params::invivo()`.
  - Reduced proliferation (modeled separately if proliferation lands).
- Multi-cycle simulation (depends on #239) would amplify the visible effect: dose 1 kills X cells, surviving cells re-equilibrate with persister state, dose 2 kills fewer cells because persisters are now enriched.
- Persisters revert over time (epigenetic, not genetic) — modeled as exponential decay after drug clearance.

## Suggested approach

1. **New state field**: `persister_fraction: f64 ∈ [0, 1]` on `CellState` (`ferroptosis-core`).
2. **Acquisition kinetics**: during cells' drug-exposure window, `persister_fraction` increments per step at rate proportional to local drug intensity (stochastic, seeded RNG per cell).
3. **Effect on biochem**: in `sim_cell_step`, scale GPX4 inhibition by `(1 - persister_resistance · persister_fraction)`, and scale MUFA cap by `(1 + persister_mufa_boost · persister_fraction)`.
4. **Reversion**: in the no-drug interval (between doses), `persister_fraction` decays exponentially.
5. **Output**: per-condition summary includes mean/distribution of persister_fraction at end of each cycle.

## Acceptance criteria

[...]

PR #280, opened 2026-06-02
feat(persister): competing-rate kinetics + calibration provenance (#262)

### LINKED ISSUE #262: Persister model refinements: competing-rate kinetics + parameter calibration (follow-up to #241)
## Context

#241 landed the drug-tolerant persister model (`ferroptosis-core::persister` + `PersisterConfig`, wired into sim-tme-3d). Two simplifications were explicitly deferred in the code and flagged in the #261 review (both non-blocking, accepted under the "placeholders pending calibration" framing):

### 1. Acquisition / reversion is a hard either-or
In the per-step loop, `acquire` runs when `drug_intensity > 0` and `revert` runs only when it is exactly `0.0`. So under sustained or sub-saturating dosing (`Constant`, or a `MultiDose` whose decaying tails never reach 0) `persister_fraction` is **monotonic non-decreasing** — reversion only fires in a truly drug-free window (an `Infusion` gap, pre-dose `Bolus`, explicit-zero `FromPk`). Biologically, acquisition and reversion are **competing rates that both operate**, giving a sub-cap equilibrium under sustained sub-saturating drug. A competing-rate update (`frac += acq(drug)·(max−frac) − rev·frac` each step) would be more faithful.

### 2. Parameters are uncalibrated placeholders
`PersisterConfig::enabled()` values (`acquisition_rate`, `reversion_rate`, `max_fraction`, `gpx4_resistance`, `mufa_boost_per_step`) were chosen to produce an observable effect, not fit to data. The literature gives qualitative direction (Hangauer 2017, Tsoi 2018), not step-level rates.

## Proposed work
- Replace the either-or with a competing-rate per-step update (acquisition and reversion both active); keep the identity-default byte-identical guarantee and the off-by-default gate.
- Add persister parameters to the calibration infrastructure (`simulations/calibration/`): document targets (e.g. fraction-surviving-vs-dose-cycles from Hangauer-style data) and a regeneration path, mirroring the existing parameter-provenance treatment.
- Consider whether the SDT/PDT path should also get a "reduced lipid-peroxide vulnerability" term (currently persisters resist SDT only via the MUFA axis, which is weak; the GPX4-resistance axis is RSL3-specific).

## Acceptance criteria
- [ ] Competing-rate kinetics (or a documented rationale for keeping the either-or)
- [ ] Persister parameters documented in calibration with sources/targets
- [ ] Bit-identical when the model is disabled (identity config); the #253 production-SHA guard still passes
- [ ] Test coverage for the sub-cap equilibrium under sustained sub-saturating dosing

## Source
Deferred from #241 / flagged in the #261 review (findings #2 and the calibration note).

ELares:
Closes #262 (the competing-rate kinetics + calibration items; the optional SDT/PDT vulnerability term stays deferred, documented).

## Why

The #241 persister model ran acquisition **xor** reversion, keyed on whether the drug was exactly `0.0`. So under `Constant` or a `MultiDose` whose decaying tails never reach zero, `persister_fraction` was **monotonic non-decreasing** — reversion only fired in a truly drug-free window. Biologically, acquisition and reversion are competing rates that *both* operate, giving a sub-cap equilibrium under sustained sub-saturating drug.

## What

**ferroptosis-core (`persister`, v0.21.0 → v0.22.0):**
- New `persister::step(frac, drug, cfg)` — one competing-rate update: `frac += acquisition_rate·drug·(max−frac) − reversion_rate·frac`, clamped. Both rates act every step, so sustained sub-saturating drug settles at `f* = acq·drug·max/(rev + acq·drug)` (below the cap) instead of ratcheting to `max_fraction`. Identity config (`acquisition_rate = 0`) ⇒ no-op. `acquire`/`revert` retained as the individual rate terms.

**sim-tme-3d:** the per-step consumer now calls `persister::step` (drops the `drug > 0 ? acquire : revert` branch).

**calibration:** a `PersisterConfig` section in `parameter_provenance.md` — the competing-rate model, the equilibrium formula, each parameter's qualitative source (Hangauer 2017 / Tsoi 2018), and the fraction-surviving-vs-dose-cycles calibration target + regeneration path. All rates flagged Assumed/placeholder.

## Acceptance criteria
- [x] Competing-rate kinetics (replaces the either-or)
- [x] Persister parameters documented in calibration with sources/targets + regeneration path
- [x] Bit-identical when disabled (the matrix runs no persister config; `summary.json` SHA `1cf88663…390ff71` unchanged; #253 guard holds)
- [x] Test coverage for the sub-cap equilibrium under sustained sub-saturating dosing

## Tests

New `step` tests: identity no-op; zero-drug pure reversion; **sub-cap equilibrium under sustained sub-saturating drug** + the analytic fixed point; equilibrium-stays-below-cap-even-at-full-drug (the competing-rate signature vs the old monotonic `acquire`). The existing persister-on sim tests (`persister_reduces_multidose_kills`, `persister_reverts_after_dosing_stops`) still pass — direction + off-byte-identity preserved. Full workspace green (237 core + 43 sim); `cargo fmt --check` clean.

**Note:** `ferroptosis-python` excluded from my local build (PyO3 standalone-link fails on a clean checkout too — environment-only; this PR doesn't touch `Cell` or the binding surface). CI builds it on Linux.

The SDT/PDT "reduced lipid-peroxide vulnerability" term (persisters currently resist SDT only weakly via MUFA) remains deferred — documented in the provenance section.


ELares:
## Code review — #262: persister competing-rate kinetics

Reviewed the diff (8 files, +147/−15). The core change is right and minimal: the either-or becomes one competing-rate integrator, and the calibration provenance is a genuinely useful addition. Rust CI green. One substantive finding (stale numbers the change itself invalidated) plus a latent-footgun note.

### ✅ Correct + verified
- **The competing-rate math is right** and the tests pin it well: `frac += acq·drug·(max−frac) − rev·frac` clamps to `[0, max]`, and the equilibrium `f* = acq·drug·max/(rev + acq·drug)` is asserted both empirically (converged fixed point) and analytically, including the key signature that reversion holds it below the cap *even at full drug* (where the old `acquire` reached the cap).
- **The `drug == 0` path is byte-identical to the old `revert`.** With `drug = 0`, `step` reduces to `(frac·(1−rev)).clamp(0, max)`, which equals the old `revert`'s `(frac·(1−rev)).max(0)` for any `frac ≤ max` (always true in practice). So only the `drug > 0` case changes behavior — exactly the intended fix — and the off-drug dynamics are preserved bit-for-bit.
- **Matrix byte-identity holds** (the matrix runs no persister config; SHA `1cf88663…390ff71` unchanged), the lower clamp can't bind for valid inputs (`frac + acq − rev ≥ frac(1−rev) ≥ 0`), and the calibration section honestly flags every rate as Assumed/placeholder with sources + a regeneration path (AC #2).

### 🟡 Findings
1. **The change invalidated specific persister-on numbers the PR didn't update.** The competing-rate model lowers `persister_fraction` (reversion is now active during a `MultiDose`'s decaying tails, not just in a fully drug-free window), so the kill reduction shrinks and `persister_mean` drops. Three cited figures are now stale:
   - `sim-tme-3d/README.md:140` — "cuts kills from 79 to 27" (the `on` value rises; reduction is no longer 66%).
   - `sim-tme-3d/README.md:141` — "persister_mean ≈ 0.49 in survivors".
   - the `persister_reduces_multidose_kills` comment "Observed ≈66%".
   The test still passes (it asserts `reduction > 0.2`, not the exact figure), but the prose/comment numbers should be re-derived from a current run. This is the one I'd fix before merge — it's the model's headline numbers contradicting the new model.

2. **`step`'s `acquisition_rate == 0` short-circuit now also gates reversion.** In the old split, `revert` ran independently of acquisition; now a single short-circuit returns `fraction` unchanged whenever `acquisition_rate == 0`, which *also* skips reversion. That's correct under the documented invariant (the identity config zeroes all rates), but a hypothetical "drug permanently withdrawn, reversion-only" config (`acquisition_rate = 0`, `reversion_rate > 0`) would freeze the fraction instead of reverting it. Unreachable with the two canonical configs, and the short-circuit isn't load-bearing for byte-identity (that comes from the consumer's `None` gate), so consider guarding on `cfg.is_identity()` instead of `acquisition_rate == 0` to make the intent collision-proof. Low severity.

### Nit
- `acquire` / `revert` are now unused by the consumer (kept as public primitives, documented as "the individual rate terms"). Fine to retain, just slightly redundant now that `step` is the per-step update.

### Verdict
Correct, byte-identical, well-tested, with a useful calibration write-up; the model change is sound. Fix #1 (re-derive the three stale persister-on figures) before merge — they're the model's advertised numbers and now contradict it. #2 is a one-line hardening of an unreachable case; the nit is informational.

PR #281, opened 2026-06-02
feat(clonal): spatial clonal expansion / dynamic selection (#266 item 3)

ELares:
Addresses item 3 of #266 (clonal evolution / dynamic selection). The inter-tumor-heterogeneity sub-item stays out of scope (separate ensemble model).

## Why

The clonal subclone map (#242) was **static** — a cell's subclone never changed, so the model had no "evolution" over a run. Real tumors under treatment undergo clonal selection: resistant subclones outlast and regrow into the territory vacated by dying vulnerable ones.

## What

**ferroptosis-core (`clonal`, v0.22.0 → v0.23.0):**
- `repopulate_dead_sites_3d`: each step, repopulate dead tumor sites from living Moore-26 neighbors — a revived site becomes a fresh daughter cell of a randomly-chosen living neighbor's subclone (its phenotype + `SubclonePerturbation`), and `subclone_ids` is rewritten. Resistant subclones (more living boundary neighbors) win more vacated sites ⇒ their territory grows. **Two-phase** (start-of-pass living snapshot → decide all repopulations against it → apply) + per-site RNG ⇒ deterministic and order-independent (a site revived this pass can't seed another the same step).
- `ClonalConfig::repopulation_rate` (default 0) + `with_repopulation(rate)` builder.

**sim-tme-3d:** call `repopulate_dead_sites_3d` each step after all deaths (clonal + `rate > 0` gated; per-site RNG derived from `cond_seed`, distinct from the setup seeds so it doesn't perturb assignment/placement). `SubcloneKillStat` gains `initial_tumor` (the at-assignment census) so `total_tumor − initial_tumor` is the net expansion.

## Byte-identity

Off by default: the matrix runs no clonal config, and `rate = 0` (every constructor's default) means `repopulate_dead_sites_3d` is a no-op. Default-matrix `summary.json` SHA `1cf88663…390ff71` unchanged; `literature_4()` keeps repopulation off, so the existing static clonal tests + behavior are untouched (the static test now also asserts `total_tumor == initial_tumor`).

## Tests

- core: `repopulate_is_noop_when_rate_zero`; `repopulate_revives_dead_sites_and_is_deterministic` (rate 1.0 revives dead sites from living neighbors, deterministic).
- `clonal_expansion_shifts_territory_toward_resistant`: under RSL3 + repopulation, the resistant subclone **grows** (1272 → 1294) and the vulnerable one **shrinks** (1401 → 1368); deterministic.
- Full workspace green (240 core + 44 sim); `cargo fmt --check` clean.

**Note:** with repopulation on, a site can flip dead→alive, so `total_dead` (end-state) ≠ cumulative kills (a revived site's death is still in `ferroptosis_kills`/`immune_kills`); the `total_dead == ferro + immune` invariant holds only with repopulation off — documented. Inter-tumor heterogeneity + a dynamic evolved-subclone-map viz overlay are out of scope/deferred (the static `subclone.npy` would desync from the evolved state). `ferroptosis-python` excluded from my local build (PyO3 standalone-link fails on a clean checkout too; this PR doesn't touch the binding surface).


ELares:
## Code review — #266 item 3: spatial clonal expansion

Reviewed the diff (7 files, +293/−12). This is the session's most invasive change — it adds proliferation (cells can revive), a genuinely new dynamic — and it's done carefully: gated, two-phase deterministic, byte-identical off. Rust CI green. The core is correct; my main finding is about how the *existing* kill metrics should now be read under evolution.

### ✅ Correct + verified
- **Two-phase update is genuinely order-independent.** Phase 1 snapshots `living` at start-of-pass; phase 2 decides every repopulation against that snapshot (donors filtered by `living[ni]`, reads `subclone_ids[donor]`/`phenotype` which phase 2 never mutates); phase 3 applies. So a site revived this pass can't seed another the same step, and the result doesn't depend on iteration order. A donor is always `living` ⇒ never itself a dead site being repopulated, so no interaction. Solid.
- **Deterministic + non-perturbing.** Per-site RNG mixes `(seed, idx, step)` with the seed derived from `cond_seed` (not a setup-seed constant), so it composes with the existing streams without perturbing assignment/placement — and the `#278` seed-guard stays valid (no new global seed). The repopulation pass is serial, running after the parallel kill passes join, so it's race-free.
- **Byte-identity holds two ways.** `rate ≤ 0` ⇒ early return (no-op), and the matrix runs no clonal config; `literature_4()` keeps `repopulation_rate = 0`, so the existing static clonal tests + behavior are untouched (the static test now even asserts `total_tumor == initial_tumor`). SHA confirmed.
- **Tests pin the right things**: no-op at rate 0; revive-from-living-neighbors + determinism; and the headline territory shift (resistant 1272→1294 grows, vulnerable 1401→1368 shrinks).

### 🟡 Findings
1. **Under repopulation the kill metrics change meaning — flag this more loudly than the one `total_dead` note does.** A site can die → revive → die again, so:
   - `ferroptosis_kills` / `immune_kills` become **cumulative death *events***, not cells killed — they can exceed the tumor population over a long run. The PR notes `total_dead ≠ ferro + immune`, but not that the counters can now exceed the population.
   - `subclone_kills.total_dead` is **end-state** dead (post-repopulation), so `kill_rate = end_dead/total` is "currently-dead fraction," not "fraction ever killed" — under evolution most dead sites were revived, so these rates read low and no longer mean what they do in the static model. The expansion test correctly uses `total_tumor` (territory) rather than `kill_rate`, but a one-line caveat in the docs that the per-subclone `kill_rate` is end-state-only under repopulation would prevent misreading it.
2. **Repopulation-on composition with the other layers is untested.** The `#278` composition tests use `literature_4()` (repopulation off). Repopulation is orthogonal (it only revives grid sites + rewrites `subclone_ids`), so it *should* compose with suppressor/checkpoints/etc., but a single composed run with `with_repopulation(>0)` would confirm it doesn't interact badly (e.g. a revived cell being correctly seen by the immune/suppressor passes next step). Low priority.

### Notes (not defects)
- **"Fitness" is emergent, not a parameter** — a subclone's expansion comes purely from having more *living* neighbors (= surviving more), not an explicit fitness term. That's a clean, honest modeling choice; worth stating so a reader doesn't look for a fitness knob.
- A repopulated cell starts `extra_iron = 0` and misses that step's `diffuse_iron` (created after it), so a newborn is iron-clean for one step — negligible.
- The per-site seed XOR-mixes two multiplied terms; a hash collision would only make two sites share an (independent) stream, not a correctness bug. Fine.

### Verdict
A careful implementation of a hard feature: deterministic, order-independent, byte-identical off, well-tested, with honest scope deferrals (dynamic-map viz, inter-tumor heterogeneity). The one thing I'd address is finding #1 — a docs caveat that the kill counters/`kill_rate` are death-events / end-state under repopulation, so they're not misread against the static model. #2 is optional. Good to merge after the caveat.

Issue #272, opened 2026-06-01
Patient-scale slab refinements: depth-collapse calibration, slab+vasculature coupling, real tumor geometry (follow-up to #240)


ELares:
Follow-up to #240 (patient-scale slab geometry, merged in #271). The v1 slab is a deliberate first-order model; this tracks the deferred refinements surfaced in the PR review (#271) and the module docs.

### Deferred / known limitations

- **Depth-collapse magnitude is uncalibrated.** The slab uses a first-order Krogh `exp(-depth/λ)` with a placeholder λ (~150 µm). The "<20% / near-zero kill at depth" headline is *illustrative of the in-vitro-vs-patient scale gap, not a validated efficacy number*. Real magnitude depends on tissue-specific penetration, vascular density, and drug PK. Calibration target: tie λ to the `drug_transport` module's 3 tissue types + the `tumor_pk` two-compartment model, and to a literature depth-vs-response curve.

- **Slab and explicit vasculature (#191) are mutually exclusive** (a `debug_assert` rejects both; slab supply wins). A real deep slab has *internal* vessels, not a single planar +z source. Natural extension: place a 3D vessel network (#191 `place_vessels_3d`) *within* the slab and combine the planar depth gradient with per-vessel proximity, so deep tissue can still have focal well-perfused pockets.

- **`virtual_tumor_mm` is cosmetic** — it only feeds the `scale_interpretation` string and the span `debug_assert`; it does not constrain a real geometry. A more physical model would derive `depth_offset` from a position in a defined tumor shape (e.g. distance-to-surface field of a large sphere) rather than a free parameter.

- **Uniform bulk phenotype mix.** `generate_slab` uses the flat 80/15/5 Glycolytic/OXPHOS/Persister "core" mix with no spatial structure. A deep slab could carry its own depth-graded biochemistry (the way #197 does radially) — e.g. more persister-like / hypoxic phenotypes toward −z.

- **Single planar supply axis.** Supply decays only along z (one-sided). Multi-face perfusion or a non-planar (curved front) supply would generalize the geometry.

### Scope notes
Each item is independent and gated, so it should preserve the default-matrix byte-identity invariant (summary.json SHA, guarded by #253). Calibration items likely need external data and may stay "placeholder pending calibration" until then — same posture as #270/#266.

Mirrors the per-module refinement-tracking pattern: #270 (#197), #268 (#191), #266 (#242), #264 (#243), #262 (#241).

ELares:
[...]

Issue #268, opened 2026-06-01
Vasculature refinements: spatial index for scale, fractal/micro-CT networks, spacing calibration (follow-up to #191)


ELares:
## Context

#191 landed explicit 3D vasculature (Option A: random internal vessel seeds + `vessel_supply_field`, wired into sim-tme-3d as a unified O2+drug supply). Three items were deferred in code/docs; tracking them here (companion to #262 / #264 / #266 realism follow-ups).

### 1. Spatial index for nearest-vessel (prerequisite for #240)
`vessel_supply_field` is brute-force `O(tumor_cells × vessels)` — fine at the 60³ matrix (~16M evals, one-time) but vessel count grows with tumor volume, so it is quadratic-ish at scale: a well-vascularized 200³ ≈ 3.8M cells × ~8900 vessels ≈ **34B** distance evals (minutes, not ms). **#240 (patient-scale ≥5 mm) cannot use vasculature until this is indexed.** Fix: a uniform spatial grid or kd-tree for nearest-vessel lookup (O(cells) with a bounded neighborhood). Must keep `vessel_supply_field` byte-identical for existing sizes (or a documented, deliberate change). Cross-references #240, #254 (sparse grid).

### 2. Fractal-branching / imported vessel networks (#191 Options B/C)
v1 uses random seed points (Option A). More realistic alternatives:
- **Option B**: fractal-branching vascular tree from the periphery inward (irregular branching, dead ends, shunts — real tumor vasculature is chaotic).
- **Option C**: import vascular networks from published micro-CT datasets of tumor vasculature.
Either changes only `place_vessels_3d` (the supply-field consumer is unchanged), so they slot in behind `VasculatureConfig`.

### 3. Calibrate inter-vessel spacing + Krogh λ
`well_vascularized()` (150 µm) / `poorly_vascularized()` (400 µm) are literature-ranged placeholders; the Krogh decay λ currently reuses the condition's O2 reference. Add to the calibration infrastructure with sources/targets (Vaupel inter-vessel distributions; pimonidazole hypoxic-fraction data, cf. #196). Could also model a distinct drug diffusion length (≠ O2 λ).

## Discipline (carry forward)
- Off-by-default byte-identical (matrix uses the edge-distance `radial_o2_field`; #253 production-SHA guard + golden tests must keep passing).
- Pure helpers in `ferroptosis-core`, consumer applies (the `oxygen`/`vasculature` pattern).

## Related
[...]

PR #282, opened 2026-06-02
feat(slab): couple patient-scale slab with internal vasculature (#272)

ELares:
Closes the **slab+vasculature coupling** item of #272 (the other two items — depth-collapse calibration and real tumor geometry — stay open, data/scope-gated).

## What
A planar +z supply source is a strong simplification: a real deep tumor carries **internal** vessels, not one feeding face. This couples the slab (#240) with the explicit vessel network (#191). Setting both `Overrides.slab` **and** `Overrides.vasculature` now combines them (previously slab supply simply won).

- **`vasculature::place_vessels_in_slab_3d`** (new, ferroptosis-core): scatters vessels **uniform-in-box** rather than `place_vessels_3d`'s central-sphere sampling, at the same density (shared `derive_vessel_count` helper). Vessels pervade the all-tumor slab so deep tissue *throughout* — not just a central pocket — gets perfused. Independent `VESSEL_SEED`; byte-identity preserved. Crate **v0.23.0 → v0.24.0**.
- **Combined supply (sim-tme-3d):** when both layers are set, each cell's unified O2/drug supply = element-wise `max(planar_depth, nearest_vessel)`, bounded in `[0,1]`. A deep cell next to a vessel gets a **focal well-perfused pocket** instead of monotonic depth collapse.
- **`--snapshot=slab-vessels`:** writes `vessel_supply.npy` recomputed on a **slab grid** (matching the run) so the O2-supply panel shows the combined field.

## Direction of effect (honest framing)
Supply scales drug/O2 **delivery**, so well-perfused ⇒ *more* drug ⇒ *more* death. Internal vessels therefore **raise** deep killing — the takeaway is that a patient-scale slab is **less therapy-resistant at depth** than the planar-only model (#240) implies, wherever real vasculature reaches. This is *not* "vessels rescue tumor"; it's "the drug reaches focal deep pockets the starved bulk escapes."

## Tests
- `slab_internal_vessels_increase_deep_killing` (deep slab, RSL3): vessels raise total kills vs the planar-only slab, deterministic.
- `slab_placement_is_deterministic_and_fills_the_box`: uniform-in-box placement reaches beyond the central sphere, stays in bounds, density matches the sphere placement.

## Safety / scope
- **Off by default / byte-identical:** gated on BOTH overrides; no matrix condition or other preset sets both. Full-production default-matrix `summary.json` SHA **unchanged** (`1cf88663…390ff71`, verified locally at 60³×180).
- Full workspace green (242 sim+core tests), `cargo fmt --check` clean, no new clippy warnings.
- **Calibration caveat (unchanged):** depth λ and inter-vessel spacing are uncalibrated placeholders; vessels are random-uniform, not a fractal/micro-CT network (a #268 follow-up). This shows the *mechanism*, not a quantitative depth-kill curve.

Docs: ferroptosis-core README, sim-tme-3d README (new "Internal vessels in a slab" subsection), CLAUDE.md version ladder + slab workstream entry.


ELares:
## Code Review — PR #282: slab+vasculature coupling (#272)

Adversarial self-review (I authored this). Overall: **correct, byte-identical, well-scoped, and honestly framed** — the direction-of-effect framing ("drug delivery ⇒ more deep killing, not rescue") is right and consistent across code/tests/docs. Verified locally: full workspace green, fmt clean, no new clippy, production 60³×180 matrix SHA unchanged (`1cf88663…390ff71`). A few findings below, one of which is a real correctness-of-documentation miss.

### Overview
- New `vasculature::place_vessels_in_slab_3d` (uniform-in-box) + shared `derive_vessel_count` helper.
- `supply_field` becomes a 3-arm match; when both `slab` + `vasculature` are set the per-cell supply is element-wise `max(planar_depth, nearest_vessel)`.
- `--snapshot=slab-vessels` preset + a slab-grid-matched `vessel_supply.npy` overlay.
- 2 tests (1 lib placement, 1 sim integration).

### 🟠 Findings to address

- **Stale comment now contradicts the code (`main.rs` ~L455).** The slab/spheroid `debug_assert` comment still reads *"(Slab vs vasculature is fine: slab supply simply takes precedence, documented at `supply_field`.)"* — but this PR is precisely what removes that precedence: slab + vasculature now **combine** (element-wise max). A reader hitting that comment will be actively misled about the new behavior. Update it to say the two now compose.

- **`place_vessels_in_slab_3d` silently assumes an all-tumor grid (`vasculature.rs`).** It samples uniform-in-box over `[0,rows)×[0,cols)×[0,layers)` while `derive_vessel_count` derives the count from `n_tumor` (tumor volume). On a slab that's exact (all cells tumor). But on a **sphere** grid (with stroma) it would scatter vessels into non-tumor space using a tumor-volume-derived count, under-vascularizing the actual tumor — a silent correctness trap for a future caller. It's safe today only because the sole call site is guarded by `slab_cfg.is_some()`. Recommend a cheap `debug_assert!` that the grid is all-tumor (e.g. `grid.cells.iter().all(|c| c.is_tumor)`), so misuse fails loudly rather than producing a quietly-wrong field.

### 🟡 Minor / optional

- **Performance note is now understated (`vessel_supply_field` docstring).** The cost note flags the `O(tumor_cells × vessels)` brute force and a 200³ sphere ≈ 34B evals, deferring a spatial index to #268. A **slab** is the worst case for this: every cell is tumor (vs ~30% in a sphere) *and* the box-volume-derived vessel count is higher, so the coupling makes patient-scale slabs ~3×+ more expensive than the sphere figure quoted. Worth one sentence so #268's urgency isn't undersold.

- **The MAX property is only tested indirectly.** `slab_internal_vessels_increase_deep_killing` validates the *outcome* (more kills) via two 24³×80 runs — good, but slow, and it doesn't pin the field invariant itself. A direct unit test (`combined[i] >= planar[i] && combined[i] >= vessel[i]` for all i) would be faster and tighter. Optional.

- **Modeling choice (acknowledged, not a defect):** element-wise MAX is a "nearest source dominates" Krogh approximation; true perfusion superposes where sources overlap. This was an explicit design decision (documented in the README as the cleaner bounded approximation) — flagging only for the record.

### ✅ Verified good
- Byte-identity: gated on **both** overrides; no matrix/preset sets both; `derive_vessel_count` extraction is pure arithmetic before any RNG draw, so `place_vessels_3d` output is unchanged (production SHA confirms).
- Snapshot overlay correctly regenerates on a **slab grid** with the same SEED/VESSEL_SEED/λ as the run, so `vessel_supply.npy` matches what the run applies (a real desync risk, handled).
- Determinism asserted; vessel count floored at 1 so `vessel_supply_field`'s non-empty assert can't trip; combined field bounded in [0,1].

### Verdict
Approve after fixing the two 🟠 items (both small: a comment + a debug_assert). The 🟡 items are optional polish.

PR #283, opened 2026-06-02
perf(vasculature): uniform-grid spatial index for nearest-vessel (#268)

ELares:
Closes **item 1 (spatial index)** of #268 — the prerequisite for vasculature on patient-scale (#240) grids. Items 2 (fractal/micro-CT networks) and 3 (spacing calibration) stay open.

## Problem
`vessel_supply_field`'s nearest-vessel lookup was brute-force `O(tumor_cells × vessels)`. Fine at the 60³ matrix (~16M evals) but quadratic-ish at scale: a well-vascularized 200³ ≈ **34B** distance evals (minutes), and the explicit blocker for vasculature on patient-scale grids — worst of all on an all-tumor **slab** (#272: every cell tumor + higher vessel count). This urgency was flagged in the #282 review.

## Fix — `VesselIndex` uniform-grid spatial index
- Bins vessels into a coarse lattice sized for ≈1 vessel/bin. A query expands outward in **Chebyshev shells** and stops as soon as the best distance found is provably ≤ the closest distance any *unsearched* bin could hold (after searching shell `R`, unsearched vessels are ≥ `R·bin` away). Conservative ⇒ **exact**.
- Returns the **same** `nearest_d2` as the brute-force min (min over the same finite set is order-independent), so `vessel_supply_field` is **byte-identical** — no output change, just faster. Roughly `O(cells)` with a bounded per-query neighborhood instead of `O(cells × vessels)`.
- Public API unchanged; the index is an internal struct built once per call.

## Verification
- **`spatial_index_matches_brute_force_bit_for_bit`** — recomputes the field by brute force and asserts `to_bits()` equality across 4 configs (sphere + slab, varied densities ⇒ varied bin sizes/occupancy). This is the byte-identity guarantee.
- **`spatial_index_is_fast_at_scale`** (`#[ignore]`, perf smoke): 100³ all-tumor slab, 2370 vessels, 1M cells → **~105 ms** indexed (brute force = 2.37B evals, ~minutes).
- Full workspace green (**294 tests**), `cargo fmt --check` clean, no new clippy.
- Default matrix doesn't use vasculature, so its `summary.json` SHA is unchanged (`1cf88663…390ff71`, re-verified at 60³×180); the bit-for-bit test is what guards the vasculature / slab-vessels paths.

## Scope
- ferroptosis-core **v0.24.0 → v0.25.0**.
- Exact acceleration only — no modeling change, no calibration. (Fractal networks + spacing calibration remain #268 items 2–3.)

Docs: ferroptosis-core README (vasculature row), CLAUDE.md (version ladder + vasculature workstream entry).


ELares:
## Code Review — PR #283: uniform-grid spatial index for nearest-vessel (#268)

Adversarial self-review (I authored this). Verdict up front: **clean, correct, and genuinely byte-identical** — the findings below are all minor/optional. The one thing that actually matters here (exactness of the acceleration) holds.

### Overview
Replaces `vessel_supply_field`'s `O(cells × vessels)` brute-force nearest-vessel scan with a `VesselIndex` uniform-grid (bin → Chebyshev-shell search → provable early-stop). Public API unchanged; index built once per call.

### ✅ The exactness argument holds (the thing that could've been wrong)
- After fully searching shell `R`, any unsearched vessel is in a bin at Chebyshev ≥ R+1, hence at lattice distance **strictly** > `R·bin` (because the query point `p < (cb+1)·bin` strictly). So `best ≤ (R·bin)²` ⇒ `best < d²` for every unsearched vessel ⇒ `best` is the true global min. The `≤` in the stop test is safe precisely because the unsearched bound is a strict `>`. Termination is conservative ⇒ exact.
- `min` over the same finite set is order-independent, so the indexed `nearest_d2` is bit-for-bit equal to the brute force. The `to_bits()` test across sphere+slab/varied densities is the right guard, and the default matrix doesn't touch vasculature so its SHA is trivially safe.
- Bin/cell index math: for valid cells `floor(r/bin) ∈ [0, nb-1]` (no clamp needed); vessel coords stay in `[0, dims)`; the clamp is purely defensive. No bin-boundary FP hazard (a boundary vessel lands in one bin deterministically and is still searched before any valid stop).

### 🟡 Minor / optional
- **Shell enumeration re-scans the full cube (`vasculature.rs`).** Each shell iterates the whole `(2R+1)³` index box and `continue`s the interior, so reaching shell `R` costs `O(R⁴)` index iterations (the vast majority skipped). Immaterial at the design point (≈1 vessel/bin ⇒ the nearest is almost always within R≤2; 153 iters to reach R=2), and both real placements (sphere vessels↔sphere cells, slab uniform vessels↔all cells) keep cells close to vessels. A true boundary-only enumeration would remove it, but it's not worth the extra code unless profiling a pathologically sparse grid says so. Worth one comment noting the characteristic.
- **`bins` allocates `nb.0·nb.1·nb.2` Vecs.** At physical spacing (≳100 µm ⇒ bin ≥ ~5 cells) this is small (≤ dims³/125). But `bin` clamps to `1.0`, so a *non-physical* `inter_vessel_um < cell_size_um` drives `nb → dims` and ~`dims³` Vec headers (≈190 MB at 200³). Can't happen with realistic inputs, but a one-line doc note (or `debug_assert!(bin >= 1.0)` is already implied) stating the index assumes spacing ≳ cell size would make the regime explicit.
- **Bin-index computation is duplicated** — the `bin_of` closure in `build` and the `cb` clamp in `nearest_d2` compute the same thing. Trivial; could share a helper.

### Performance
- Measured: 100³ all-tumor slab / 1M cells / 2370 vessels → ~105 ms (brute = 2.37B evals). The `#[ignore]` perf smoke is a nice touch (compiled/type-checked, not run in CI).

### Verdict
Approve. Findings are minor — recommend addressing the two 🟡 doc/comment notes (shell-scan characteristic + spacing≳cell-size regime); the DRY one is optional. No correctness or byte-identity concerns.

Issue #285, opened 2026-06-02
Generate the Tier-1 manuscript figures from the scientific-review roadmap (calibrated 2D results)


ELares:
Follow-up to the scientific review merged in #284 (`analysis/manuscript-scientific-review.md`, "Figure roadmap" section). Several of the manuscript's load-bearing quantitative claims are stated **only in prose** where a plot would be far more convincing. These all come from **calibrated 2D results** (safe to present quantitatively), unlike the uncalibrated 3D suite.

## Tier-1 figures to generate (calibrated, quantitative)

1. **Hypoxia kill-collapse vs O₂ penetration depth** — RSL3 vs SDT kill % over λ = 80/100/120/150 µm. The lynchpin mechanism (§7.1), currently one sentence. Data: `sim-tme`. **Caption must carry the post-#284 SDT-O₂-dependence caveat** (the "bypass" is a hypothesis; SDT's own O₂-dependence is contested).
2. **Bliss synergy** — RSL3 / FSP1i / combination, observed-vs-Bliss-expected bar chart (1.99×). Data: `sim-combo-mech`. Highest value-to-effort.
3. **Depth-kill curves** — PDT vs SDT (Beer-Lambert vs acoustic attenuation) + RSL3 uniform baseline. Data: `sim-tme` / `sim-tissue-pk`.

## Tier-2 (medium confidence, plot with caveats)
- Vulnerability-window kinetics (`sim-window`).
- Immune-coupling spatial DAMP — present the **2D** calibrated value, or caption the 104:1→4:1 3D discrepancy explicitly (per §8.4).

## Tier-3 (defer)
- pH ion-trapping (RSL3 pKa is the model's most uncertain parameter — defer until experimental validation).
- 3D-suite overlays: illustrative/mechanistic only, never quantitative.

## Optional flagship
- Integrated 2×2 "resistance-mechanism asymmetry" panel (hypoxia / stromal / pH / immune, RSL3 vs SDT). **Must not entrench the contested hypoxia leg** — caption each panel with its confidence tier. Build only after the Tier-1 set.

## Acceptance / mechanics
- New/extended generator scripts under `scripts/`; outputs to `article/figures/figN_*.{pdf,png}`.
- Wire each into `FIGURES.yaml` + the `figs` dict in `scripts/generate_latex.py`, add `[FIGURE N: ...]` placeholders in `v1.md`, regenerate `v1.tex` (must compile).
[...]

PR #284, opened 2026-06-02
docs: calibration-status consolidation + manuscript 3D-suite honesty pass

ELares:
An integrity/honesty pass (no code change), chosen at a deliberate pause in feature work. The `sim-tme-3d` track grew to ~12 "realism layers," almost all uncalibrated. An audit (manuscript + docs) found the real exposure is **not over-claiming** — it's **drift and silence**:

- The 112-page manuscript **does not mention the 3D suite at all**, while the repo ships it.
- Several §8.4/§7.1 "the model **cannot** do X" caveats are now **false**: the library *does* model adaptive evolution (`clonal`/`persister`), immune dampening (`suppressor`/checkpoints), and explicit vasculature — all uncalibrated.
- Stale inventory: Appendix A says version **0.2.0** (actual 0.25.0); Ch 11 says **"ten modules"** (actual 21); Appendix B's binary list omits **sim-tme-3d** and **sim-tumor-pk**.
- The honesty caveats that *do* exist were scattered across module docs, two READMEs, and `3d_validation_report.md` — no single accounting.

## What this PR does

**New: `simulations/calibration/CALIBRATION_STATUS.md`** — the single honest per-layer accounting. Every core + 3D layer tagged **Calibrated / Partially anchored / Self-consistency only / Uncalibrated (illustrative)**, with what data would calibrate each, plus a priority-ordered **calibration roadmap**. Linked from the calibration README. The headline it states plainly: *the suite is broad but mostly uncalibrated mechanistic scaffolding, excluded from the manuscript's quantitative claims.*

**Manuscript (`v1.md`; `v1.tex` regenerated):**
- Added a consolidated **§8.4 paragraph** "The 3D simulation suite, and why its numbers are not in this book" — names the suite, states its uncalibrated/self-consistency status, notes the deliberate exclusion from quantitative claims, and **surfaces the 104:1 → 4:1 immune-ratio shrinkage** seen when §7.2's keystone result is re-run in 3D.
- **Reconciled the stale "cannot do X" caveats** (§8.4 adaptive evolution, §7.1 vasculature): softened from "structurally cannot" to "the analyses in this book do not use these uncalibrated layers."
- **Drift fixes**: version 0.2.0→0.25.0; "ten modules"→accurate list + README as authoritative (anti-re-stale); binary list adds the two missing binaries; provenance pointer now also cites CALIBRATION_STATUS.md; the "adaptive evolution would be the most impactful extension" note updated (it now exists, uncalibrated).

[...]

PR #286, opened 2026-06-02
feat(figures): Tier-1 quantitative figures — hypoxia collapse + Bliss synergy (#285)

ELares:
Closes the Tier-1 portion of #285 (the figure roadmap from the #284 scientific review). Adds the two genuinely-new manuscript figures that turn prose-only quantitative claims into plots — both from **calibrated 2D results**, never the uncalibrated 3D suite.

## Figures added
- **Figure 21** (`fig24_hypoxia_killcurve`): RSL3 vs SDT kill, normoxic vs hypoxic, and robustness across O₂ penetration length λ=80–150 µm. Source: `sim-tme` (2D). Reproduces §7.1 **exactly** — RSL3 3.7%→0.1%, SDT 91.9%→87.8%. It's the quantitative companion to the *conceptual* Fig 15 cross-section. **Caption carries the post-#284 SDT-O₂-dependence caveat**: SDT is modeled as O₂-independent (an optimistic upper bound), so the *direction* is defensible but the *magnitude* of the gap is contested.
- **Figure 22** (`fig25_bliss_synergy`): RSL3+FSP1i observed (84.1%) vs Bliss-expected (42.2%) = **1.99× synergy**, plus pairwise synergy scores (SDT pairs excluded — 100% single-agent ceiling). Source: `sim-combo-mech`. Reproduces §6.3 exactly.

## Honest scoping
The third Tier-1 item (depth-kill curves) is **already Figure 8** ("depth-kill curves for PDT, SDT, RSL3"), so I did **not** add a redundant figure — noted, not duplicated.

## Data-source correctness
The infra map initially pointed the hypoxia figure at `tme-3d/summary.json` — the **uncalibrated 3D** output (gradient RSL3 ≈22%, not the 2D 0.1%). I used the correct 2D `sim-tme` source so the figure matches the §7.1 claim it illustrates.

## Mechanics & verification
- Generators in `scripts/generate_figures.py` (wired into `main()`); `FIGURES.yaml` entries (manuscript 21/22); `figs` dict captions (confidence-tiered); `[FIGURE 21/22]` placeholders in §7.1/§6.3.
- Input JSONs are gitignored (regenerate via the documented `cargo run` commands); only rendered pdf/png committed. Reverted the timestamp-only re-renders of the other 23 figures so the diff is just the 2 new figures.
- `v1.tex` regenerated, **pdflatex compiles clean (117 pp)**; figure-traceability + manuscript-inventory + doc-count guards all pass (`tests/` 96). Counts reconciled (20→22 manuscript figures, 23→25 files) in CLAUDE.md + README.md.


ELares:
## Code review: PR #286 — Tier-1 figures (hypoxia collapse + Bliss synergy)

Adversarial self-review (I authored this). The figures are correct and faithful to the data, but there's one real ordering bug to fix before merge, plus a few notes.

### ✅ Verified correct
- **Numbers match the manuscript exactly** (from the generator run): Fig 21 RSL3 3.7%→0.10%, SDT 91.9%→87.8%; Fig 22 RSL3+FSP1i 84.1% obs vs 42.2% expected = 1.99×. These reproduce §7.1 and §6.3 to the decimal.
- **Correct data source**: hypoxia uses 2D `sim-tme` (the source of the §7.1 numbers), NOT the uncalibrated `tme-3d` output (which would show ~22%). The infra map suggested the wrong one; the code uses the right one.
- Confidence-tiered captions, with the SDT-O₂-dependence caveat carried into Fig 21 (so the figure doesn't re-assert the "bypass" framing #284 softened). Depth-kill correctly NOT duplicated (already Fig 8). Timestamp-only re-renders of the other 23 figures reverted, so the diff is just the two new figures. Graceful skip if inputs are missing.

### 🟠 Must fix: figures are numbered out of reference order
In reading order the placeholders fall: Fig 8 (§6.1, L714) → **Fig 22** (§6.3, L758) → Fig 15 (§7.1, L789) → **Fig 21** (§7.1, L791). Two problems:
1. The two new figures are **reversed relative to each other**: the Bliss figure (§6.3) is referenced *before* the hypoxia figure (§7.1) but numbered 22 vs 21.
2. Fig 22 (§6.3) is referenced **before** the existing Figs 15–20 (all in §7.1), so a figure numbered 22 appears before figures numbered 15–20.

The manuscript otherwise follows strict sequential-by-reference numbering (Fig 8 in Ch 6 < Fig 15 in Ch 7). **Minimum fix:** swap the two new numbers — Bliss = 21 (§6.3, referenced first), hypoxia = 22 (§7.1) — in the `figs` dict, `FIGURES.yaml`, and the two placeholders. That restores correct order *between the new figures*. The deeper tension (any Ch-6 figure appended as 21+ still sorts after the Ch-7 figures 15–20) would only be fully resolved by a renumber cascade — author's call whether that's worth it, but the cheap swap should land regardless.

### 🟡 Minor / notes
- **Input reproducibility.** The committed pdf/png are snapshots of a local sim run; the input JSONs are gitignored and regenerated by `sim-tme` / `sim-combo-mech`. If those binaries are unseeded Monte Carlo, a re-run won't byte-reproduce the figure (headline numbers should be stable to ~1%, but not exact). Consistent with the existing figure pattern, but worth confirming the sims are seeded given the repo's reproducibility ethos.
- **No guard linking figure numbers to prose.** `test_figure_traceability` checks wiring/existence, not that Fig 21 actually shows 3.7%→0.1%. If sim output drifts, the figure could silently diverge from the §7.1 prose. Hard to CI without running the sims; flagging as a known gap.
- **Caveat redundancy.** The SDT-O₂ caveat appears both in the figure's embedded footnote and the LaTeX `\caption`. Belt-and-suspenders for an important point — defensible, but redundant.
- **String coupling.** `fig24` hard-codes the `gradient_{λ}um` condition-name format; a `sim-tme` rename would silently skip the figure (graceful, but a stale figure would persist).

### Verdict
Approve after the figure-number swap (🟠). Everything else is notes. No code-correctness or security risk; the figures are accurate and the data source is right.

<sub>Reviewed with Claude Code.</sub>

ELares:
## Review addressed (commit d362eab)

- **🟠 figure ordering — this was my own review error, not a bug.** On verification: `generate_latex.py` emits each figure as `\caption{} + \label{}` with **no hardcoded number**, so LaTeX assigns the printed "Figure N" by float **position**, and inline refs use `\ref`. The `[FIGURE N]` placeholder is only a `figs`-dict key. Proof it's decoupled: the existing manuscript's keys are already non-sequential — the key-20 figure's placeholder sits at line 202, key-9 at 242, key-6 at 938. So the §6.3 Bliss figure prints *before* the §7.1 hypoxia figure automatically regardless of whether their keys are 21/22 or 22/21; a swap is a cosmetic no-op on the rendered PDF. I'm therefore **not** making the swap — fixing a non-issue would be the wrong call. Record corrected.
- **🟡 input reproducibility — resolved (positive).** Both source binaries are seeded: `sim-tme` uses `SEED=42`, `sim-combo-mech` uses per-cell `seed_from_u64`. So the figures regenerate **deterministically** (exact reproduction). Documented this in both FIGURES.yaml notes.
- **🟡 remaining** (no prose-vs-figure CI guard; SDT-O₂ caveat in both footnote and caption; `gradient_{λ}um` string coupling) — accepted as minor/known: the prose guard would need sims running in CI; the caveat duplication is deliberate for a load-bearing point; the string coupling fails safe (graceful skip + "missing conditions" message).

Net: the only substantive item (reproducibility) is fine and now documented; the headline finding was a misread of the LaTeX numbering mechanism. Doc-only change; traceability test passes.

PR #292, opened 2026-06-03
feat(vasculature): fractal-branching vessel topology (#268)

ELares:
## What

Adds a **fractal-branching vessel topology** to the `vasculature` module, alongside the existing uniform-random placement.

- New `VesselTopology::{Random, Fractal}` enum + `topology` field on `VasculatureConfig` (default `Random`) + `with_fractal()` builder.
- New `place_vessels_fractal_3d`: a breadth-first branching vessel tree. Trunks enter from the tumor periphery and bifurcate inward with high, tumor-like angle/length variability and occasional dead ends, producing a hierarchical-but-chaotic network with avascular gaps rather than the smooth coverage of uniform-random points.
- Capped at the **same density target** as `place_vessels_3d`, so the two are directly comparable. At matched vessel density the fractal network leaves a **higher hypoxic fraction** (perfusion holes).

## Why

The original #191 model seeds vessels as uniform-random points, which over-smooths perfusion. Tumor vasculature is famously disorganized: high variability in segment length and branching angle, dead ends, and a higher space-filling fractal dimension (~1.89 vs ~1.70 normal) — **Baish & Jain, *Cancer Res* 2000, PMID 10919633**. This gives the simulator a more realistic, gappier perfusion field for the patient-scale work.

## Byte-identity

- `place_vessels_fractal_3d` uses an **independent** `StdRng(seed)`, never advancing the main grid stream.
- Off by default (`topology = Random`). Verified bit-for-bit: the full **60³ × 180 production matrix** `summary.json` still matches the pinned SHA `1cf88663…`.

## Tests

- `vasculature::tests::fractal_placement_is_deterministic_and_gappier_than_random` (library): determinism, density cap, fractal hypoxic fraction > random.
- `fractal_vasculature_is_hypoxier_than_random` (sim-tme-3d integration): end-to-end topology dispatch, determinism, fractal `vascular_hypoxic_fraction` > random.
- Full workspace: 243 lib + 48 sim-tme-3d + 4 doc-tests green; `cargo fmt --all --check` clean on the pinned 1.96.0 toolchain.

## Notes / scope

- sim-tme-3d dispatches on `cfg.topology` for the sphere/spheroid path; the slab path keeps its uniform-in-box placement (#272).
- Bumps `ferroptosis-core` 0.25.0 → 0.26.0. This version also carries the **#270 spheroid volume-based zone fix** (already merged behavior; version synced here). Appendix A version string updated and `v1.tex` regenerated.
- The branch/angle/length/dead-end constants encode the *qualitative* tumor-vessel chaos and are **not calibrated** against micro-CT morphometry — flagged in the doc comment as a follow-up.
- **No `--snapshot` preset yet** (would need ~a dozen preset-literal edits); deferred and noted in docs.


ELares:
## Adversarial self-review — #292 (fractal vessel topology, #268)

Reviewed the diff, re-derived the geometry, and **measured the central claim empirically** rather than trusting the test name. Verdict: mechanically clean and byte-safe, but the headline "gappier at matched density" needs an honest reframe. No blockers; several things to address or explicitly accept.

### What I verified (good)
- **Byte-identity holds.** `place_vessels_fractal_3d` uses an independent `StdRng(seed)`; default `topology = Random`. I ran the full 60³×180 production matrix and `summary.json` still matches the pinned SHA `1cf88663…` bit-for-bit. The off-by-default contract is intact.
- **Dispatch is minimal and correct** — the sphere/spheroid path swaps only the point generator; `vessel_supply_field`/`VesselIndex` consume fractal points identically to random ones (both are float lattice coords). No special-casing leaks.
- **Determinism is tested** (library `a == b`; integration `total_dead == total_dead`), constants are honestly flagged uncalibrated, and the Baish & Jain 2000 (PMID 10919633) grounding is real.
- **Empirical check of the claim** (40³ test grid): `target=58, fractal_n=58, random_n=58` — counts are *exactly* matched, so the higher hypoxia is **not** a count artifact here. `hyp_fractal=0.622` vs `hyp_random=0.031`.

### Substantive — the 20× gap is partly built-in (please reframe)
That 0.62 vs 0.03 is a ~20× difference, and it's near-tautological given the parameterization:
- `STEP = 1.0` ⇒ the 58 fractal points are a **1-cell-spaced filament**, ~58 contiguous cells of branch (across only **2 trunks** — `n_trunks = round(58/30) = 2`).
- `place_vessels_3d` spreads its 58 points **volumetrically** through a ~30k-cell tumor sphere.

So at equal *point count* the two networks have wildly different *effective coverage*: matching count is **not** matching perfusion density. "At matched vessel density the fractal network leaves a higher hypoxic fraction" overstates it as a controlled topology experiment — what's actually shown is "clustering N points onto a filament covers far less volume than scattering N points." Directionally right and qualitatively the intended behavior, but the magnitude shouldn't be read as a fair head-to-head.
- **Ask:** soften the README/CLAUDE.md/doc-comment wording to "same point-count target (not the same effective coverage)" or, better, compare on unique-covered-cells / total branch length. At minimum stop implying it's density-controlled.

### Substantive — the test tolerances hide the count confound elsewhere
- The library test allows `a.len() >= 0.5 * target` — a **2× undercount** passes. On grids/configs where the tree dead-ends or exits the sphere early, `fractal_n` could be well below `target`, and then the "gappier" result *is* count-confounded. It happens to be 58=58 on the current grid, but the assertion doesn't guarantee that.
- The **sim-tme-3d integration test checks no counts at all** — only `hyp_fractal > hyp_random`. If a future tweak makes fractal sparser, this test still "passes" for the wrong reason.
- **Ask:** assert counts are within a tight band of each other (e.g. `fractal_n >= 0.9 * random_n`) so the structural claim can't silently degrade into a count claim.

### Code quality
- **`for sign in [1.0, -1.0]` is statistically inert.** `perp(dir, &mut rng)` is re-rolled *inside* the loop and already applies a full 2π `roll`, so flipping it by `sign` just yields another uniformly-distributed perpendicular. The two children are independent random forward branches, **not** a symmetric bifurcation — despite reading like one. Either hoist `let axis = perp(...)` out of the loop (true Y-bifurcation, more anatomically faithful) or drop `sign` for `0..2`. As written it's misleading.
- **Silent topology ignore on the slab path.** `with_fractal()` is a **no-op** for slab geometry (`if slab_cfg.is_some()` always calls `place_vessels_in_slab_3d`). Intended per the PR body, but it's a footgun — document it on `with_fractal()` and/or `VesselTopology::Fractal` ("ignored for slab geometry").
- **`n_trunks = (target/30).clamp(2,64)`** is an uncalibrated magic ratio that strongly shapes the result (2 trunks for the test). Already covered by the "uncalibrated" hedge, but worth a word on *why* /30.
- Branch crossings can push near-duplicate points that still count toward the `target` cap, marginally lowering unique coverage. Harmless, noted.

### Risk summary
Low. Off-by-default + verified byte-identity means zero blast radius on the production matrix; the issues above are about **interpretive honesty and test strength**, not breakage. I'd merge after the wording reframe (substantive #1) + a count-band assertion (substantive #2); the `sign`/bifurcation cleanup and the slab-no-op doc note are nice-to-haves.

PR #287, opened 2026-06-02
feat(figures): treatment-timing window figure — Tier-2 (#285)

ELares:
Adds the Tier-2 figure from the #285 roadmap: **Figure 23** (`fig26_vulnerability_window`), turning the §6.2 prose/table into a plot. From calibrated-ish 2D data (`sim-window`, seeded/deterministic; **medium confidence** — defense-recovery half-times are literature-estimated).

## What it shows
- **(a)** Post-chemotherapy kill over time: RSL3 collapses **42.4%→1.4% by day 3** and to ~0 by day 7 (GPX4 re-expression), while SDT/PDT hold **~100%→99.5%** through day 28. Shaded "RSL3 window open" region + CI bands. Reproduces the §6.2 table exactly.
- **(b)** The mechanism — RSL3 kill (left axis) tracks **mean GPX4 recovery** (right axis, twin): the re-expression of the primary defense is what closes the window.

The clinical point: RSL3 needs intervention within days; physical modalities offer a multi-week scheduling window.

## Why this figure (and not the flagship)
Per the regroup, the flagship 2×2 asymmetry panel was deferred because it embeds the *contested* hypoxia leg + *deferred* pH; this one is a single, clean, calibrated-ish mechanism with low over-claim risk.

## Mechanics & verification
- Generator in `generate_figures.py` (wired into `main()`); `FIGURES.yaml` entry (manuscript 23, determinism noted); `figs` dict caption with the medium-confidence caveat; `[FIGURE 23]` placeholder in §6.2.
- Input JSON gitignored (regenerate via `cargo run --release -p sim-window`); only rendered pdf/png committed; timestamp-only re-renders of the other 25 figures reverted (diff is just the new figure).
- Reproduces §6.2 numbers exactly; `v1.tex` compiles clean (117 pp); figure-traceability + manuscript-inventory + doc-count guards pass (`tests/` 96). Counts updated (22→23 manuscript figures, 25→26 files).

#285 remains open for the flagship 2×2 panel + Tier-3 pH figure (deferred until calibrated).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #293, opened 2026-06-03
feat(figures): Tier-1 depth-kill curves — tracked generator for Figure 8 (#285)

ELares:
## What

Tier-1 #3 from the scientific-review figure roadmap (#285): the **depth-kill / tissue-penetration** figure.

The manuscript already carried **Figure 8** (depth-kill, §6.1), but its FIGURES.yaml generator was `function: null` ("figure produced by the binary or external post-processing") — i.e. **not reproducible**. This PR replaces it with a tracked Python generator and upgrades the content to the Tier-1 spec.

## The figure

`fig8_simulation_by_treatment()` reads `sim-spatial`'s `depth_kill_curves.csv` (calibrated 2D physics). Two panels:

- **(a) Observed kill vs depth** across 1 cm of tissue: SDT (ultrasound, acoustic) stays ~95-100% throughout; PDT (light, Beer-Lambert) collapses from ~93% at the surface to ~0% by 10 mm; RSL3 (systemic drug) is a flat, near-zero baseline at every depth.
- **(b) The driving penetration physics**, computed from the model's own equations (`physics.rs` defaults): PDT light `exp(-μ_eff·z)` with μ_eff=0.31/mm (δ≈3.2 mm), SDT acoustic `10^(-α·f·z/10)` with α=0.7 dB/cm/MHz at 1 MHz, RSL3 uniform at 100%.

The story: light reaches **millimeters**, ultrasound reaches **centimeters**, and RSL3 reaches every depth but kills little — a **biochemical** limit, not a penetration one.

## Honesty / confidence tier

- The caption states the tier explicitly: the **depth-profile physics is high-confidence** (Beer-Lambert optics, acoustic attenuation are well-measured; confirmed sound in the #284 scientific review), but **absolute kill % rests on uncalibrated biochemistry**, so the profile *shape* is the result, not the magnitudes.
- Carries the post-#284 **SDT-O₂-dependence caveat** (SDT modeled O₂-independent = optimistic upper bound, §7.1), as the issue requires.

## Traceability

- FIGURES.yaml fig8 entry now points to the tracked generator (`scripts/generate_figures.py::fig8_simulation_by_treatment`) + the CSV input — retires a `function: null` gap.
- `v1.md` placeholder + `generate_latex.py` caption updated; `v1.tex` regenerated.

## Verification

- `tests/test_figure_traceability.py` + `tests/test_manuscript_inventory.py` pass; **full Python suite 96/96 green**.
- `pdflatex` compiles `v1.tex` with no errors (1.7 MB PDF).
- `v1.tex` diff is exactly the fig8 caption (2 lines); figure is deterministic (seeded `sim-spatial`).

## Reproduce

```
cd simulations && cargo run --release -p sim-spatial
python scripts/generate_figures.py   # (or just fig8_simulation_by_treatment())
python scripts/generate_latex.py
```

## Notes / scope

- Replaces panel (b)'s old "2D death heatmaps" with the penetration-physics panel — the Tier-1 ask is specifically the Beer-Lambert vs acoustic vs uniform comparison. Heatmaps remain available as `spatial_death_*.csv` and are covered by figs 17/18.
- Completes the Tier-1 figure set (#286 hypoxia + Bliss, this depth-kill); Tier-2 window shipped in #287.


ELares:
## Adversarial self-review — #293 (Tier-1 depth-kill figure, #285)

Traced the data path end-to-end and checked the caption numbers against the actual binned output rather than trusting them. Verdict: faithful, honest, and well-tiered; one real maintainability finding (a decoupled physics re-implementation), the rest nits. No blockers.

### Verified (good)
- **Panel (b) physics is faithful to the sim, confirmed.** `local_ros_multiplier(row, ...)` computes `z = row × cell_size` and dispatches to `sdt_intensity_at_depth` / `pdt_intensity_at_depth` / `rsl3_concentration` — exactly the three formulas panel (b) reproduces. The hardcoded constants (μ_eff=0.31/mm, α=0.7 dB/cm/MHz, i0=1) match `params.rs` defaults, and `sim-spatial`'s defaults (`--dli-h 0`, `Photosensitizer::Uniform(1.0)`) give PDT yield = 1.0 — so under the documented regen command (`cargo run --release -p sim-spatial`) panel (b) **equals** the sim's per-cell multiplier. The "depth from irradiated surface" axis is correct (energy enters at row 0).
- **Caption numbers check out** against the binned data (PDT ~93%→~0% over 10 mm; SDT ~97%→100%, within the stated "95 to 100%"). Confidence tier + SDT-O₂-dependence caveat present, as #285 requires.
- **Right call** weighting the depth bins by tumor-cell count — suppresses noise from the spheroid's sparse poles.
- Retires a `function: null` traceability gap; figure + inventory tests pass; `pdflatex` compiles.

### Substantive — panel (b) is a decoupled re-implementation (silent-drift risk)
Panel (a) comes from the Rust sim; panel (b) is a **Python copy** of the physics with **hardcoded** constants, and it implicitly assumes the **default** sensitizer/DLI (yield = 1.0). Two ways the panels can silently desync:
1. If `params.rs` defaults are retuned (e.g. `pdt_mu_eff`, `sdt_alpha`), panel (a) follows; panel (b) does not.
2. If the CSV is regenerated with non-default flags (`--photosensitizer porfimer=…`, `--dli-h 24`), panel (a)'s PDT is scaled by the drug yield while panel (b) still draws `exp(-0.31 z)` at yield 1.0.

Either makes the figure internally inconsistent, and nothing guards it (there's a comment citing `physics.rs`, no test).
- **Ask:** add a tiny guard test asserting `params.rs` still carries `pdt_mu_eff=0.31` / `sdt_alpha=0.7` (cheap, catches facet 1), and strengthen the panel-(b) comment to "assumes default `sim-spatial` flags (yield=1.0); regenerate the CSV with defaults or update these curves." Reading the constants from a shared source would be the fuller fix.

### Minor
- **Unused variable:** `sx, sy = series["SDT"]` — `sx` is never used (F841-class). No Python lint gate so it won't fail CI, but it's sloppy; drop it (`_, sy = …`).
- **Unguarded final print:** `py[0]` / `px[-1]` would `IndexError` on an empty curve, while the PDT annotation guards with `if py:`. PDT always has cells so it never triggers, but the inconsistency is worth tidying.
- **Dual meaning of green across panels:** in (b) the green line is ~100% (RSL3 *availability*); in (a) it's ~0% (RSL3 *kill*). The caption resolves this ("reaches every depth yet kills little, a biochemical limit, not a penetration one"), so it's acceptable, but a reader could momentarily misread; a different RSL3 style in (b) would remove all doubt.

### Note (not blocking)
The committed PDF is trust-based — no regenerate-and-diff test (figures aren't byte-stable across matplotlib versions), so a stale PDF wouldn't be caught. General repo pattern, not new here. SDT's slight 97→100 non-monotonicity is a sparse-pole small-sample artifact; the caption's "near 95 to 100%" doesn't overclaim.

### Risk
Low. Additive figure + docs; no sim/runtime code path touched, byte-identity not in play. I'd merge after the panel-(b) coupling guard (substantive) lands; the variable/print/colour items are nits.

Issue #295, opened 2026-06-03
Test-coverage gaps: untested core formulas (wilson_ci, 2D immune, persister-recovery), unguarded C-FFI ABI, figure data-drift


ELares:
Surfaced by the regroup gap-analysis (test/CI-coverage sweep). Several pieces of live, shipped code have **no test coverage**, and one stated integration point is never exercised in CI. None of these is tracked by an existing issue. Grouping them here (the #289 consolidation pattern). Each is independent.

## Untested core formulas (silent-skew risk)

1. **`ferroptosis_core::stats::wilson_ci`** — `simulations/ferroptosis-core/src/stats.rs:30`. Pure closed-form Wilson-score binomial CI that produces the `ci_low`/`ci_high` reported by **sim-original, sim-window, sim-invivo, sim-tissue-pk, sim-combo-mech, sim-tumor-pk, and the Python bindings**. Zero `#[test]`. A wrong z-constant or formula error would silently skew every reported confidence interval. Easy, high-value: known-value asserts (p=0.5/large n → symmetric; n=0 → (0,0); bounds clamped to [0,1]; monotonic in k). This is the "invariant/formula" coverage #139 added for tagging/weights but never extended to the reported statistics. **Medium.**

2. **2D `immune` module** — `ferroptosis-core/src/immune.rs:41` (`calculate_damp_release`), `:61` (`immune_cascade`). Live code consumed by **sim-icd and sim-combo**, zero `#[test]`. By contrast the 3D successor `immune_spatial` carries 28 tests (exhaustion/suppressor/multi-checkpoint). The older 2D DAMP/T-cell cascade that still drives two shipped binaries is entirely unguarded. **Medium.**

3. **`cell.rs` persister-recovery model** — `ferroptosis-core/src/cell.rs:159` (`recovered_persister_means`), `:173` (`gen_recovered_persister`). The exponential post-chemo FSP1/GPX4/NRF2/GSH re-expression path consumed by **sim-combo and sim-window**, zero `#[test]`. Distinct from `persister.rs`'s competing-rate step (#241/#262), which is well-tested. A known-value test (frac at t=t_half → 0.5; means converge as days→∞) would anchor it. **Low.**

## CI gaps

[...]

Issue #289, opened 2026-06-02
Re-home 3D follow-up TODOs orphaned by #194 closure (cluster-radius retune, field perf-hoist, ImmuneConfig dedup)


ELares:
Surfaced by the regroup gap-analysis (Rust deferred-work sweep). Several code TODOs were deferred "to #194 (sim-spatial-3d)", but **#194 is now CLOSED** — so these follow-ups are untracked and would otherwise be lost.

## Orphaned items (all in `simulations/ferroptosis-core/` + `sim-tme-3d/`)
1. **3D persister cluster-radius volume retune** — `grid.rs:421-428`. `cluster_radius=4.0` yields ~268 cells in 3D vs ~50 in 2D (clusters ~5× larger relative to tumor volume than the 2D analog). Volume-equivalent would be `cluster_radius ≈ 2.4`; v1 kept 4.0 to match 2D code shape. Affects 3D persister-cluster realism; needs a deliberate retune (and ideally calibration against cluster-size data). **Medium** — changes 3D cluster sizing.
2. **Per-call geometry-constant hoist** — `oxygen.rs:91-93` (`radial_o2_field`) and `ph.rs:105` (`radial_ph_field`) recompute `(center_r, center_c, center_l, tumor_radius)` every call. Negligible at current N, but `sim-tme-3d`'s inner loop should hoist them once. **Low** (perf).
3. **`ImmuneConfig` struct duplication** — `sim-tme-3d/src/main.rs:110-117` duplicates a config struct that could be lifted into / shared from `ferroptosis-core`. **Low** (DRY).

## Why file this
These are real, located TODOs with no live tracking issue (their referenced #194 is closed). Small housekeeping; item 1 has genuine modeling impact and should be done deliberately rather than discovered later. Items 2-3 are quick polish.

Source: regroup gap-analysis (Rust deferred-work sweep), file:line cited above.

ELares:
Regroup gap-analysis found a fourth orphaned item with the same pattern (deferred to a now-CLOSED issue), to add to this issue's scope:

[...]

PR #296, opened 2026-06-03
chore(honesty): fix stale "deferred" comments contradicted by shipped code + calibration-doc fractal row

### LINKED ISSUE #285: Generate the Tier-1 manuscript figures from the scientific-review roadmap (calibrated 2D results)

ELares:
## What

A documentation/comment honesty pass surfaced by the regroup gap-analysis. Several comments and docs froze their pre-merge framing after the underlying work actually shipped — the "letting historical framing survive after the underlying analysis changed" failure `CLAUDE.md` explicitly warns against. **All comment/doc-only; no logic changed; byte-identical** (production matrix `summary.json` SHA verified unchanged).

## Fixes

- **persister** (`sim-tme-3d/src/main.rs`): removed the stale "acquire-xor-revert … a competing-rate model would give a sub-cap equilibrium; deferred" block. It directly contradicted the `persister::step` call right below it (the #262 competing-rate model, shipped v0.22.0) and that call's own accurate comment. Kept the still-valid `rsl3_drug_avail` invariant note.
- **spheroid MUFA** (`sim-tme-3d/src/main.rs`): the "MUFA is an initial condition, not durable to late steps … a per-cell `scd_mufa_max` is deferred to calibration" comment was wrong post-#270. `apply_radial_cells_3d` (called just above) sets the per-cell `cell.mufa_cap` and `update_mufa_protection` saturates toward it, so the radial MUFA spread **is durable now**. Rewrote to describe the shipped path.
- **stromal** (`ferroptosis-core/src/stromal.rs`): the module doc framed lifting `StromalConfig` into the library as "explicit follow-up work" — but #220/#224 did exactly that. It now lives in `params::StromalConfig` and both the 2D and 3D consumers import it. Updated to point at the real single source of truth.
- **CALIBRATION_STATUS.md**: added the missing **fractal vessel topology** row (#292). The doc claims to cover "every layer," but the fractal branch constants (`BASE_ANGLE`/`LENGTH_RATIO`/`DEAD_END_PROB`/`n_trunks`) were absent — and the only #268 row said "exact, no parameters," implying the opposite of what the fractal addition introduced. Now an honest *Uncalibrated (illustrative); structure literature-anchored* row.
- **Cosmetic**: manuscript page count `~117` → `~115` (CLAUDE.md, README; matches the compiled PDF and CLAUDE.md's own two other mentions); FIGURES.yaml header comments `(1-23)` → `(1-24)`.

## Verification

- `cargo fmt --all --check` clean; workspace **243 + 48 + 4 doc-tests** green.
- Full Python suite **107/107**.
- Full **60³×180 production matrix** `summary.json` matches the pinned SHA (comment-only ⇒ byte-identical).
- `git diff` confirms zero non-comment Rust lines changed.

## Context

From the regroup gap sweep. Companion bookkeeping this cycle: closed #285 (complete), filed #295 (test-coverage gaps), folded the `lp_at_death` rename into #289. This PR clears the stale-comment debt + the calibration-doc omission.


ELares:
## Adversarial review — #296 (honesty / doc-hygiene pass)

For an honesty PR the failure mode that matters is a "fix" that introduces a *new* false claim, so I verified each rewritten comment and the new CALIBRATION row line-by-line against the code (3-lens sweep, all grounded in source). **Verdict: clean. Every new claim is true, nothing still-accurate was lost, no regressions.** One worth-acting-on scope finding; the rest are nits.

### Verified accurate (the whole point of the PR)
- **Spheroid-MUFA comment**: confirmed `apply_radial_cells_3d` runs *above* this block (main.rs:541 vs 788), sets per-cell `cell.mufa_cap` (spheroid.rs:192), and `update_mufa_protection` saturates toward `cell.mufa_cap.unwrap_or(params.scd_mufa_max)` (biochem.rs:131/228/341). "0.25 is only the fallback" is exact (Params::spheroid sets it). The new "durable" claim is true.
- **Persister comment**: the removed either-or/"deferred" block described behavior that no longer exists — `persister::step` (main.rs:1142) is the #262 competing-rate model (acquisition + reversion each step → sub-cap equilibrium), and the old *monotonicity* property is now false (tests confirm reversion is always live). So the deletion removed a now-false claim, not a true invariant; the still-valid `rsl3_drug_avail` note was kept.
- **Stromal module-doc**: `StromalConfig` really lives in `params.rs:375` with the stated Default values (0.06/18.0/0.003/0.25, unit-tested), and both sim-tme (import :45, use :1124) and sim-tme-3d (import :70, use :679) construct it from the library — no local copy. The now-invalid `sim-tme/src/main.rs:406` pointer was correctly dropped for a symbol link.
- **CALIBRATION_STATUS fractal row**: all six constants match `vasculature.rs` exactly (BASE_ANGLE 0.61≈35°, LENGTH_RATIO 0.80, MIN_LEN 3.0, MAX_DEPTH 14, DEAD_END_PROB 0.12, n_trunks=(target/30).round().clamp(2,64)); PMID 10919633 + D≈1.89 match the code citation; "higher hypoxic fraction at matched point-count" is backed by the count-controlled test (it asserts count-parity *then* the hypoxia inequality); "off-by-default Random byte-identical" is true.
- **Cosmetics**: `pdfinfo` confirms the PDF is genuinely 115 pages (117 was stale; all mentions now consistent); FIGURES.yaml max `manuscript_figure` is 24 so "(1-24)" is correct. No em/en dashes added (scanned every `+` line). Byte-identity holds without needing the SHA: the Rust diff has **zero non-comment lines** and both crates compile.

### Substantive (minor) — the PR's own premise leaves a same-file drifted reference
The stromal fix correctly replaced a hardcoded `sim-tme/src/main.rs:406` pointer (the struct moved, so it was *wrong*) with a symbol link. But the same fragile pattern survives, including **one in the very file this PR edited**:
- `stromal.rs:10` cites `sim-tme/src/main.rs:431-452` for "8-Moore neighbors" — that range is now ROS-multiplier / `lp_at_death` init code.
- also `oxygen.rs:377` (→:136-138, now a brace+separator), `immune_spatial.rs:122` (→:707, drifted ~2 lines), `sim-tme-3d/main.rs:685` (→:573, now DAMP-clearance decay).

These are weaker than the case the PR fixed (line-drift, not a vanished symbol), so not a blocker. But since the PR's premise is doc honesty and `stromal.rs` still carries a drifted line-ref two lines from the one it fixed, I'd either fix `stromal.rs:10` in this pass (convert to a symbol/function reference like the :406 fix) or file a small follow-up to de-hardcode all four `main.rs:LINE` pointers.

### Nits
- CALIBRATION row writes "n_trunks = points/30" where the code rounds before clamping — fair plain-language summary for a status table.
- The shortened persister comment points "below" (position-relative) rather than to a symbol — accurate today, mildly fragile.
- Pre-existing, not introduced here: `CLAUDE.md:57` still narrates a past edit as "0.2.0→0.25.0" while the crate is 0.26.0 — it's a historical description of that prior pass, untouched by this PR.

### Risk
Negligible. Comment/doc-only, byte-identical (verified), all triggered CI green (Rust test + fmt + CodeQL). I'd merge after fixing `stromal.rs:10` (same file, same premise) — or with a one-line follow-up issue for the four drifted line-refs.

PR #297, opened 2026-06-03
test(ferroptosis-core): cover wilson_ci, 2D immune cascade, persister-recovery (#295)

ELares:
## What

Adds unit tests for three pieces of live, shipped `ferroptosis-core` code that had **zero coverage** — items 1-3 of #295 (the test-coverage gaps surfaced by the regroup sweep). **Test-only** (`#[cfg(test)]` modules); no production code touched, so release builds and the sim-tme-3d byte-identity matrix are unaffected by construction.

## Coverage added (19 tests, lib 243 → 262)

**`stats::wilson_ci`** (7) — feeds `ci_low`/`ci_high` in sim-original, sim-window, sim-invivo, sim-tissue-pk, sim-combo-mech, sim-tumor-pk, and the Python bindings. A wrong z-constant or formula slip would silently skew *every* reported confidence interval. Pins:
- `n=0 → (0,0)`; `[0,1]` clamping at `k=0` and `k=n`;
- interval contains the point estimate;
- exact symmetry about 0.5 at `p=0.5`;
- monotonic lower bound in `k`; narrowing with `n`;
- closed-form known value (`n=100,k=50 → 0.4038/0.5962`, z=1.96).

**`immune`** (6) — the 2D ICD cascade driving sim-icd and sim-combo (its 3D successor `immune_spatial` already had 28 tests; the 2D original had none). Pins DAMP arithmetic (empty→0, total/per-cell), the **quality-over-quantity** activation invariant (few high-LP deaths out-activate many low-LP deaths — the SDT-vs-RSL3 asymmetry the model encodes), anti-PD-1 raising kills, and the cap at remaining alive tumor.

**`cell` persister-recovery** (6) — `recovered_persister_means` / `gen_recovered_persister` driving sim-window/sim-combo (distinct from `persister.rs`'s well-tested competing-rate step). Pins the day-0 persister baseline, convergence to normal targets, the exact half-time midpoint (gpx4 → 0.85 at `t_half`), monotonicity in time, and the generator's determinism + all `.max()` floors + no MUFA cap.

## Scope

Items 4 (wire the C-FFI `test_ffi.c` into CI) and 5 (generalize the figure data-drift fixture pattern) of #295 are deferred to follow-up PRs — they're CI-workflow / fixture work, separable from these pure unit tests.

## Verification

- `cargo fmt --all --check` clean; **full workspace 262 + 48 + 4 doc-tests green**.
- `git diff` is purely additive test modules (267 insertions, zero production-code lines).
- No byte-identity run needed: `#[cfg(test)]` code is excluded from release builds.

Closes items 1-3 of #295.

### DIFF
--- a/simulations/ferroptosis-core/src/cell.rs
+++ b/simulations/ferroptosis-core/src/cell.rs
@@ -183,3 +183,100 @@ pub fn gen_recovered_persister(days: f64, rates: &RecoveryRates, rng: &mut StdRn
         mufa_cap: None,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // The post-chemo persister-recovery means/cell generator drives sim-combo
+    // and sim-window's vulnerability-window result but had no coverage (#295).
+    // This is a separate formula from persister.rs's competing-rate step. Pin
+    // the exponential-recovery endpoints, the half-time midpoint, monotonicity,
+    // and the stochastic generator's determinism + floors.
+
+    #[test]
+    fn means_at_day_zero_are_persister_baseline() {
+        let (fsp1, gpx4, nrf2, gsh) = recovered_persister_means(0.0, &RecoveryRates::default());
+        assert!((fsp1 - 0.15).abs() < 1e-12, "fsp1={fsp1}");
+        assert!((gpx4 - 0.7).abs() < 1e-12, "gpx4={gpx4}");
+        assert!((nrf2 - 0.7).abs() < 1e-12, "nrf2={nrf2}");
+        assert!((gsh - 4.8).abs() < 1e-12, "gsh={gsh}");
+    }
+
+    #[test]
+    fn means_converge_to_normal_targets() {
+        let (fsp1, gpx4, nrf2, gsh) = recovered_persister_means(1.0e6, &RecoveryRates::default());
+        assert!((fsp1 - 1.0).abs() < 1e-6, "fsp1={fsp1}");
+        assert!((gpx4 - 1.0).abs() < 1e-6, "gpx4={gpx4}");
+        assert!((nrf2 - 1.0).abs() < 1e-6, "nrf2={nrf2}");
+        assert!((gsh - 5.0).abs() < 1e-6, "gsh={gsh}");
+    }
+
+    #[test]
+    fn gpx4_is_halfway_recovered_at_its_half_time() {
+        // At days == t_half the recovered fraction is exactly 0.5, so
+        // gpx4 = 0.7 + (1.0 - 0.7)·0.5 = 0.85, independent of the other rates.
+        let rates = RecoveryRates::default();
+        let (_, gpx4, _, _) = recovered_persister_means(rates.gpx4_half_recovery_days, &rates);
+        assert!(
+            (gpx4 - 0.85).abs() < 1e-9,
+            "gpx4 at t_half = {gpx4}, expected 0.85"
+        );
+    }
+
+    #[test]
+    fn means_are_monotonic_in_time() {
+        let r = RecoveryRates::default();
+        let (f0, g0, n0, s0) = recovered_persister_means(0.0, &r);
+        let (f1, g1, n1, s1) = recovered_persister_means(1.0, &r);
+        let (f10, g10, n10, s10) = recovered_persister_means(10.0, &r);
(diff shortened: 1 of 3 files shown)


ELares:
## Adversarial review — #297 (test-coverage for wilson_ci / 2D immune / persister-recovery)

For a *guard* PR the two failure modes that matter are (a) pinning a WRONG expected value (worse than no test) and (b) tests that pass tautologically and would never catch a regression. I checked both by independently recomputing every pinned value from first principles and by **mutating the source** to confirm each test actually bites (clean 262-test baseline, all mutations reverted). **Verdict: strong PR — no wrong-value pins, the stats and immune suites are genuinely protective. Two coverage gaps to flag.**

### Verified solid (mutation-tested)
- **`wilson_ci` pinned value is correct.** `wilson_ci(100,50)` independently recomputes to (0.403830, 0.596170) vs the asserted (0.4038, 0.5962); the hardcoded `z=1.96` vs scipy's exact 1.959964 differs by ~2e-6, well inside the 1e-3 tolerance. The symmetry test is algebraically exact (center = 0.5·(1+z²/n)/(1+z²/n) = 0.5 for any z, n), and the comment correctly scopes it as z-independent.
- **`wilson_ci` guards bite.** Mutations: `z 1.96→1.0` → `known_value` + `clamped` fail; swap `+s/-s` → 5 fail; remove the `n==0` guard (→ 0/0 = NaN) → `n_zero` fails; drop `.max(0.0)` → 2 fail. The 1e-3 tolerance pins z to ~[1.95, 1.965] (tight enough to reject 90%/99% z-values), not over-loose.
- **immune suite is precise.** `anti_pd1_increases_killing` is genuinely uncapped (raw 9.82/28.15 vs remaining 999,980 → tests the brake, not the cap); `immune_kills_capped` r2 genuinely exceeds the cap (raw 64.06 vs remaining 5 → the `.min` really fires); and `activation_tracks_per_cell_quality` catches the exact total-vs-per-cell bug it documents (rewiring to total DAMP flips few_high below many_low → fails). Each immune mutation is caught by exactly one intended test.
- **persister-recovery values + determinism correct.** day-0 baseline (0.15/0.7/0.7/4.8), convergence (1/1/1/5), and the gpx4=0.85 half-time midpoint all recompute bit-for-bit; the determinism test re-seeds two fresh `StdRng` and asserts exact f64 equality (real, not tautological). All thresholds match source exactly.

### Substantive — `gen_recovered_persister_respects_floors_and_caps`: 6 of 7 floor assertions are tautological
I deleted each `.max()` clamp in `gen_recovered_persister` (lines 176-182, not the `gen_cell` duplicates) and re-ran. **Only the `basal_ros` floor actually fires** (seed 163 drives a raw sample to 0.04758 < 0.05). Removing the `iron`/`gsh`/`gpx4`/`fsp1`/`lipid_unsat`/`nrf2` clamps all leave the test **green** — at `days=1.0` those means sit 3+σ above their floors, so no seed in 0..200 reaches them. A regression dropping any of those six clamps would ship undetected. (Thresholds themselves all match source, so this is a coverage gap, not a harmful pin.)
- **Suggested fix:** add `days=0.0` to the sweep (brings `fsp1` mean 0.15 toward its 0.01 floor) **and** add the higher-value guard this test is really reaching for — assert the generator's *sample means* track `recovered_persister_means(days)` over the seed sweep (ties `gen_recovered_persister` to the means function; catches a wrong-field/ignored-means regression, which the current test does not). Keep the `basal_ros` floor check (it bites).

### Minor — `wilson_ci` upper clamp `.min(1.0)` is uncaught
`interval_is_clamped_to_unit` probes the upper clamp at `k=n`, but there the unclamped `c+s` is algebraically exactly 1.0, so `.min(1.0)` is a no-op and the test passes with the clamp removed (verified: dropping `.min(1.0)` → all 7 green). The lower-clamp catch also depends on `n=10` producing negative fp-noise (at n=100/1000 it's exactly +0.0). Defensive code with thin coverage; optional to harden.

### Nit
`cell.rs` has duplicate floor literals in `gen_cell` (Persister arm) and `gen_recovered_persister` — any future floor test must target the right lines. Not actionable here.

### Risk
Negligible — test-only (`#[cfg(test)]`, excluded from release ⇒ byte-identity intact by construction), fmt clean, 262 + 48 + 4 green, diff purely additive. Worth the floor-coverage fix before merge; the rest is solid and the high-value targets (the CI-feeding `wilson_ci` and the untested immune cascade) are well guarded.

PR #298, opened 2026-06-03
ci: verify the C-FFI ABI (PhysiCell boundary) on every PR (#295 item 4)

### LINKED ISSUE #295: Test-coverage gaps: untested core formulas (wilson_ci, 2D immune, persister-recovery), unguarded C-FFI ABI, figure data-drift

ELares:
## What

Wires the `ferroptosis-ffi` C harness into CI so the repo's stated **PhysiCell integration boundary** is verified on every PR — #295 item 4.

## Why

`ferroptosis-ffi` exposes a `#[no_mangle] extern "C"` surface (`ferro_rng_new`/`free`, `ferro_params_default`/`invivo`, `ferro_gen_cell`, `ferro_sim_cell`) with a genuine integration test at `tests/test_ffi.c` — but that `.c` file was **build-by-hand only**: `cargo test --workspace` never links the C side, and no workflow referenced it. So the C ABI could break (or drift from the cbindgen header) and ship undetected.

## How

Adds one step to the Rust CI `test` job:
```
cargo build --release -p ferroptosis-ffi          # also re-runs build.rs/cbindgen
cc -o /tmp/test_ffi ferroptosis-ffi/tests/test_ffi.c -L target/release -lferroptosis_ffi -lm
LD_LIBRARY_PATH=target/release /tmp/test_ffi       # gates on exit code
```
- The harness returns non-zero on a NULL handle or an out-of-range persister kill rate (range [30, 55]%), so a real ABI break fails CI.
- The `cargo build -p ferroptosis-ffi` re-runs `build.rs` → cbindgen, so a **header/Rust ABI drift** now also fails here (previously only surfaced at manual build).
- `cc` ships with `ubuntu-latest`'s build-essential.

## Verification

- Ran the full flow locally (macOS equivalent: `.dylib` + `DYLD_LIBRARY_PATH`): build → cbindgen header regenerates in place with **no drift** → `cc` links → harness prints `All tests PASSED` → **exit 0**.
- Workflow YAML validated (parses; step present in the `test` job).
- **CI-only change** — no Rust/sim source touched, so byte-identity and `cargo fmt` are unaffected. The meaningful verification is this PR's own CI run (the new step executes the Linux equivalent).

Closes #295 item 4. (Item 5 — generalizing the figure data-drift fixture pattern — remains; items 1-3 landed in #297.)


ELares:
## Adversarial review — #298 (wire the C-FFI ABI harness into CI)

Verified both of the PR's empirical claims by mutate-rebuild-revert experiments locally (rustup 1.96.0; macOS `.dylib`/`DYLD_LIBRARY_PATH`, which shares CI's exit-code gating semantics; tree restored clean). **The gate is real and the step genuinely closes the gap. One substantive accuracy issue: the "header drift fails CI" claim is overstated.** No blockers.

### Verified — the gate actually fails CI on a broken ABI (good)
GitHub Actions runs a bare `run:` block under `bash --noprofile --norc -e -o pipefail`, so any non-zero line aborts the step. Confirmed by forcing each failure mode:
- **Harness assertion fails → step exits 1**: tightened the kill-rate window so the real 39.7% is rejected → `FAIL: Death rate 39.7% outside expected range` → exit 1.
- **`cc` link failure → exit 1**: wrong `-l` (`library not found`) and a bogus extern symbol (`symbol(s) not found`) both propagate.
- **Symbol/field rename that breaks the harness's C usage → exit 1**: renamed a `FerroResult` field → cbindgen regenerated the header → the unchanged harness failed to compile (`no member named 'final_lp'`).
- **Happy path on the clean tree → exit 0** (`All tests PASSED.`).

This is a genuine improvement: `cargo test --workspace` builds the cdylib but never *links it from a C translation unit*, so struct-layout / calling-convention / symbol-name mismatches at the real boundary were unexercised. Now they're linked, run, and gated.

### Substantive — the "a header/Rust ABI drift now also fails here" claim is overstated
`build.rs` writes cbindgen output **directly into the git-tracked** `ferroptosis-ffi/include/ferroptosis.h` and the build **passes** while silently overwriting it (no `cargo:rerun-if-changed`, so it regenerates every build). Empirically: adding a `#[no_mangle] ferro_abi_drift_probe` → `cargo build` exit **0**, header silently changed, harness still compiles against the fresh header → **CI would pass with a stale committed header**. So the step actually *masks* committed-header drift rather than catching it. What it really catches is narrower: cbindgen *generation failure*, or an ABI change that breaks a symbol/field `test_ffi.c`'s **source** references (a `cc` error) — not "a stale committed header."
- **Ask:** add `git diff --exit-code -- ferroptosis-ffi/include/ferroptosis.h` right after the build so a regenerated-but-uncommitted header fails CI (makes the claim true), and scope the wording in the workflow comment + CLAUDE.md + PR body accordingly ("verifies the C ABI links and runs; a drift that breaks the C consumer fails the compile").

### Minor / nits
- **Cross-platform flakiness: low risk.** RNG is `StdRng` (ChaCha12, bit-identical integer stream) with a fixed seed, so 39.7% is deterministic per platform; only libm transcendental sub-ULP noise could shift it, and the margin to the [30,55] bounds is ~9.7 pp (~6σ). Comfortable; widen the band only if it ever drifts on ubuntu.
- **`-lm` is unnecessary** — `test_ffi.c` includes `math.h` but uses no libm symbols (clangd flags the unused include). Harmless; drop `-lm` or leave it.
- **`/tmp/test_ffi` hardcode** — fine on the ephemeral runner; `$RUNNER_TEMP` would be tidier and matrix-safe.
- **Single-job signal** — the ABI check shares the `test` job with `cargo test --workspace`, so a red `test` doesn't distinguish a unit-test failure from an FFI-boundary failure. A separate job (like `fmt`) reads cleaner, at the cost of a second build. Optional.
- **`--release` after the debug `cargo test`** is correct (separate target dirs, no cache clobber; release is what PhysiCell would link). The extra ferroptosis-core release recompile is acceptable cost.

### Risk
Negligible — CI/doc-only, no Rust/sim source touched (byte-identity + fmt unaffected), and the PR's own `test` check ran the new step on Linux and passed. I'd merge after adding the `git diff --exit-code` header guard + scoping the drift wording; the rest are optional polish.

PR #291, opened 2026-06-02
fix(spheroid): volume-based radial phenotype zones, grounded in real spheroid structure (#270)

ELares:
Addresses the **zone-geometry** item of #270 (the histology calibration of gradient *strengths* stays data-blocked). Fixes a known correctness bug the gap-analysis surfaced.

## The bug
The spheroid phenotype assignment used **radial-fraction** thresholds (`glycolytic_frac=0.66`, `oxphos_frac=0.33`). By volume (∝ r³) this gave an **inverted** distribution: persister core ≈ **4%** of volume, glycolytic rim ≈ **71%** — the opposite of real spheroids (thin proliferating rim, large hypoxic/quiescent core). The prior code comment documented this as a known caveat.

## The fix
- **`radial_phenotype` now compares cumulative VOLUME fraction (`frac³`)** against the thresholds, so they specify zone *volumes*. Only the zone *assignment* changes — the MUFA/GSH/iron gradients stay radial (diffusion is radial; that's correct).
- **Defaults regrounded in Browning et al. (eLife 2021, DOI 10.7554/eLife.73020)** — verified — whose measured limiting structure is a necrotic core at ~73% of the outer radius and a proliferative boundary at ~90% (~35 µm rim). Converting to volume fractions: core = 0.73³ ≈ **0.39** (`oxphos_frac`), rim begins at 0.90³ ≈ **0.73** (`glycolytic_frac`). Core is now **~39% of volume (was ~4%)**, rim ~27%.
- Config/`literature()` docs rewritten (inversion caveat → resolved, literature-grounded); `CALIBRATION_STATUS.md` + `CLAUDE.md` updated (geometry now grounded; gradient strengths still placeholders).

## Tests
- `phenotype_zones_are_volume_based_rim_to_core`: asserts the volume-based assignment, the cube-root radial boundaries (~0.73R core, ~0.90R rim), and a substantial-core sanity check.
- `radial_spheroid_changes_kills_vs_random` updated: the volume-correct (larger, more SDT-resistant) hypoxic core now kills **~9.7% fewer** than the random baseline (radial≈4601 vs random≈5093) — asserted **directionally** + a 5% material floor (was a 10% |Δ| threshold tuned to the buggy config; the sim is deterministic so the value is stable).

## Safety
- **Off by default** (spheroid is an opt-in Override), so the default matrix is unaffected — production `summary.json` SHA **unchanged** (`1cf88663…390ff71`, re-verified at full 60³×180 scale).
[...]

PR #300, opened 2026-06-03
feat(contact): cell-cell contact-mediated ferroptosis resistance (#270 item 3)

ELares:
## What

Adds a genuinely new resistance axis the model lacked: **dense, highly-contacting tumor cells resist ferroptosis.** E-cadherin junctions activate Merlin/NF2 → Hippo → YAP inhibition, suppressing the YAP targets **ACSL4** (PUFA incorporation) and **TFRC** (iron import) that fuel lipid peroxidation; sparse/surface cells lose this brake and stay sensitive.

Foundational mechanism: **Wu et al., "Intercellular interaction dictates cancer cell ferroptosis via NF2-YAP signalling", Nature 2019, PMID 31341276** (verified against PubMed before citing).

## How

New `ferroptosis-core::contact` module:
- `contact_fraction_3d` = (tumor 26-Moore neighbours) / 26 ∈ [0,1] — interior cells ≈ 1.0, surface cells lower.
- `apply_contact_resistance_3d` scales each tumor cell's **durable** `lipid_unsat` (PUFA / ACSL4 axis) and `iron` (TFRC axis) down by `1 − strength·contact`. Effectively a per-cell ferroptosis-threshold modulation; stronger in 3D (up to 26 neighbours vs 8 in 2D). **Geometric — no RNG**, so it never perturbs grid generation; composes multiplicatively with the clonal lipid/iron axes.
- Wired into `sim-tme-3d` behind `Overrides.contact` (applied after the clonal perturbation). Off-by-default identity ⇒ default matrix byte-identical.

## Verification

- **Byte-identity**: full 60³×180 production `summary.json` matches the pinned SHA (contact off-by-default).
- **Tests**: 3 library (identity no-op; interior > surface contact; monotonic lipid/iron reduction, never increasing, non-tumor untouched) + 1 sim-tme-3d integration — `contact_resistance_reduces_rsl3_kills` (RSL3 kills **71→1** at literature strengths, deterministic). Empirically the layer also cuts SDT kills ~18% (less PUFA substrate).
- `cargo fmt` clean; workspace **268 + 49 + 4 doc-tests** green; full Python suite **116**; manuscript inventory passes; `v1.tex` compiles.

## Honesty / scope

- `ContactConfig::literature()` strengths (lipid 0.4 / iron 0.2) are **uncalibrated placeholders** encoding the documented direction (contact ⇒ resistance), not fit to data — logged in `CALIBRATION_STATUS.md` as *uncalibrated (illustrative); mechanism literature-anchored*.
- Bumps `ferroptosis-core` 0.26.0 → 0.27.0 (new module); docs updated (CLAUDE.md, core README, Appendix A + regenerated `v1.tex`).
- **No `--snapshot=contact` preset yet** (deferred, noted in docs).

**Part of #270** (item 3, contact signaling). Items 1-2 (per-cell MUFA durability, zone geometry) shipped in #291; the nutrient/waste-gradient half of item 3 and the zone-strength calibration remain, so #270 stays open.


ELares:
## Adversarial review — #300 (cell-cell contact resistance, #270 item 3)

Verified the two real risks with grounded experiments (strength sweeps, identity-config probes, field-read tracing; tree restored clean). **APPROVE — the code is correct and the byte-identity story is airtight; the direction (dense contact ⇒ resistance) is the robust result. Three non-blocking findings worth addressing, all about honesty of the magnitude/axis framing and one latent geometry trap.**

### Verified solid
- **Byte-identity is triple-gated and airtight**: `Overrides::default()` ⇒ `contact: None`; the apply site is `if let Some(cfg)` so None skips; and even `Some(default())` early-returns via `is_identity()`. Empirically, `Some(ContactConfig::default())` yields the exact baseline (72 == 72) at the sim level.
- **The load-bearing correctness question resolves cleanly**: `iron`/`lipid_unsat` are **static-only `Cell` fields with no `CellState` shadow** (`CellState` has no iron/lipid_unsat), and `sim_cell_step` reads both fresh every step (`effective_iron = cell.iron + extra_iron`; `effective_unsat = cell.lipid_unsat*(1-mufa_protection)`). So scaling `cell.*` alone is correct **and complete** — unlike clonal's `gpx4_mul`, which must also scale `state.gpx4` because gpx4 is dynamic. The post-init scaling takes full, durable effect (RSL3 72→0 confirms it propagates the whole run).
- **Composition with clonal is order-independent** (both pure `*=` on the same two fields); contact runs after clonal by design; RNG-free, so it's correctly exempt from the #278 seed-distinctness guard. Biology mapping (lipid↔ACSL4, iron↔TFRC) and the honest "uncalibrated; mechanism literature-anchored" CALIBRATION row are faithful.

### Substantive — the 98% magnitude is threshold-proximity-driven, and the effect is almost all PUFA (iron axis nearly inert)
A strength sweep (grid 24, 80 steps, RSL3 baseline 72) shows the kill reduction is steep and nonlinear: lipid 0.05 → ~56% of baseline, 0.1 → 29%, 0.2 → 5.6%, **0.4 (literature) → 0**. Even a 5% PUFA cut nearly halves kills — because `lipid_unsat` enters *both* the LP seeding term and the autocatalytic propagation term, so it has a super-linear effect on crossing `death_threshold`. The ~72 baseline-killed cells sit just above threshold, so a modest substrate cut tips most below it. **So the 98% is an artifact of where dose/grid/step sit relative to threshold, not a calibrated effect — the direction is the result.** Separately, attributing by axis: lipid-only(0.4)→0 dead, **iron-only(0.2)→68 dead (only 5.6%)** — the TFRC/iron axis is nearly inert for RSL3 (a 20% iron cut is a small perturbation to total ROS), so the module's "two co-equal YAP-target axes" framing overstates iron.
- **Ask:** add a one-line note (module doc + CALIBRATION row) that the magnitude is threshold-proximity-sensitive (only the direction is asserted) and that the iron/TFRC axis is a minor contributor under RSL3 (would matter more in Fenton-dominated/iron-loaded conditions). The PR is already honest that magnitudes are uncalibrated; this just closes the lipid-vs-iron asymmetry gap.

### Minor — fixed-26 denominator mis-scores a slab's outer shell; slab+contact is a reachable (unguarded) override pair
For the default **centred spheroid** the fixed-26 denominator is correct (verified: **0 tumor cells touch the box boundary**, so low-contact == true spheroid surface — exactly the intended biology; distribution: median 1.0, mean 0.94, 81% of tumor cells ≥0.9). **But** `generate_slab` makes an all-tumor block filling the box to its boundary, and `slab + contact` is *not* mutually-exclusive (only `slab + spheroid` is guarded). A slab+contact run would score the slab's entire outer face/edge/corner shell as low-contact==sensitive, when those are interior tissue cells whose real neighbours lie outside the simulated domain — the fixed-26 denominator silently treats the domain boundary as tissue surface. Latent (off-by-default, untested), but both layers are user-selectable.
- **Ask:** guard `slab + contact` with a `debug_assert!` like the existing `slab + spheroid` guard (cleanest), or document the spheroid-only assumption on `apply_contact_resistance_3d`.

### Minor — commit/doc says "71→1" but this build measures "72→0"
Illustrative and slightly stale (the test only asserts `< 0.5×baseline`, so it's robust). Soften the exact "71→1" figure in any doc/commit text to "far fewer / well under half," consistent with the uncalibrated framing.

### Risk
Low. Off-by-default, byte-identical (verified), no default-path change, correct composition. I'd merge after the slab+contact guard + the magnitude/iron-asymmetry doc note; the "71→1" wording is cosmetic.

PR #309, opened 2026-06-04
fix(contact): re-apply contact resistance to clonally-repopulated cells + close test gaps (#302)

### LINKED ISSUE #302: clonal repopulation doesn't re-apply contact resistance (clonal+contact incoherent/unguarded) + contact composition/guard/io test gaps
Surfaced by the regroup gap-analysis sweep (post-#300). A latent cross-layer correctness inconsistency in the new `contact` layer, plus the related test-coverage gaps.

## Correctness — clonal mid-run repopulation doesn't re-apply contact resistance

`apply_contact_resistance_3d` runs **once at setup** (`sim-tme-3d/src/main.rs:852`). `repopulate_dead_sites_3d` (#266 clonal spatial expansion) runs **every step inside the loop** (`main.rs:1353`) and builds a fresh `GridCell` for a revived dead site (`gen_cell` + `CellState::from_cell` + the clonal `SubclonePerturbation`, `clonal.rs:305-320`) but **never re-applies the contact scaling**. So with `clonal(with_repopulation)` + `contact` both enabled, a revived dense-interior site gets full, non-contact-reduced `cell.lipid_unsat`/`cell.iron` → it is *more* ferroptosis-sensitive than its contact-reduced neighbours, the **opposite** of the modeled biology (dense interior cells resist).

`clonal + contact` is a reachable override pair with **no guard** (only `slab + contact` is guarded, `main.rs:488`). Off-by-default ⇒ not a default-matrix bug, but a real composition inconsistency. Fix options:
- (a) re-apply contact resistance to the revived cell inside `repopulate_dead_sites_3d` (most faithful, but contact is grid-geometric — would need the donor/site contact fraction);
- (b) guard `clonal(with_repopulation) + contact` mutually exclusive (cheapest, mirrors the slab+contact `debug_assert`);
- (c) document the limitation in the contact/clonal module docs.

## Related test-coverage gaps
- **No `contact + spheroid` composition test** — the combination #300's docs call "the intended, correct" one (the fixed-26 contact denominator is correct for a centred spheroid). The #278 composition tests (`main.rs:4619/4648/4726`) omit contact entirely.
- **The `slab+contact` and `slab+spheroid` mutual-exclusion `debug_assert`s have no `#[should_panic]` test.** CI runs `cargo test --workspace` in debug (asserts live), so a `#[should_panic(expected="mutually exclusive")]` per guard would lock the invariant cheaply; without it a future refactor dropping a guard passes CI.
- **`io.rs` has zero test coverage** (4 CSV-writer pub fns: `write_heatmap_csv`/`write_depth_curves_csv`/`write_window_csv`/`write_json`) — the only 0-test ferroptosis-core module. These feed the FIGURES.yaml-traced figure generators; a header/precision (`{:.6}`/`{:.1}`) drift would silently corrupt a downstream figure with nothing catching it.

## Discipline
Pure ferroptosis-core/sim-tme-3d + tests + a guard; off-by-default ⇒ byte-identical. Source: regroup gap sweep.

ELares:
## What

Closes #302. Fixes a latent cross-layer correctness inconsistency in the `contact` layer (#270) surfaced by the post-#300 regroup sweep, and closes the related test-coverage gaps.

## The bug

`apply_contact_resistance_3d` runs **once at setup**. Clonal spatial expansion (#266, `repopulate_dead_sites_3d`) runs **every step** and rebuilds a revived dead site as a fresh, full-strength cell, but never re-applied contact scaling. So with `clonal(with_repopulation)` + `contact` both enabled, a revived dense-interior site came back with full `lipid_unsat`/`iron` and was therefore *more* ferroptosis-sensitive than its contact-reduced neighbours, the **opposite** of the modeled biology. (Off-by-default, so never a default-matrix bug.)

## Fix (option a, the faithful one)

Contact fractions are purely geometric (`is_tumor`-based, which neither death nor repopulation changes), so a revived site's correct reduction equals its setup-time reduction. I kept the layers composable rather than guarding the pair:

- **contact**: new `apply_contact_resistance_at_3d(grid, idx, cfg)` (per-cell); the bulk pass is factored through it.
- **clonal**: `repopulate_dead_sites_3d` now returns the **revived-site indices** (was a count; `.len()` recovers it). Not exposed via Python/FFI, so no downstream break.
- **sim-tme-3d**: after repopulation, re-apply contact to those indices when the layer is on (matching the setup order: clonal perturbation then contact).

## Test-coverage gaps closed

- **contact + spheroid** composition test (the pairing #300's docs call "the intended, correct" one).
- **`#[should_panic]`** tests locking the `slab+contact` and `slab+spheroid` mutual-exclusion `debug_assert`s, gated to debug builds where the asserts live.
- **First `io`-module tests** (the only 0-test ferroptosis-core module): header rows + `{:.1}`/`{:.6}` precision for the three CSV writers + the JSON writer. These feed the FIGURES.yaml figure generators, so a silent format drift would have corrupted a figure with nothing catching it.
- **clonal coherence test**: a revived dense cell is contact-reduced by exactly the geometric amount.

## Also

- Adds the `--snapshot=contact` overlay. `contact` was the only realism layer without one (the gap my sweep flagged); it shows the radial survival gradient on a centred sphere under RSL3.
- Bumps `ferroptosis-core` 0.27.0 → 0.28.0 (new pub fn + the `repopulate` return-type change), synced in Cargo.toml/lock + v1.md Appendix A.

## Verification

- `cargo test -p ferroptosis-core` → **272** + 4 doctests; `cargo test -p sim-tme-3d` → **52** (incl. the new compose + 2 guard tests).
- **Byte-identity preserved**: `constant_path_golden_kill_counts`, `dosed_path_golden_kill_counts`, and `clonal_k1_identity_is_byte_identical` all pass. The default 24-condition matrix is unchanged.
- `cargo fmt --all --check` clean; **116** Python tests pass; manuscript inventory (version pin) green.

Closes #302


ELares:
## Adversarial self-review (PR #309)

10 files, +408/-30. Fixes the clonal+contact incoherence (#302), closes three test-coverage gaps, adds the `--snapshot=contact` overlay, bumps `ferroptosis-core` 0.27.0 → 0.28.0. I re-derived the correctness argument and checked completeness against the code rather than trusting the description.

### Correctness — verified
- **Non-compounding / order-independent.** `repopulate_dead_sites_3d` overwrites a revived site with a fresh `GridCell` (full-strength), then the re-apply reduces it exactly once. `apply_contact_resistance_at_3d` reads only `is_tumor` (never neighbours' lipid/iron) and writes only the target cell, so re-applying across the revived set is order-independent. Empty revived set → no-op loop.
- **Faithful to setup.** Setup order is clonal-perturbation then contact; `repopulate` applies the clonal perturbation and the re-apply runs after it, matching. Contact fractions are `is_tumor`-geometric (death/repopulation never change the tumor-position map), so a revived interior site gets exactly the reduction it had at setup. The coherence test pins this to `base * (1 - lipid_strength*frac)` within 1e-12.
- **Completeness.** The only mid-run `gen_cell` (fresh full-strength `Cell`) is in `repopulate`. The other state builders in the step path (`CellState::from_cell_with_ros*`) rebuild only the dynamic `CellState` from an existing `cell`, leaving the static `Cell.lipid_unsat`/`iron` (where contact lives) intact. So the re-apply covers every place a fresh cell appears mid-run — no hole.
- **Byte-identity preserved.** Both layers are off-by-default; `constant_path_golden_kill_counts`, `dosed_path_golden_kill_counts`, and `clonal_k1_identity_is_byte_identical` pass. The `apply_contact_resistance_3d` refactor (bulk pass now calls the per-cell fn) is behaviour-identical to the prior inline loop.
- **Version bump complete.** Only remaining `0.27.0` is the legitimate v0.27.0/#270 history entry in CLAUDE.md; v1.md Appendix A and Cargo.toml/lock are 0.28.0; ferroptosis-python/ffi use `path` deps and auto-track. `repopulate`'s return-type change is not exposed via Python/FFI.

### Notes (nuances, not defects)
1. **Contact fraction ignores neighbour death** — a revived cell whose neighbours have since died still scores high contact (they remain `is_tumor`). This is faithful to the #270 model's own definition (geometric, set once, death-independent), so #302 introduces no new inconsistency; it is a known simplification of the uncalibrated layer, worth a calibration-era revisit (live-contact vs position-contact), not a fix here.
2. **`--snapshot=contact` full render is not CI-tested** — consistent with every preset (the ~333 MB production snapshot is local-only). The preset→`Overrides` wiring is one compile-checked line, and the underlying run path is covered by `contact_composes_with_spheroid` + `contact_resistance_reduces_rsl3_kills`. A cheap `resolve_snapshot("contact")` assert (`preset.contact == true`) would lock the wiring, but no other preset has one, so I left it out to avoid asymmetry. Flagging in case you'd prefer it added.
3. **io tests pin exact CSV strings** — intentional drift detection. `str::lines()` strips a trailing `\r`, so the asserts are robust to the csv crate's terminator across platforms (csv pinned `1.3`, resolves 1.4.0). The pre-existing `io::Error::new(ErrorKind::Other, _)` on `write_json` draws a clippy `io_other_error` suggestion, but it predates this PR and clippy is not a CI gate.

### Design choice worth confirming
I took **option (a)** from the issue (re-apply, keeping clonal+contact composable) over **(b)** (guard them mutually exclusive). Rationale: the suite's premise (#278) is that realism layers compose, the contact fraction is geometric so the re-application is exact, and resistant-subclone expansion inside a contact-resistant tumor is a meaningful combination to keep available. The cost is ~3 extra lines at the call site. If you'd rather forbid the pair for simplicity, (b) is a one-line `debug_assert` swap.

### Verdict
Correct, complete, and byte-identical, with solid test coverage (the new coherence + composition + guard + io tests). No blocking issues. The only optional add is the `resolve_snapshot` wiring smoke test (note 2).

Issue #377, opened 2026-06-05
Non-drug stress-niche persister entry trigger (hypoxia/nutrient) — follow-up to #342


ELares:
**Follow-up to #342.** The persister model's entry trigger is currently the **drug-driven acquisition** (`acquisition_rate` scaled by the per-step `drug_intensity`). The classic drug-tolerant-persister biology also has a **non-drug, stress-niche** entry route: hypoxic / nutrient-poor drug-sanctuary microenvironments drive cells into a slow-cycling, drug-tolerant persister state independent of (or in addition to) drug exposure (e.g. hypoxia-induced drug tolerance; quiescent perivascular/hypoxic niches).

**Why it needs its own PR.** Adding a stress entry trigger cleanly requires **decoupling the entry drive from the drug-resistance/locking drive** in the library: today `persister::step`/`step_with_locking` take a single `drug_intensity` that drives acquisition AND the GPX4 resistance AND the locking EMA. A hypoxia-driven entry must raise *acquisition* without inflating resistance or the locking EMA, so it needs a new function signature (e.g. a separate `entry_drive` parameter) or a small config-level entry term.

**Task.** Add a `PersisterConfig` stress-entry term (off-by-default `0.0` ⇒ byte-identical) that boosts persister entry in low-O2 / low-supply zones, wired in sim-tme-3d via the per-cell `o2_supply_for_exo` (or the vascular/slab supply) field. Decouple entry-drive from drug-intensity in the library.

**Acceptance.** A stress-entry term with documented (uncalibrated) direction; an A/B test showing elevated persister entry in hypoxic vs normoxic zones at fixed drug; off-by-default byte-identity preserved (production matrix SHA unchanged); CALIBRATION_STATUS update.

Issue #342, opened 2026-06-04
Persister entry triggers + reversible-to-irreversible (epigenetic locking) transition


ELares:
**Gap.** The persister model captures **recovery** after drug withdrawal (exponential half-lives) and a competing-rate acquire/revert step, but not (a) what **triggers entry** into the slow-cycling/quiescent persister state, nor (b) the empirical observation that beyond a threshold of sustained drug exposure (~10-14 days in several systems) persistence becomes **epigenetically locked / effectively irreversible**.

**Why it matters.** Treating all persistence as freely reversible affects predictions about dosing schedules (intermittent vs continuous) and the vulnerability window. FSP1/HDAC-mediated suppression of alternative defenses in persisters is now documented (Sci Adv 2026, PMID 41481741), supporting a durable, drug-induced persister state.

**Task.** Add (1) stress/drug-exposure entry triggers into the persister state, and (2) a reversible->irreversible transition gated on cumulative sustained exposure, so long continuous dosing produces a locked sub-population. Off-by-default/byte-identical.

**Acceptance criteria.** Entry + locking dynamics with documented (uncalibrated) thresholds; a test contrasting intermittent vs continuous dosing outcomes; CALIBRATION_STATUS row. Builds on the persister competing-rate work (#262, closed).

Source: Sci Adv 2026 PMID 41481741 (FSP1/HDAC persister-cell ferroptosis).

ELares:
**Progress (PR #367 merged):** the substantive **reversible-to-irreversible locking transition** is in (ferroptosis-core 0.40.0):

- `persister::step_with_locking` + `PersisterState {reversible, locked, cumulative_exposure}` + `PersisterConfig.lock_rate`/`lock_threshold`/`exposure_decay` (all off by default).
- A sustained-exposure EMA gates locking, so **continuous dosing locks an irreversible pool while intermittent dosing does not** (criterion 2 + the intermittent-vs-continuous test + CALIBRATION_STATUS row, all delivered).
- `lock_rate=0` reduces exactly to `step` (byte-identical); documented in the manuscript §8.4.

**Remaining (keeping this open):**
1. **Explicit entry triggers (criterion 1).** Entry is currently the pre-existing drug-driven acquisition (`acquire`/`step`, #241/#262). If a more explicit stress/threshold entry handle is wanted, it is not yet added.
[...]

PR #378, opened 2026-06-05
feat(#342): wire persister epigenetic locking into sim-tme-3d + spatial dosing demo

ELares:
## Summary

Wires the persister **epigenetic-locking** transition (library work in PR #367: `step_with_locking` + `PersisterState`) into the `sim-tme-3d` binary, and adds the spatial **intermittent-vs-continuous** demonstration. Persister cells were already wired (#241); this consumes the new locking dynamics.

### Changes
- **`CellState`** gains `persister_reversible` / `persister_locked` / `persister_cum_exposure` (all `0.0`-inert default, `#[serde(default)]`, consumer-owned exactly like `persister_fraction`; **not in the C ABI**, cbindgen header unchanged). The per-cell persister update calls `step_with_locking` and caches `total()` back into `persister_fraction`, so the GPX4/MUFA couplings + the mean/snapshot keep reading one scalar.
- **`persister_locked_mean`** summary metric reports the irreversible sub-pool (`None` when persister off; `Some(0.0)` when `lock_rate=0`).
- **`continuous_dosing_locks_persisters_but_intermittent_does_not`** spatial test: continuous dosing drives the per-cell exposure EMA above `lock_threshold` and locks a non-reverting fraction; the ~25%-duty intermittent schedule keeps the EMA below threshold and locks nothing; `lock_rate=0` locks nothing under any schedule.

### Byte-identity
- `step_with_locking` reduces **exactly** to `step` when `lock_rate=0`, and `total()` then equals the old reversible value, so `persister_fraction` tracks the pre-#342 sequence. The three new `CellState` fields are touched only inside the persister gate (the matrix has persister off), so they stay `0.0` there.
- **Full 60³×180 production `summary.json` SHA verified `OK`** against `expected_summary.sha256`. Golden 20³×80 + the existing persister regression tests (`persister_off_is_inert_and_unreported`, `persister_reduces_multidose_kills`, `persister_reverts_after_dosing_stops`) all pass.
- `cargo fmt` clean; `cargo test --workspace` green; FFI/C-ABI unchanged.

### Entry triggers (criterion 1)
The persister-**entry** trigger is the existing **drug-driven acquisition** (`acquisition_rate` scaled by the per-step `drug_intensity`), so entry is drug-exposure-gated, and it is now spatially exercised. The issue also mentions non-drug **stress** entry; a hypoxia/nutrient **stress-niche entry trigger** (the classic drug-sanctuary biology) is filed as a focused follow-up (#377) because it needs a small library API change to decouple the entry drive from the drug-resistance/locking drive.

### Docs / versioning
Crate `0.42.0 → 0.43.0` (additive `CellState` fields). Updated Appendix A version pin (guarded by `test_manuscript_version_matches_cargo`), `MODEL_CARD.md`, the `CALIBRATION_STATUS.md` persister-locking row, and `CLAUDE.md`.

This delivers the **sim-tme-3d wiring + spatial dosing-schedule demonstration** that #342 stayed open to track (PR #367 delivered the library locking). Marked `refs #342`; the remaining stress-niche entry trigger is the follow-up above.

refs #342


ELares:
## Deep review addressed (0 must-fix, 1 should-fix, APPROVE-WITH-NITS)

A four-dimension adversarial review (byte-identity/determinism, correctness/safety, biology/completeness, style/docs) verified the change empirically: the full 60³×180 production SHA passes, the `lock_rate=0 ⇒ step_with_locking ≡ step` reduction is provably exact, the rayon per-cell update is data-race-free, PMID 41481741 (Higuchi 2026, Sci Adv) is real and on-point, and the version bump + house style are clean. Fixed in the latest commit:

1. **should-fix.** The CALIBRATION_STATUS + CLAUDE.md prose attributed "survives washout" to the spatial sim test, but that test drives drug continuously to the end (no washout). Reworded so the sim test claims only what it shows (continuous locks a fraction, intermittent locks nothing) and washout-survival is attributed to the library unit test (which runs a full drug washout).
2. **nit.** Added `persister_locked_mean == None` assertions to `persister_off_is_inert_and_unreported` (pins the skip_serializing_if guard at PR-test speed).
3. **nit.** Decoupled the "~10-14 days" sustained-exposure timescale from PMID 41481741 in `params.rs` (the PMID sources the FSP1/HDAC mechanism, not the day figure; the timescale is illustrative and not mapped to the dimensionless EMA threshold).

Per the review, **#342 stays open with `refs`**: this delivers criterion 2 (locking) + the sim wiring + the demonstration + the drug-exposure entry route, but the non-drug stress-niche entry trigger (criterion 1's other route) is split to follow-up #377 (it needs a library API change to decouple entry-drive from drug-intensity). #342 closes when #377 lands.

Issue #336, opened 2026-06-04
Model oxygen-dependent SDT/sonosensitizer ROS yield (Type I vs Type II) and resolve the contested hypoxia-leg claim


ELares:
**The most contested claim.** Our hypoxia leg predicts SDT "maintains 87.8% kill" under hypoxia because exogenous ROS is treated as largely O2-independent. The current literature contradicts this for the clinically relevant agent.

**What the field shows.** The lead clinical SDT agent, **SONALA-001** (IV 5-ALA -> protoporphyrin IX, activated by MR-guided focused ultrasound), is a **Type II / oxygen-dependent** sensitizer (singlet-oxygen via energy transfer, mechanistically like PDT). First-in-human recurrent-high-grade-glioma results (*Science Translational Medicine* 2025, DOI 10.1126/scitranslmed.ads5813) report safety and only **modest cell death**. Type I (oxygen-independent, electron-transfer) sonosensitizers exist but are preclinical only. There is also an SDT **hypoxia paradox**: ROS generation consumes O2 and can worsen local hypoxia mid-treatment.

**Task.** Make the sonosensitizer ROS yield explicitly **oxygen-dependent** with a configurable Type I/Type II fraction (default to the clinically grounded Type II/O2-dependent case), so the hypoxia comparison reflects the agent actually in trials; optionally model O2 consumption during treatment. Re-run the hypoxia panel and update §7.1 accordingly.

**Acceptance criteria.** A `sdt_o2_dependence` parameter (Type I fraction) in the sonosensitizer/exo-ROS path; the default Type II case shows the SDT hypoxia advantage shrinking/closing; manuscript §7.1 + figure caption updated to match; off-by-default for the production matrix. This is the model-side counterpart to the manuscript caveat we already added.

Sources: SONALA-001 Sci Transl Med 2025 (10.1126/scitranslmed.ads5813); npj Acoustics 2026 preclinical-SDT review.

PR #379, opened 2026-06-05
feat(#358): O2-dependent SDT reachability + quantify the §7.1 hypoxia bound (both bounds)

### LINKED ISSUE #358: Follow-up to #336: expose O2-dependent SDT via a preset + re-compute the §7.1 hypoxia figure under the O2-dependent assumption
Follow-up to #336 (merged in #357), which landed the off-by-default O2-dependent SDT/PDT exo-ROS capability (`oxygen::o2_dependent_exo_factor` + sim-tme-3d `Overrides.sdt_o2_dependence`, byte-identical default) but left two pieces as deferred scope (flagged in the #357 review):

1. **Reachability.** The mode is currently exercisable only via the library function + the internal test (no CLI flag or `--snapshot=` preset). Add a `--snapshot=sdt-o2dep` preset (SDT on a hypoxic spheroid with `sdt_o2_dependence=1.0`) so the hypoxic-core survival front is visualizable, and/or a CLI flag. Note: this requires adding the field to all `SnapshotPreset` literals (off everywhere except the new preset), keeping the matrix byte-identical.

2. **Scientific payoff: re-compute the contested §7.1 hypoxia leg.** The manuscript §7.1 / Figure 24(a) numbers (SDT 91.9% -> 87.8% under hypoxia) still reflect the O2-INDEPENDENT upper bound. Re-run the hypoxia comparison with the O2-dependent (Type II) assumption to QUANTIFY how much the SDT hypoxia advantage shrinks (it likely shrinks substantially or closes), and report both bounds. This turns "the model assumes O2-independent" into "here is the gap under each assumption", which is the real contribution. The `dependence` value is itself uncalibrated (the actual Type I/Type II split is sonosensitizer-specific), so report a range.

Acceptance: a preset (or CLI) that runs the O2-dependent mode; a re-computed §7.1/Fig-24a comparison under O2-dependent SDT with both bounds reported; manuscript + CALIBRATION_STATUS updated with the quantified shrinkage.

ELares:
## Summary

Closes #358 (follow-up to #336). Delivers the two deferred pieces: **reachability** for the O2-dependent SDT mode, and the **scientific payoff** of quantifying the contested §7.1 hypoxia leg as both bounds.

### Reachability
`sim-tme` (the 2D binary that produces the §7.1 numbers) gains a `--sdt-o2-dependence <f>` CLI flag (default `0.0`). It applies `oxygen::o2_dependent_exo_factor(o2_supply, dep)` to the SDT/PDT `exo_ros_peak` in the gradient conditions, where `o2_supply = exp(-depth/λ)` is the same per-cell field that attenuates `basal_ros`. At `dep=0` the factor is exactly `1.0` and the sampled exo-ROS (and RNG stream) are untouched, so the committed `tme_summary.json` / Fig-21 fixture stay **byte-identical** (verified: default run reproduces 91.9% / 87.8%; the figure-data guard `test_fig21_fixture_matches_live` passes). The issue's deliverable 1 was "a preset **and/or a CLI flag**"; this is the CLI path, on the binary where §7.1 lives.

### Scientific payoff: §7.1 quantified as both bounds
Re-running the **identical** 2D hypoxia comparison at full O2-dependence (`--sdt-o2-dependence 1.0`, the Type II / SONALA-001 extreme):

| | O2-independent (upper bound, current §7.1) | O2-dependent (lower bound, dep=1.0) |
|---|---|---|
| Uniform SDT | 91.9% | 91.9% |
| Gradient SDT (overall, λ avg) | 87.8% | **~3.7%** (2.7% to 4.9% across λ=80 to 150 µm) |
| Hypoxic-zone SDT | 86.6% | **0%** |

So the SDT hypoxia advantage **disappears entirely** at the Type II extreme, collapsing to RSL3-like levels (RSL3 is 3.7% to 0.1%). The real value lies between the two bounds (the Type I/Type II split is sonosensitizer-specific and uncalibrated). §7.1 and the Fig-21 caption now report **both bounds**, turning "the model assumes O2-independence" into a quantified range. This is the manuscript's single most contested assumption, now bracketed.

### Tests / docs
- `o2_dependent_sdt_collapses_hypoxic_kill` (sim-tme's first unit test): guards the `dep=0` knob-off identity (with vs without a supply field) AND the `dep=1.0` collapse direction.
- §7.1 (`v1.md` + regenerated `v1.tex`), the Fig-21 caption, and the `CALIBRATION_STATUS.md` O2-dependent-SDT row updated with the quantified shrinkage.
- `cargo fmt` clean; sim-tme tests + figure-data guards + manuscript-inventory test green. The committed default output / fixture are unchanged (`output/tme` is gitignored; the default `cargo run -p sim-tme` regenerates the O2-independent data).
- **No ferroptosis-core change** (reuses the existing `oxygen` helper), so no crate version bump.

The optional 3D `--snapshot=sdt-o2dep` visualization preset (the issue's leading example) is a trivial mechanical follow-up; the O2-dependent mode is already reachable in 3D via `Overrides.sdt_o2_dependence` and in 2D via this CLI flag, which satisfies the acceptance criterion.

Closes #358


ELares:
## Deep review addressed (0 must-fix, 3 should-fix, APPROVE-WITH-NITS → close #358)

A four-dimension adversarial review verified byte-identity bit-for-bit (the default run reproduces SDT 91.9%/87.8% and RSL3 3.7%/0.1% at full f64 precision against the unchanged fixture), confirmed the both-bounds numbers are faithful to a real `--sdt-o2-dependence 1.0` run, and found no dash-style violations. Fixed in the latest commit:

1. **SF1 (consistency).** The knob was not threaded into the immune-on / anti-PD-1 / cycling gradient paths, so a flagged run produced a self-contradictory summary (immune-off gradient SDT collapsed while immune-on stayed at ~88%). Threaded `sdt_o2_dependence` through `run_spatial_with_immune` (3 gradient call sites) and `run_spatial_cycling`; now all gradient rows collapse consistently (verified: immune-on 4.1%, anti-PD-1 4.2%, cycling 4.8%, all hypoxic-zone 0%). Off-by-default factor=1.0 keeps the default byte-identical (figure guard passes).
2. **SF2 (overstatement).** Scoped "asymmetry disappears entirely" to the deep hypoxic zone (both ~0%) and noted the residual rim-driven overall gap (SDT ~3.7–4.0% vs RSL3 0.1%) in §7.1, the Fig-21 caption, v1.tex, and CALIBRATION_STATUS.
3. **SF3 (mixed conventions).** Report the λ=120 lower bound (4.0%) like-for-like with the λ=120 upper bound (87.8%), label 3.7% as the across-λ mean, and add the range to the Fig-21 caption.

Nits: reject non-finite `--sdt-o2-dependence` (NaN would survive `clamp`) with a warning; added the O2-dependent lower bound to the abstract.

Per the review, **#358 closes with this PR** (the CLI flag satisfies the "preset OR CLI" acceptance; the recomputed both-bounds §7.1 + CALIBRATION are delivered). The 3D `--snapshot=sdt-o2dep` visualization preset the issue led with is split to a tracked follow-up so it is not silently dropped.

Issue #365, opened 2026-06-05
Wire NCOA4-ferritinophagy hypoxia-iron coupling into sim-tme-3d + re-examine the §7.1 hypoxia leg (follow-up to #340)


ELares:
**Follow-up to #340 (closed).** #340 added the off-by-default library mechanism: `biochem::ferritinophagy_iron_factor` (NCOA4-ferritinophagy labile-iron release) and `oxygen::hypoxia_iron_factor(o2_supply, sensitivity)` (hypoxia raises iron import). Both are tested and byte-identical off-by-default, but `hypoxia_iron_factor` is currently a **library helper with no consumer** in any shipped binary.

## Gap

The scientifically important part, demonstrating that hypoxia can RAISE the Fenton/iron contribution to ferroptosis even as it LOWERS the O2-dependent SDT exo-ROS yield (#336), is not yet wired into the spatial model. Without that, the manuscript's contested §7.1 "hypoxia uniformly protects against RSL3" framing is only qualified by a doc note, not by a runnable spatial result.

## Proposed task

1. Add a sim-tme-3d `Overrides` knob (e.g. `ferritinophagy_release` + `hypoxia_iron_sensitivity`) and apply `oxygen::hypoxia_iron_factor(local_o2, sensitivity)` to the per-cell iron in the spatial paths (mirroring how #336/#357 wired `o2_dependent_exo_factor` for SDT exo-ROS). Off-by-default (sensitivity 0) keeps the production matrix byte-identical.
2. Add a `--snapshot=ferritinophagy` (or similar) overlay.
3. Run an A/B in hypoxic vs normoxic zones showing that under the dynamic-iron + hypoxia coupling, RSL3 kill in hypoxic zones can rise rather than fall, quantifying the §7.1 qualification.
4. Re-examine / re-compute the §7.1 hypoxia figure under the combined #336 (O2-dependent SDT) + #340 (hypoxia-iron) assumptions, and update the manuscript text if the combined picture shifts.

## Acceptance criteria

- `hypoxia_iron_factor` reachable from a sim-tme-3d override; off-by-default byte-identical.
- A spatial A/B result quantifying the hypoxia-iron effect on RSL3 kill.
- §7.1 manuscript text/figure re-examined under the combined hypoxia assumptions.

Related: #336 / #358 (O2-dependent SDT leg). Builds on ferroptosis-core 0.38.0 (`ferritinophagy_iron_factor`, `hypoxia_iron_factor`).

Issue #340, opened 2026-06-04
Dynamic labile iron pool + NCOA4 ferritinophagy (replace the static iron model)


ELares:
**Gap.** The model treats labile iron as a static per-cell value (modulated only by pH/contact). In reality the labile iron pool is dynamically regulated: TfR1/DMT1 import, ferritin storage, and **NCOA4-mediated ferritinophagy** (selective autophagy of ferritin) release iron to drive Fenton chemistry. Recent work shows ferritinophagy is itself kinase/E3-regulated (ATM phosphorylates NCOA4; DTX2 degrades it).

**Why it matters.** Autophagy/ferritinophagy is a ferroptosis **gate**: persisters with elevated autophagy can be sensitized or protected depending on iron handling; hypoxia shifts iron handling (TfR1 up) in ways that can *increase* ferroptosis vulnerability, which qualifies our "hypoxia uniformly protects RSL3" framing.

**Task.** Add a minimal dynamic-iron module (import / ferritin storage / NCOA4 ferritinophagy release) feeding the Fenton/LP term, off-by-default and byte-identical for the matrix, with a hypoxia coupling so the iron-scavenging response can be explored.

**Acceptance criteria.** A `ferritinophagy`/dynamic-iron path with documented (uncalibrated) parameters; a test showing iron release modulates ferroptosis; CALIBRATION_STATUS row + a note on the hypoxia-iron interplay.

PR #387, opened 2026-06-05
feat(#383): O2-dependent Fenton option — correct the §7.1 deep-core hypoxia-iron artifact

### LINKED ISSUE #383: O2-dependent Fenton rate (couple the H2O2 substrate to local O2) — surfaced by #365
**Surfaced by the #365 deep review.** The #365 wiring revealed that the model's **Fenton ROS term is oxygen-independent**: `fenton = (cell.iron + extra_iron) * ferritinophagy_iron_factor * params.fenton_rate * noise`, added to `total_ros = basal_ros + exo + fenton` (biochem.rs), with no O2 coupling. Hypoxia in the model scales only `basal_ros` down (the §7.1 caveat states this: "we modulate only basal_ros, not Fenton rate directly").

**Consequence (documented in §7.1).** When the #365 hypoxia-iron coupling raises labile iron most where O2 is lowest, the Fenton ROS rises in the deep anoxic core regardless of anoxia, so the model predicts RSL3 *rescues* the anoxic core (hypoxic-zone kill 0%→1.3% at λ=50, even more at λ=40). This is a **model artifact**: a real anoxic core would also lose the superoxide→SOD→H2O2 substrate that drives the Fenton reaction. The model therefore **overstates** RSL3 efficacy (and the hypoxia-iron rescue) in genuinely anoxic regions.

**Task.** Couple the Fenton rate (or the H2O2 substrate term) to local O2, off-by-default (a `fenton_o2_dependence` knob, identity at 0 ⇒ byte-identical), so that in the anoxic core the Fenton ROS is suppressed by the loss of the O2-derived H2O2 substrate. This would make the modeled deep-core protection real (and would interact correctly with the #340/#365 hypoxia-iron coupling: more iron but less H2O2 in hypoxia). Then re-examine the §7.1 deep-core claim under the O2-coupled Fenton.

**Acceptance.** An off-by-default O2-dependent Fenton term (byte-identical at 0); an A/B showing the anoxic core is protected once the Fenton H2O2 substrate is O2-coupled, even with hypoxia-iron on; §7.1 updated. Related: #340, #365, #336/#358.

ELares:
## #383: O2-dependent Fenton rate (couple the H2O2 substrate to local O2)

Closes #383. Corrects the deep-core **model artifact** the #365 review surfaced and §7.1 flagged.

### The artifact
The model's Fenton term is **O2-independent** (`fenton = iron × fenton_rate`, added to `total_ros` with no O2 coupling; the model scales only `basal_ros` in hypoxia). So when the #340/#365 hypoxia-iron coupling raises labile iron *most where O2 is lowest*, the model predicts RSL3 **rescues** the anoxic core (hypoxic-zone kill 0% → 1.3% at λ=50, more at λ=40). A real anoxic core would also lose the superoxide → SOD → H2O2 substrate the Fenton reaction consumes, so the model **overstates** deep-core RSL3 efficacy.

### The fix (off-by-default, byte-identical)
- **`oxygen::fenton_o2_factor(o2_supply, dependence) = (1−dep)+dep·o2_supply`** (clamped `[0,1]`, identity at `dependence=0`), mirroring `o2_dependent_exo_factor`'s linear form. Biology: the Haber-Weiss chain (superoxide → SOD → H2O2; **Kehrer 2000, PMID 10963860**) and the O2-requirement of ferroptotic lipid peroxidation (**Stockwell 2017, PMID 28985560**). Both PMIDs NCBI-verified.
- **sim-tme-3d `Overrides.fenton_o2_dependence`**: because `cell.iron` feeds **only** the Fenton term, the consumer applies the factor by scaling `cell.iron`, where it composes **multiplicatively** with the #365 `hypoxia_iron_factor`. Hypoxia raises the iron but gates the H2O2 substrate, so the **net** deep-core Fenton can fall — exactly what the issue asks ("more iron but less H2O2 in hypoxia").

### Validation
- **A/B test** `o2_dependent_fenton_protects_the_anoxic_core_under_hypoxia_iron` (same grid 60 / λ=50 / 180-step config as the #365 test): with hypoxia-iron **on** (the artifact regime, precondition asserted) **and** the Fenton fully O2-coupled (`dependence=1.0`), the deep hypoxic-zone kill drops by **more than half** back toward the un-rescued baseline. Byte-identity at `dependence=0` and determinism also asserted.
- **`oxygen::fenton_o2_factor`** unit test: identity at 0, gates to 0 at anoxia, monotone, clamps, and the **net sign-flip** (iron up × H2O2 gated ⇒ net loss).
- **Byte-identity**: full 60³×180 production matrix `summary.json` SHA unchanged (`sha256sum -c` OK); golden 20³×80 tests pass.
- **C ABI**: `ferroptosis.h` byte-identical after `build.rs` regen (not in the FFI).

### Docs
ferroptosis-core 0.43.0 → **0.44.0** (new public `fenton_o2_factor`). Updated §7.1 (the artifact is now *addressable*, not just flagged), CALIBRATION_STATUS (new row + cross-ref from the #340/#365 row), CLAUDE.md version history, README `oxygen` row (now lists all three O2-coupling helpers), Appendix A version pin, and v1.tex.


ELares:
## Review fixes applied (commit 05fd85c)

The adversarial multi-dimension review returned **0 blockers, 4 should-fix, 3 nits** (all 5 dimensions approved). All addressed:

**should-fix**
1. **`extra_iron` bypassed the O2 gate** (flagged by the math + completeness reviewers). The Fenton flux is `(cell.iron + extra_iron)·…`, and only the static `cell.iron` was gated at setup — the per-step neighbor-death-diffused `extra_iron` stayed an O2-independent Fenton source in the core. Now gated by the **same** local-O2 factor inside the per-step loop, so the **whole** Fenton substrate is O2-coupled. Behind `if fenton_o2_dependence > 0.0`, reading `o2_supply_for_exo[idx]` shared-immutable by own index ⇒ the documented rayon byte-identity invariant holds and the default path is untouched.
2. **Untested rim-vs-core differential.** The §7.1 / CALIBRATION_STATUS claim that the gate suppresses the core more than the rim is now asserted in the A/B test (fractional reduction in the hypoxic zone > in the normoxic zone).
3. **Stockwell 2017 over-attribution.** It was cited for a verbatim "requires molecular O2" the Primer abstract doesn't state. Softened everywhere to "ferroptotic LPO is an O2-consuming lipid-radical chain (reviewed in …)". The load-bearing mechanistic cite remains Kehrer 2000 (Haber-Weiss).

**nits**
4. New test: with hypoxia-iron **off**, the gate only reduces/holds kill and is identity at 0.
5. `MODEL_CARD.md` version 0.43.0 → 0.44.0.
6. v1.tex regenerated for the softened §7.1 prose.

**Re-verified:** fmt clean; golden 20³×80 byte-identity + both #383 tests pass; FFI header unchanged; and the **full 60³×180 production matrix `summary.json` SHA is unchanged** after the hot-loop edit (`sha256sum -c` OK).

Issue #332, opened 2026-06-04
Bayesian/ABC parameter calibration + uncertainty propagation: report headline claims as intervals, not point estimates


ELares:
**Why.** The headline simulation numbers (1.99x Bliss; 3.7%->0.1% hypoxia; 104:1->4:1 immune; 40%->1.8% penetration) are presented as point estimates from a single parameter set, when nearly all parameters are uncertain placeholders. A contribution-grade model should report **prediction intervals**.

**Approach.** Approximate Bayesian Computation (ABC) / Bayesian calibration using ferroptosis-core as the forward model: maintain posterior distributions over the uncertain rate constants (informed by the GDSC/identifiability work), then propagate them through the headline outputs (Monte Carlo over the posterior) to produce credible intervals.

**Task.** Implement an ABC/ensemble calibration loop; produce credible intervals for each headline claim; add an interval column to the manuscript figures/text where appropriate (clearly marked as model uncertainty, not measurement).

**Acceptance criteria.** Headline claims reported with credible intervals; the ensemble harness is reproducible; documentation states which uncertainty is captured (parameter) vs not (structural).

Depends on the calibration-dataset (#GDSC) and identifiability (#Sobol) issues for the priors. Part of the validation epic.

ELares:
PR #394 (merged) delivers the **prior-predictive (forward) uncertainty foundation**: `scripts/uncertainty_intervals.py` places uniform priors over the 11 PRCC rate-constant ranges and propagates 2000 Monte-Carlo draws through `sim_batch` to 2.5/50/97.5 credible intervals on the single-cell death rate for 8 phenotype × treatment conditions (`analysis/uncertainty-intervals-report.md`; §5 manuscript note). Headline result: the RSL3 death rates span nearly [0,1] under the parameter uncertainty (so they read as order-of-magnitude), while the Glycolytic-RSL3 floor and Persister-SDT ceiling are robust.

**Staying open for the two remaining pieces the issue names:**
1. **Data-conditioned posterior (ABC).** This PR is a forward prior-predictive, not a posterior: it has no kill-rate dataset to condition on. A true Bayesian/ABC posterior needs the external GDSC/DepMap/PRISM drug-response data (issue **#330**) through a likelihood/ABC-rejection step over the same forward model.
[...]

Issue #331, opened 2026-06-04
Global sensitivity (Sobol) + structural identifiability analysis on the three headline claims


ELares:
**Why.** The manuscript has univariate PRCC (#134, complete) but no multivariate sensitivity or identifiability analysis. We do not currently know which parameters (or parameter *interactions*) actually drive the 1.99x Bliss synergy, the hypoxia kill-collapse, and the immune ratio, nor which parameters are even constrainable from the observables we measure.

**What others do.** Variance-based **Sobol** global sensitivity indices (first-order + total-effect) and **structural/practical identifiability** analysis (profile likelihood, Fisher-information rank) are standard in systems-biology/QSP model qualification before a model's predictions are trusted.

**Task.**
1. Sobol indices over the biochemical rate constants (lp_propagation, death_threshold, fsp1 rate, gpx4/gsh dynamics, MUFA setpoint) for each headline output; report first-order vs interaction (total minus first-order) contributions.
2. Structural + practical identifiability: which parameters can be constrained by the calibration-target observables, and which are under-determined? Fix or flag the non-identifiable ones in `parameter_provenance.md`.

**Acceptance criteria.** A sensitivity report (which parameters/interactions drive each claim) + an identifiability table feeding `parameter_provenance.md`; non-identifiable parameters explicitly marked so they are not presented as if data-constrained. Extends, does not duplicate, the univariate PRCC work (#134).

ELares:
## Status after PR #386 (merged) — issue stays open

PR #386 delivered the variance-based Sobol screen + practical-identifiability verdict for the **shared single-cell kill switch** (Persister × RSL3 death rate, `scripts/sobol_sensitivity.py` + `analysis/sobol-sensitivity-report.md`). It screens the PRCC's exact 10 RSL3-relevant rate constants over the PRCC's ranges and reproduces the PRCC top-three driver ranking in order (`lp_propagation` ST=0.50, `gpx4_rate` 0.29, `lp_rate` 0.18; ΣS1≈1.05; six of ten params flagged non-identifiable in `parameter_provenance.md`).

(GitHub auto-closed this on merge because the first PR commit body carried a `Closes #331` keyword that the squash aggregate re-triggered; reopened — the work below remains.)

### Remaining scope: per-headline-output Sobol (Bliss / hypoxia / immune)

[...]

Issue #330, opened 2026-06-04
Calibrate ferroptosis-core kill rates against public drug-response datasets (GDSC/DepMap/PRISM) with held-out validation


ELares:
**The central gap.** Nearly every ferroptosis-core layer is currently "uncalibrated (illustrative)" (see `simulations/calibration/CALIBRATION_STATUS.md`). The single highest-value step toward a real scientific contribution is held-out validation against independent experimental data, of which there is a lot we have not touched.

**What others have / what exists.**
- **GDSC** (Genomics of Drug Sensitivity in Cancer, https://www.cancerrxgene.org/) and **DepMap** (https://depmap.org/) have measured IC50/AUC for many ferroptosis-relevant compounds across 500+ annotated cell lines.
- **PRISM** repurposing screen (Corsello et al., *Nature Cancer* 2020, PMID 32613204) profiled ~4,500 compounds across ~500 cell lines (barcoded pooled viability), including redox/ferroptosis modulators.
- Genome-wide CRISPR ferroptosis modifier screens (Conrad/Stockwell labs) annotate sensitizers/suppressors per genetic background.

**Task.** Build a calibration harness that (1) pulls per-cell-line response for the ferroptosis-relevant inducers in these datasets, (2) maps a tractable subset of phenotypes/genotypes to ferroptosis-core cell setups, (3) **partitions into train/test**, calibrates the free biochemical parameters on the train split, and (4) reports prediction error on the held-out test split (by tissue/genotype). Wire results into a new calibration tier in `targets.yaml` (status: calibrated when a target meets tolerance on held-out data).

**Acceptance criteria.** A reproducible script + documented mapping; at least one headline kill-rate parameter moved from "illustrative" to "calibrated / data-anchored (held-out)"; honest reporting of where the model fails to predict.

Part of the validation epic (see the research roadmap). Complements the closed per-layer follow-ups; this is about *external data*, not internal self-consistency.

Issue #134, opened 2026-04-23
Add global sensitivity analysis for simulation parameters (Sobol or PRCC)


ELares:
## Source
Reviews 3 and 4.

## Problem
The manuscript has a local sensitivity analysis (tornado plot, 22 perturbations of ±50% on 11 rate constants). This is good for directional robustness but does not quantify parameter interactions or identify which parameters the model outputs are most sensitive to globally. Review 3 calls for "global sensitivity analysis, parameter identifiability checks, calibration targets tied to actual experimental data, and clearer separation of directional robustness from quantitative forecastability."

The simulation uses 1,000,000 cells and produces narrow confidence intervals around its assumptions — but narrow CIs from a Monte Carlo do not quantify structural uncertainty. A global sensitivity analysis (e.g., Sobol indices or PRCC) would identify which parameters matter most and where calibration effort should focus.

## What Needs to Be Done
1. Implement a global sensitivity analysis (Sobol indices via SALib, or PRCC) for the ferroptosis-core model
2. Identify parameter interactions (e.g., does GSH depletion rate interact with FSP1 recovery?)
3. Report which simulation outputs are robust vs. parameter-sensitive
4. Use the results to prioritize which parameters most urgently need experimental calibration
5. Add findings to the manuscript (Chapter 5 or Appendix A)

### Owner Refinements
- **Staged approach: PRCC first, Sobol only if warranted.** Do not jump straight to Sobol indices, which require large sample counts and significant compute. The implementation order must be:
  - **Stage 1 (required for close):** PRCC (Partial Rank Correlation Coefficients) on key model outputs (e.g., overall kill rate, lipid peroxidation fraction, GSH depletion time). This is computationally cheap and answers the most important question: which parameters matter most?
  - **Stage 2 (only if warranted):** Sobol indices, but only if PRCC reveals significant non-monotonic interactions that PRCC cannot capture. Document the decision criterion for proceeding to Stage 2.
- **Avoid making this a large compute project before it's a useful methodological result.** The goal is to produce a useful sensitivity ranking, not to demonstrate computational sophistication. If PRCC on 11 parameters with 1000-5000 samples answers the question, that is sufficient. Do not over-engineer.

## Acceptance Criteria
- [ ] PRCC analysis implemented for key model outputs (kill rate, lipid peroxidation, GSH depletion time, etc.)
- [ ] PRCC results identify which parameters the model is most sensitive to
- [ ] Parameter interactions flagged by PRCC are documented
[...]

PR #394, opened 2026-06-05
feat(#332): prior-predictive uncertainty intervals on the single-cell kill switch

ELares:
refs #332 (not Closes — single-cell foundation + harness; data-conditioned posterior + spatial-output intervals are the remaining scope, same single-cell-first scoping as #331).

The manuscript reports the headline numbers as point estimates from one parameter set, when nearly every rate constant is an uncertain placeholder. This adds the foundational piece of "report as intervals": a **prior-predictive uncertainty propagation** over the single-cell death rate.

### What it does
- `scripts/uncertainty_intervals.py`: uniform priors over the **11 biochemical rate constants** across their PRCC ranges (`analysis/prcc-results.json` — the same source the PRCC #134 + Sobol #331 use), **2000 joint Monte-Carlo draws** through `sim_batch` (n=2000, seed=42) per phenotype × treatment, reporting **2.5 / 50 / 97.5 death-rate percentiles** + the default-parameter point estimate. Generic, analytic-testable estimator (`prior_predictive_intervals`).

### Result (`analysis/uncertainty-intervals-report.md`)
| Phenotype | Treatment | point | 2.5% | 97.5% |
|---|---|--:|--:|--:|
| Persister | RSL3 | 40.2% | 1.6% | 99.7% |
| Glycolytic | RSL3 | 0.0% | 0.0% | 0.0% |
| OXPHOS | RSL3 | 1.1% | 0.0% | 75.1% |
| Persister | SDT | 100.0% | 98.2% | 100.0% |
| Glycolytic | SDT | 86.6% | 3.5% | 100.0% |
| OXPHOS | SDT | 99.9% | 68.1% | 100.0% |

The **RSL3 death rates carry very wide intervals** (the bistable tipping point sits inside the plausible range), so they are the least parameter-constrained and most order-of-magnitude; the Glycolytic-RSL3 floor and Persister/OXPHOS-SDT ceilings are robust.

### Honest scope
This is a **prior-predictive (forward)** uncertainty, **not** a data-conditioned Bayesian posterior — there is no in-repo kill-rate calibration dataset. A true posterior needs the external GDSC/DepMap drug-response data (#330) through an ABC step; propagating the same priors through the **spatial** headline outputs (Bliss, hypoxia, immune, penetration) is a deferred ensemble step. Both are the remaining #332 work, hence `refs`.

### Verification
- 3 new tests (analytic estimator validation against the uniform's closed-form quantiles + a binding-based smoke test, `importorskip`-guarded). doc test-count 124→127 / "139 Python tests".
- Python-only (no Rust change, no byte-identity concern). §5 sensitivity note added; v1.tex regenerated.


ELares:
## Review fixes applied (commit 9e17d56)

The adversarial review returned **0 blockers, 1 should-fix, 4 nits** (2/3 reviewers approve; the estimator math + reproducibility were both confirmed sound). All addressed:
- **(should-fix) `point` column accuracy.** It was labeled 'the manuscript's operating point' but ran at n=2000 while Figure 7 uses n=1M (40.2% vs the frozen 42.5%, a sample-size effect). Fixed by computing the point column at `POINT_N_CELLS`=1M — it now reproduces Figure 7 **exactly** (Persister × RSL3 = 42.5%, Glycolytic × SDT = 87.2%); the interval draws stay at n=2000 (the parameter spread dominates the small per-draw MC noise). Report label + §5 note + CLAUDE.md updated.
- **(nit) OXPHOS × SDT 'ceiling'** — it has a ~68% lower tail, so the report + §5 note now distinguish the tight Persister-SDT ceiling (>98%) from the wider OXPHOS/Glycolytic-SDT tails.
- **(nit) NRF2-rescue case added** — `PersisterNrf2 × RSL3/SDT`, the manuscript's flagged primary RSL3-resistance risk: resistant at point (0%) but the interval reaches ~38%, so even that resistance is not parameter-robust.
- **(nit) estimator shape robustness** — replaced the n_samples==n_outputs-ambiguous transpose heuristic with an explicit 'first axis is samples' assert + reshape.

Re-verified: all 3 #332 tests pass (analytic estimator validation + binding smoke over the 8 conditions); doc test-count 127/139; v1.tex regenerated. Python-only (no byte-identity concern). The `refs #332` scoping (single-cell prior-predictive foundation; data-conditioned posterior via #330 + spatial-output intervals deferred) is unchanged.

PR #399, opened 2026-06-06
fix(pdf): map all Greek letters so the release-pdf build exits 0

ELares:
## Problem
The **Build & Release PDF** workflow has failed on the last several pushes to `main`.

Root cause: a literal Greek capital **Σ** (U+03A3) in the §5 sensitivity paragraph (`ΣS1 ≈ 1.0`, added with the #332 uncertainty-intervals text). `pdflatex` with `inputenc=utf8` maps a *limited* non-ASCII set (accented Latin, `±`, curly quotes, em/en dashes) but **not** bare Greek code points, so `Σ` is a hard `! LaTeX Error: Unicode character Σ (U+03A3)`. pdflatex recovers and still emits the full PDF, but exits **non-zero**, which failed the `set -e` compile step even though the document compiled fine (124 pages).

## Fix (defense-in-depth)
1. **`scripts/generate_latex.py`** — a shared `map_greek()` helper maps **every** Greek letter (lowercase + the distinct-glyph capitals Γ Δ Θ Λ Ξ Π Σ Υ Φ Ψ Ω) to its `$\command$`, applied in both the body (`cvt`) and the citation-footnote path. Regenerated `v1.tex` — the only change is `ΣS1` → `$\Sigma$S1`.
2. **`tests/test_latex_unicode_safe.py`** (new) — a CI guard in the Python suite (runs on PR) asserting `v1.tex` has no bare Greek / pdflatex-unsafe Unicode, catching a recurrence **before** the post-merge PDF build. (+3 tests; doc count 146 → 149.)
3. **`.github/workflows/release-pdf.yml`** — tolerate pdflatex's exit code, but gate the release on the *real* signals: PDF exists, no `Emergency stop`/`Fatal error`/`No pages of output` in the log, and a 500 KB size floor; recovered Unicode errors are surfaced as a CI `::warning::`. So a recoverable cosmetic error can never again block a release, while a genuine compile failure still does.

## Verification
- Local 2-pass `pdflatex -interaction=nonstopmode v1.tex` now **exits 0** on both passes (124-page, 1.76 MB PDF, zero `! ` errors, zero `Unicode character` errors).
- Full `tests/` suite (137) passes, including the new guard.
- `v1.tex` non-ASCII after regeneration: only `inputenc`-safe chars (±, em-dash, accented Latin, curly quotes); the Greek block is empty.


ELares:
## Adversarial review pass (resolved in 58ac17c0)

A 2-dimension adversarial review (generator-correctness + workflow/tests, each finding independently re-verified) raised 3, confirmed 1:

**[workflow, should-fix] `|| true` could release a stale pass-1 PDF.** The reviewer empirically reproduced it: with both passes `|| true`, if pass 1 writes a complete PDF and pass 2 is then **killed** (OOM/timeout) before rewriting, the verify step's checks (PDF exists, no fatal marker in the truncated log, size over the floor) all pass and a *stale* pass-1 PDF would be released. **Fixed**: `rm -f v1.pdf v1.log` before compiling, and capture pass 2's exit code — tolerate a non-zero exit **only** when the log shows the known recoverable `Unicode character` class; any other non-zero exit (fatal error or aborted run) now fails loudly. Verified locally: healthy 2-pass → RC=0 (1.76 MB PDF); simulated killed pass 2 (RC≠0, no Unicode marker) → fails as intended.

The other 2 raised (generator double-wrapping risk; nested-math) were verified as **false positives** — `map_greek` mirrors the pre-existing α/δ/λ replacement pattern, v1.md has no literal math spans containing bare Greek, and the regeneration's only diff is `ΣS1` → `$\Sigma$S1` (a 2-pass local compile exits 0 with zero Unicode errors).

PR #386, opened 2026-06-05
feat(#331): Sobol global sensitivity + practical identifiability of the kill switch

ELares:
## Summary

refs #331. Adds variance-based **Sobol** global sensitivity + **practical identifiability** of the ferroptosis kill switch, as a model-qualification check that extends the univariate PRCC (#134). Self-contained (Saltelli 2010 S1 + Jansen 1999 ST; no SALib dependency).

### Result
- `scripts/sobol_sensitivity.py` computes Sobol indices of the single-cell kill rate (`ferroptosis_core` binding, **Persister + RSL3** mid-range operating point — chosen because OXPHOS-RSL3 sits near the 0 floor and most SDT points near the 1 ceiling, compressing their variance), over 8 biochemical rate constants, ±50% sweep, Saltelli N=2048 (20,480 evaluations, ~40s).
- **`lp_propagation` dominates** (total-effect ST=0.66), `gpx4_rate` second (0.33), together ~94% of the kill-rate variance. **Interactions are small (ΣS1≈0.97)**, which *confirms* the manuscript's "Stage 2 (Sobol) not warranted" PRCC call from the other direction (the univariate ranking was adequate).
- **Practical identifiability:** 6 of 8 screened rate constants (`fenton_rate`, `gsh_scav_efficiency`, `gsh_km`, `fsp1_rate`, `death_threshold`, `gpx4_degradation_by_ros`) move the kill rate negligibly (ST<0.05), so they are **not constrainable from kill-rate data**, and are flagged as such in `simulations/calibration/parameter_provenance.md` (between markers) rather than presented as data-fitted.

### Changes
- `scripts/sobol_sensitivity.py` + `analysis/sobol-sensitivity-report.md` (reproducible; reads only the committed binding; fixed seeds).
- `parameter_provenance.md`: a regenerated identifiability block (the #331 deliverable).
- `v1.md` §5 (+ regenerated `v1.tex`): a note recording the Sobol confirmation + identifiability verdict alongside the PRCC subsection.
- `tests/test_sobol_sensitivity.py`: the Saltelli/Jansen estimator validated against the **analytic Ishigami** indices (S1/ST within 0.04) + invariants + a small binding run guarding the lp_propagation headline.

No `ferroptosis-core` change (reuses bindings); no version bump. Test count 121→124 (136 documented).

[...]

PR #401, opened 2026-06-06
feat(#331): per-headline Morris sensitivity harness + Bliss-synergy screen

ELares:
**Part 2 of #331** (follow-up to the single-cell Sobol #386). Consumes the #331 PR 1 `FERRO_PARAM_OVERRIDES` hook to screen which biochem rate constants drive a **headline** output that only the simulation binaries produce.

## Method
**Morris elementary effects (screening), not Saltelli/Sobol.** A full Sobol design needs `N_base*(k+2)` evals (tens of thousands); each binary run is orders of magnitude costlier than the `sim_batch` call the single-cell Sobol used, so full Sobol is infeasible for the binaries. Morris gives `mu*` (importance ranking, the screening analogue of a total effect) and `sigma` (interaction/nonlinearity) in `r*(k+1)` runs. Self-contained estimator (no SALib), analytically validated.

## What
- **`scripts/headline_sensitivity.py`**: Morris screen over the same 11 PRCC rate constants, driving `sim-combo-mech` via the env override (each run isolated in a temp cwd → safe parallelism). `--smoke`, `--trajectories`, `--workers` flags; deterministic.
- **Bliss-synergy headline** (RSL3 + FSP1i `synergy_score`). Converged result (r=100):

  | finding | result |
  |---|---|
  | top drivers | `lp_propagation`, `gsh_scav_efficiency`, `gpx4_rate` (the LP + GSH/GPX4 axis, consistent with the single-cell Sobol) |
  | interaction structure | `sigma > mu*` for **every** active param → the synergy is strongly interaction-driven/nonlinear, which a univariate PRCC cannot see |
  | structural zeros | `sdt_ros` (no SDT in the pair) and `rsl3_gpx4_inhib` (combo-mech bakes RSL3 GPX4-inhibition into a fixed `DrugEffect`) → not constrainable from this headline |

- **`tests/test_headline_sensitivity.py`** (4 tests): analytically validate the Morris estimator on a known linear+interaction function (ranking, sigma signal), the one-coordinate-per-step trajectory invariant, and the PRCC param set. Runs in CI (no binary needed).
- Report: `analysis/headline-sensitivity-report.md`. Manuscript §5 + CLAUDE.md updated; test count 149 → 153.

## Honest scoping
The two **sim-tme** headlines (hypoxia kill-collapse, immune ratio) are the documented **next increment**, not omitted by oversight: the SDT kill sits near a ceiling that compresses its variance, and the raw `immune_kills` ratio is **confounded** (SDT pre-clears most cells by ferroptosis, so the raw immune ratio runs *opposite* to the headline's DAMP-amplification framing). Each needs a deliberately-chosen observable. **#331 stays open** for these.

No library/version change (Python script + manuscript only).


ELares:
## Adversarial review pass (resolved in f7123f03)

2-dimension adversarial review (estimator + science/docs), each finding re-verified. 5 raised, 3 confirmed (two were the same finding):

**[should-fix] The report mislabeled `gpx4_degradation_by_ros` as a third "structural" zero.** Only `sdt_ros` (no SDT in the pair) and `rsl3_gpx4_inhib` (combo-mech bakes RSL3 GPX4-inhibition into a fixed `DrugEffect`) are truly structural — never read on this code path. `gpx4_degradation_by_ros` IS on the active biochem path (used in the GPX4 dynamics, `biochem.rs:369`); its `mu*=0` is an *empirical/below-resolution* zero (a perturbation never flips the bistable outcome over its PRCC range), not a structural disconnection. This also contradicted the PR's own manuscript text ("two parameters are structurally inert"). **Fixed**: `write_report` now splits `mu*=0` into structural (named explicitly) vs empirical/below-resolution zeros; report regenerated.

**[nit] "interaction indicator" overstated.** Morris `sigma` doesn't separate interaction from single-parameter nonlinearity. **Fixed**: report + manuscript §5 now say "interaction/nonlinearity indicator" / "non-additive".

The 2 unconfirmed findings were verified false positives. Report and manuscript are now consistent; 141 Python tests pass.

(Note: `python-test` did not auto-trigger on the first push — a path-filter quirk — so this commit re-fires it.)

Issue #343, opened 2026-06-04
Reaction-diffusion (non-exponential / biphasic) oxygen + drug fields; benchmark vs PhysiCell / published PDE models


ELares:
**Gap + contradiction.** Our O2/drug supply uses edge-distance or Krogh exponential proxies (or explicit vessels with exp-decay). Real tumor fields are not monotonic: abnormal/fractal vasculature produces **biphasic** profiles with high-drug/low-O2 pockets and avascular cold zones. An exponential model averages these away and can mis-rank where drugs reach but O2 does not.

**What others do.** Computational-oncology platforms (PhysiCell, CompuCell3D, Chaste, Morpheus) solve **reaction-diffusion** PDEs for O2/nutrient/drug fields against explicit vessel sources with consumption, which reproduces non-monotonic gradients. We already ship a PhysiCell C-FFI, so interoperability is within reach.

**Task.** Add a reaction-diffusion solver (steady-state and/or transient) for O2 and drug over the vessel network (source + diffusion + consumption), as an alternative to the exponential supply field; benchmark the resulting fields and kill predictions against PhysiCell (via the existing FFI) and/or a published tumor reaction-diffusion model.

**Acceptance criteria.** A reaction-diffusion supply option (off-by-default/byte-identical default); a benchmark comparing it to the exponential proxy and to an external PDE model; documentation of where the proxy is adequate vs misleading. Builds on the vasculature work (#191/#268, closed).

PR #406, opened 2026-06-06
feat(#343): reaction-diffusion supply field (PR 1 of 2, library)

ELares:
Part 1 of #343. Adds the `reaction_diffusion` module to ferroptosis-core: a steady-state linear reaction-diffusion solver for an O2/drug supply field over the same vessel network the exponential proxy uses. The sim-tme-3d wiring + the proxy-vs-RD benchmark report are PR 2.

## Why (#343)
`vasculature::vessel_supply_field` and the oxygen/Krogh proxies are **monotonic in distance to the nearest source**. Real tumor O2/drug fields are not: an irregular vessel network superposes the contributions of *all* nearby vessels while tissue consumption draws the field back down, so the steady state has **non-monotonic pockets** (a point can be far from any single vessel yet well-supplied because it sits between several, or close to one vessel yet starved because consumption around it is high). An exponential nearest-vessel proxy averages these away. This module is the off-by-default alternative that can represent them.

## What

```
D ∇²c − k·c = 0   in tumor (consuming) tissue
D ∇²c       = 0   in stromal tissue
c = 1             at vessel voxels (Dirichlet sources)
∂c/∂n = 0         no-flux domain boundary (finite-volume)
```

- 6-point von Neumann stencil; `γ = k·h²/D = (h/λ)²`, tumor-only consumption; finite-volume no-flux boundary (drop the out-of-grid neighbor, the same reflective convention the slab/oxygen layers use).
- Deterministic SOR (fixed linear-index Gauss-Seidel sweep with over-relaxation).
- The diffusion-consumption length `λ = sqrt(D/k)` is set equal to the proxy's `lambda_um`, so a **single isolated source reproduces `exp(−dist/λ)`** (the proxy is exactly the single-source limit) and any divergence is the genuine multi-vessel/consumption effect.
- Public API: `ReactionDiffusionConfig`, `reaction_diffusion_solve` (raw field + convergence diagnostics), `reaction_diffusion_supply_field` (drop-in alternative to `vessel_supply_field`, non-tumor→1.0), `analytical_1d_slab`.

## Validation
7 unit tests, including:
- **`solver_reproduces_1d_analytical_slab`** — the converged solver matches the closed-form 1-D **Thomlinson & Gray (1955)** cosh profile (`cosh((W−x)/λ)/cosh(W/λ)`), the published-model benchmark, to max abs error < 0.01.
- **`two_sources_enrich_the_midpoint_above_the_nearest_vessel_proxy`** — between two vessels the RD field exceeds the nearest-vessel exponential proxy, i.e. the proxy under-supplies multi-vessel pockets (the headline #343 effect).
- `single_isolated_source_matches_exponential_proxy` (e-folding length λ), `solver_is_deterministic`, `supply_field_overrides_non_tumor_to_unity`, `analytical_1d_slab_reduces_to_exponential_for_wide_slab`, empty-vessels panic.

## Byte-identity
Opt-in alternative supply field — **nothing in the production matrix calls it**, so this is byte-identical. Verified locally: golden 20³×80 tests pass and the **full 60³×180 production `summary.json` SHA-256 matches `expected_summary.sha256`**. Not in the C ABI (header unchanged). Crate `0.49.0 → 0.50.0`.

refs #343


ELares:
## Adversarial review addressed (commit d09c3322)

Ran a 4-dimension adversarial review (numerics / byte-identity / Rust quality / docs-claims) with per-finding adversarial verification. 4 of 7 findings confirmed; all 4 addressed:

1. **`analytical_1d_slab` NaN for tiny λ** (should-fix) — the direct `cosh(W/λ)` form overflows to `inf` for `W/λ ≳ 710`, giving `inf/inf = NaN` (and `clamp` doesn't sanitize NaN). Rewrote in the algebraically-identical, overflow-safe subtracted-exponent form `(e^{-x/λ} + e^{-(2W−x)/λ})/(1 + e^{-2W/λ})`; added a finite/positive-λ `debug_assert` and a tiny-λ regression test.
2. **"published-model benchmark" overstated a self-consistency check** (should-fix) — the analytical slab is the closed form of the *same* PDE the solver discretizes, so the test verifies the discretization converges to its own equation (numerical verification, not an external cross-check). Relabeled as an "analytical self-consistency check" in the module/fn docstrings, test comment, README row, and CLAUDE.md clause, and stated plainly that the genuine external benchmark (proxy-vs-RD + PhysiCell/published-PDE) and the proxy-adequacy docs are **PR 2**.
3. **`reaction_diffusion_supply_field` discarded the non-convergence signal** (nit) — added `debug_assert!(sol.converged, …)` after the solve + a docstring pointer to use `reaction_diffusion_solve` when convergence must be checked in release.
4. **Thomlinson & Gray 1955 attribution** (nit) — that paper is the diffusion-limited-distance concept, not the derivation of the cosh form; separated the formula (generic textbook solution) from the lineage citation everywhere.

All changes are test-only / doc-only / opt-in: golden 20³×80 byte-identity tests still pass and no production-matrix code path changed. 8 module tests pass; fmt + clippy clean.

Issue #346, opened 2026-06-04
Improve tagger recall (55%) via MeSH hierarchical expansion + embedding/semantic retrieval


ELares:
**Why.** The keyword tagger is 96% precision but only **55% recall** on the 100-article gold set, so absence claims are provisional and rare mechanisms are under-counted. Two complementary, well-established upgrades:
1. **MeSH hierarchical expansion**: PubMed indexes MeSH natively; expanding each mechanism with its MeSH subtree + entry-term synonyms recovers articles that free-text queries miss.
2. **Embedding/semantic retrieval**: biomedical sentence embeddings (e.g. PubMedBERT/SPECTER-class) rank candidate articles by semantic similarity to a mechanism description, catching non-standard terminology.

**Task.** Add a MeSH-expansion layer and an embedding-based retrieval/ranking layer to the tagging pipeline; **re-measure precision/recall on the existing gold set** to quantify the gain; keep the keyword tagger as the reproducible baseline.

**Acceptance criteria.** Recall on the gold set improved and reported (with the precision tradeoff); the new layers are reproducible and documented; the manuscript's coverage caveats updated with the new numbers.

ELares:
## Status: MeSH leg delivered + measured; remaining scope migrated

**Delivered (PRs #410 + #413):** an off-by-default MeSH-descriptor evidence fallback (`FERRO_MESH_EXPANSION=1`) + the committed, reproducible re-measurement harness (`scripts/retag_gold_set.py`) + a CI precision-floor regression guard. On the 100-article gold set it lifts **evidence-level detection recall 55.2% → 67.8% at 95.2% precision** (`analysis/evidence-gold-mesh-eval.md`), off-by-default so the production corpus stays byte-identical and the 55% figure remains the corpus reality until deliberately promoted.

**Two scoping findings that reshaped this issue:**
1. The "55% recall" this issue cites is **evidence-LEVEL** detection recall (what the MeSH leg moves). 
2. The existing gold set is **circular for mechanism recall** (every row's `sample_mechanism` is already in the tagger's predicted mechanisms, because rows were sampled *from* tagger positives), so **mechanism** recall — the "rare mechanisms under-counted" motivation — is **currently unmeasurable**.

**Remaining named scope, migrated to focused follow-ups:**
[...]

PR #410, opened 2026-06-06
feat(#346): MeSH-descriptor evidence-detection fallback (off-by-default, PR 1 of 2)

ELares:
Part 1 of #346. The keyword/pub_type evidence tagger is 96% precision but only **55% recall** on the gold set, so absence claims are provisional. This adds a curated **controlled-vocabulary MeSH fallback** to `match_evidence_level`: when the pub_type + keyword passes find no evidence level, infer one from the article's MeSH descriptors via exact set-membership against small, precision-safe **leaf** descriptor families.

PR 2 adds the committed re-measurement harness + regenerated `evidence-gold-eval.md` + a CI precision-floor regression test + the honesty/docs updates.

## Scope note (important)
Investigation found the "55% recall" is **evidence-level detection** recall, and that the gold set is **circular for mechanism recall** (every row's `sample_mechanism` is already in the tagger's predicted mechanisms — articles were sampled *from* tagger positives). So this PR targets the measurable evidence-level recall; a non-circular mechanism gold set + the embedding/semantic leg are deferred to follow-up issues (opened in PR 2).

## Design (per the #346 feasibility analysis)
- **Not MeSH-tree expansion.** The descriptor strings are already concatenated into the searchable text, so ancestor/subtree expansion is a near no-op and re-introduces leaky broad terms. The lever is **exact leaf set-membership** on `fm['mesh_terms']`.
- **Conflict order** invivo > invitro > clinical-other (a paper with both animal and cell-line MeSH is an in-vivo study).
- **clinical-other** restricted to narrow study-*design* descriptors (Retrospective/Prospective/Cohort/Case-Control Studies, Registries), **not** bare Humans/age-groups (clinical *context*, over-tags), and **vetoed by opinion pub-types** (editorial/comment/letter) so a commentary citing a cohort isn't promoted.
- **theoretical gets no MeSH descriptor** on purpose: candidates (Transcriptome / Gene Expression Profiling) co-occur with clinical/animal MeSH ~40% of the time → would regress precision. That residue + empty-MeSH (~19.5% of records) is the documented floor for the deferred embedding leg.

## Byte-identity
Off by default (`FERRO_MESH_EXPANSION=1` to enable), so the production corpus run and every existing tag stay **byte-identical** until deliberately promoted. The fallback is reached only after the review/protocol guard and below the authoritative pub_type + keyword passes, so it never overrides them or re-tags reviews. The `flag-OFF-reproduces-baseline` test locks this.

## Measured gain (local; committed harness is PR 2)
Flag ON lifts binary evidence-detection recall **55.2% → 67.8%** (48 → 59 TP) at **95.2%** precision (96.0% baseline; 1 net-new defensible FP) — matching the feasibility target. Permanent floor 26.4% (empty-MeSH + non-discriminative) motivates the deferred embedding leg.

## Tests
13 unit tests (`tests/test_mesh_fallback.py`): conflict order, editorial veto (PMID 32167722 regression guard), empty-MeSH floor, non-discriminative rejection, pub_type precedence, review guard, keyword-pass-still-wins, and the flag-OFF reversibility gate. **Full pytest 155 green**; test-count guard + CLAUDE.md/README doc counts bumped 154 → 167.

refs #346


ELares:
## Adversarial review addressed (commit ad4c45e1)

4-dimension adversarial review (precision-correctness / byte-identity / claims-measurement / python-quality) with per-finding verification; 2 of 6 confirmed, both fixed:

1. **(should-fix) Opinion-pubtype veto only gated `clinical-other`.** An editorial/letter/comment/news carrying a topical organism or cell-line MeSH descriptor (an editorial *about* an animal study still carries "Animals") was promoted to a preclinical level — the exact FP class the PR claims to defend against. The verifier empirically confirmed **9 opinion-pubtype corpus records** would be mis-promoted when the flag is enabled (e.g. 23254777 Editorial → preclinical-invivo). **Fixed** by hoisting the veto to the top of `match_evidence_mesh` so it gates **all** branches; updated the module comment + docstring; added a regression test (editorial + {Animals} → "").
2. **(nit) Tests didn't anchor descriptor strings to the corpus format.** Synthetic `_fm()` tests fed the exact literal strings, so a typo (missing comma) would pass all tests yet never fire. **Fixed** with a corpus-anchored test asserting each marker family has ≥1 descriptor actually present in `corpus/by-pmid` mesh_terms.

The other 4 findings were verified and refuted.

**Gold-set metrics unchanged by the fix** (the gold-set FP was a clinical-other Retrospective case, not an opinion piece): flag ON still **67.8% recall / 95.2% precision**; the veto hoist improves corpus-wide precision when the flag is promoted. Full pytest **157 green**; test-count guard 155→157 + doc strings 167→169. Still off-by-default / byte-identical.

Issue #412, opened 2026-06-06
Build a non-circular mechanism-tagging gold set (mechanism recall is currently unmeasurable; #346 follow-up)


ELares:
**Context.** While scoping #346, the existing evidence gold set (`analysis/evidence-gold-set-v2.csv`) was found to be **circular for mechanism recall**: every one of the 270 rows has its `sample_mechanism` already present in the tagger's predicted `mechanisms` (100%), because the articles were sampled *from* the tagger's positive predictions. So it has **no false negatives by construction** and cannot measure mechanism-tagging recall — only the EVIDENCE-LEVEL recall (the documented 55% / now 67.8% with the MeSH leg) is measurable on it. The gold set also has no manual *mechanism* labels (only `gold_evidence_level`).

**Why it matters.** #346's motivation includes "rare mechanisms are under-counted", i.e. mechanism recall, but that quantity is currently **unmeasurable**. Field-level absence/coverage claims (CLAUDE.md: "corpus-level non-detection is provisional") rest on mechanism recall we cannot quantify.

**Task.** Build an independent mechanism-tagging gold set sampled WITHOUT conditioning on the tagger's positives:
- sample candidate articles per mechanism by an independent route (e.g. PubMed mechanism queries, or random corpus draws), so true-positives-the-tagger-missed can appear;
- add manual mechanism labels (per-article: which mechanisms genuinely apply);
- extend `scripts/evaluate_evidence_gold_set.py` (or a sibling) to score mechanism precision/recall against those labels.

**Acceptance criteria.** A non-circular mechanism gold set + a reproducible mechanism precision/recall measurement; the manuscript's mechanism-coverage caveats updated with the (first-ever) measured mechanism recall.

**Note.** This is a labeling effort (needs manual mechanism annotation), so it is human-in-the-loop, not a pure-code PR.

ELares:
Reopening: #417 (merged) delivered the non-circular mechanism **recall** measurement (MeSH concordance: 9 mechanisms, 90.0% volume-weighted / 87.5% macro, `scripts/mechanism_recall.py` + `analysis/mechanism-recall-report.md`). It was auto-closed by a stray "close #412" phrase in the PR description, but two acceptance criteria remain:

1. **Manuscript update** — the mechanism-coverage caveats must cite the (first-ever) measured mechanism recall. Pending follow-up PR.
[...]

Issue #411, opened 2026-06-06
Embedding/semantic-retrieval tagger leg for the MeSH-unrecoverable residue (#346 follow-up)


ELares:
**Context.** #346's first leg (the off-by-default MeSH-descriptor evidence fallback, PRs #410 + the measurement PR) lifted gold-set evidence-detection recall 55.2% → 67.8% at 95.2% precision. The feasibility analysis (`analysis/evidence-gold-mesh-eval.md`) showed a **hard MeSH floor**: of the 39 baseline false-negatives, **9 carry empty `mesh_terms`** (recent / not-yet-indexed) and a further ~14 carry MeSH that is present but non-discriminative (notably preclinical-invitro, which PubMed rarely tags with a Cell Line descriptor — only 1/9 recoverable). MeSH cannot reach that residue.

**Task.** Add the second #346 leg: an embedding/semantic-retrieval ranking layer (biomedical sentence embeddings, e.g. PubMedBERT / SPECTER-class) that ranks candidate articles by semantic similarity to a mechanism/evidence-level description, to recover the empty-MeSH and non-discriminative residue. Keep the keyword + MeSH layers as the reproducible baseline; re-measure on the gold set (extend `scripts/retag_gold_set.py` to a third column) and report the additional lift + precision tradeoff.

**Acceptance criteria.** A semantic layer that measurably recovers part of the MeSH-unrecoverable residue, with the precision tradeoff reported; reproducible.

**Blocker / dependency decision (why this is separate from the MeSH leg).** A biomedical embedding model is **not** in `requirements-lock.txt` (only numpy/scipy/requests are pinned), and the repo's CI contract is **offline/reproducible** (the python-test workflow has no network/model download). So this needs a deliberate dependency story: vendor/cache a model, pin it in `requirements-lock.txt` + `PROVENANCE.yaml`, and decide whether the embedding pass runs in CI or only locally with cached embeddings committed. That decision is the gating step.

PR #413, opened 2026-06-06
feat(#346): MeSH-fallback re-measurement harness + BEFORE/AFTER report + CI precision-floor guard (PR 2 of 2)

ELares:
Completes the MeSH leg of #346 (PR 1 = #410, the off-by-default fallback). Adds the committed, reproducible measurement + a CI guard + honesty/docs.

## What
- **`scripts/retag_gold_set.py`** — re-tags the 100 manually-labeled gold PMIDs with the MeSH fallback OFF (committed baseline) and ON, offline/deterministic (no corpus mutation, no network). Writes `analysis/evidence-gold-mesh-eval.md` (BEFORE/AFTER binary precision+recall, per-level recall lift, empty-MeSH-floor split) + a per-PMID `evidence-gold-set-v1-retag.csv`. Exposes `compute_metrics()` so the report and the CI guard never diverge.
- **Measured** (the headline #346 deliverable): recall **55.2% → 67.8%** (+11 TP) at **95.2%** precision (96.0% baseline; 1 net-new FP). Per-level lift in clinical-other (3→6) and preclinical-invivo (19→22); **9 of 39 baseline FNs are empty-MeSH** (the documented unrecoverable floor).
- **`tests/test_mesh_gold_regression.py`** — CI guard: flag-ON precision ≥ 0.93 AND recall ≥ 0.62 (robust, not point-pinned — the 100-row set makes one FP a ~2-point swing), so a future leaky-descriptor edit fails loudly. Runs in the existing pytest job.

## Honesty / scope
- CLAUDE.md + README note the MeSH gain is **off-by-default and not applied to the production corpus**, so the 55% figure remains the corpus reality. The manuscript's recall caveats describe the production tagger and stay accurate (the layer is not promoted), so no manuscript change.
- **Follow-ups opened:** #411 (embedding/semantic leg for the MeSH-unrecoverable residue, gated on a model dependency vs the offline-CI contract) and #412 (a non-circular mechanism-tagging gold set — the current gold set is **circular for mechanism recall**, so mechanism recall is unmeasurable today).

## Verify
Full pytest **158 green**; test-count guard 157→158 + doc strings 169→170. Off-by-default / byte-identical (no corpus re-tag). Regenerate the report: `python scripts/retag_gold_set.py`.

refs #346 #411 #412


ELares:
## Adversarial review addressed (commit 15209623)

4-dimension review (measurement-correctness / reproducibility / ci-guard / claims-docs); 2 of 3 findings confirmed, both nits on the regression guard's honesty, both fixed:

1. **`n>=90` guarded the CSV, not the corpus.** A partial corpus loss (≤5 dropped gold records) passed silently while the docstring claimed it guarded "missing corpus records". **Fixed**: `compute_metrics` now returns `records_found` (gold PMIDs whose corpus `.md` actually resolved) and the test asserts `records_found >= 90`.
2. **Docstring oversold the precision floor.** With only 13 gold negatives the 0.93 floor resolves a **≥2-net-new-FP** regression, not "high-80s"; a single-FP leak is below resolution. **Fixed**: softened the docstring to the real resolution + the small-set limitation.

The measurement-correctness, reproducibility, and the third finding were verified clean/refuted (the harness reproduces the committed baseline exactly, is offline/deterministic, and mutates no frozen artifact).

**On closing #346:** per the review's recommendation, this PR does **not** close #346 — two of its named legs (embedding/semantic retrieval, mechanism recall) are unmet and migrated to #411/#412. #346 stays open with a scope-migration note (the MeSH leg is delivered + measured + documented here).

Report artifacts unchanged by the fix; full pytest **158 green**.

PR #414, opened 2026-06-06
feat(#332): Bliss-synergy prior-predictive interval (spatial-headline extension)

ELares:
Extends the prior-predictive uncertainty work (#332) from the single-cell kill switch to the **first spatial/combination headline**: the RSL3 + FSP1i Bliss synergy (the manuscript's 1.99×). `scripts/uncertainty_intervals.py`'s own docstring named this as the tractable follow-up; the full ABC posterior remains blocked on the #330 GDSC data, so this delivers the tractable prior-predictive piece.

## What
- **`scripts/headline_uncertainty.py`** — Monte-Carlo samples the joint uniform prior over the same 11 PRCC rate constants (seeded, deterministic) and runs `sim-combo-mech` under `FERRO_PARAM_OVERRIDES` for each draw, **reusing `headline_sensitivity.run_bliss` + the param ranges** so the prior and observable match the Morris screen exactly. Writes `analysis/headline-uncertainty-report.md`.
- **Result (300 draws):** the Bliss synergy spans a **95% prior-predictive interval ~[1.00×, 5.24×]** (median ~1.35×) around its 1.99× point estimate. The **supra-additive direction is robust** at the lower 95% bound (a small prior tail reaches ~0.95×, just sub-additive); the **magnitude is not** — consistent with the manuscript's directional-not-quantitative framing.

## Scope / honesty
- Prior-predictive (parameter) uncertainty over **uniform** priors — **not** a data-conditioned ABC posterior (blocked on #330) and not structural uncertainty. The costlier `sim-tme` headlines (hypoxia, immune, penetration) are the next extension.
- Manuscript §5 sensitivity paragraph updated: the spatial-headline propagation it called "deferred" now reports the Bliss interval; hypoxia/immune/penetration remain deferred. No frozen number changed.

## Verify
5 pure-estimator unit tests (no binary, mirroring `test_headline_sensitivity`); v1.tex regenerated (1-line diff, pdflatex-safe). Full pytest **163 green**; test-count guard 158→163 + doc strings 170→175. Regenerate: `python scripts/headline_uncertainty.py`.

refs #332


ELares:
## Adversarial review addressed (commit 40423452)

4-dimension review (stat-correctness / reproducibility / claims-manuscript / python-quality); 4 of 6 findings confirmed, all fixed:

1. **(BLOCKER) NaN synergy poisoned every percentile.** `sim-combo-mech` emits `NaN` when the Bliss baseline ≤ 0.001 (both single agents kill ~0%, a ferroptosis-resistant prior corner). `float("NaN")` succeeds, so it slipped past the `try/except` and `np.percentile` propagated NaN to **all** quantiles. The committed seed 332/n=300 was NaN-free *by luck*, but 3/8 other seeds (and seed 332 at n≥800) produced all-NaN reports with `n_failed=0`. **Fixed**: a pure, unit-tested `_partition` helper splits outcomes into valid / undefined-Bliss / subprocess-failure; percentiles only see finite values. Seed 332/n=300 is unchanged (0 undefined), so the interval [1.00×, 5.24×] / median 1.35× **and the manuscript number are unaffected**.
2. **(should-fix) Misleading `n_failed`** (NaN counted as "succeeded"). Now reports three categories and documents the undefined-Bliss region as an honest robustness note.
3. **(should-fix) `_default_bliss` duplicated CSV-parsing.** Extracted a shared `read_bliss_synergy(workdir)` in `headline_sensitivity.py` (behavior-preserving); both paths reuse it.
4. **(nit)** Replaced the stale stream-of-consciousness comment in `main()`.

+2 partition unit tests (NaN/inf/None dropped). `headline_sensitivity` tests still pass (refactor behavior-preserving). Full pytest **165 green**; guard 163→165 + docs 175→177. Report regenerated (interval unchanged); v1.md/v1.tex untouched.

PR #416, opened 2026-06-06
feat(#332): penetration prior-predictive intervals — completes the headline set

ELares:
Propagates the same uniform PRCC prior through the **last headline family** — the tissue-specific drug penetration (the manuscript's 40% in 2D → 12.1% / 2.6% / 1.8% vessel-wall RSL3 kill). With this, **all four headline families** now read as prior-predictive intervals (single-cell + Bliss + sim-tme + penetration); only the data-conditioned ABC posterior remains, blocked on #330.

## What
- **Rust**: wires the #331 `FERRO_PARAM_OVERRIDES` hook into `sim-tissue-pk`'s `main()` (mirroring sim-tme/sim-combo-mech). **Env unset → no overrides → byte-identical** (verified: default `tissue_pk_summary.json` unchanged, so the penetration figure data is untouched).
- **Harness**: `headline_uncertainty.py --headline penetration` runs `sim-tissue-pk` (~4 s/run) per draw and collects the RSL3-like vessel-wall death rate for each tissue via a shared `run_tissue_pk_observables` / `extract_tissue_pk_observables`. Defensive per-tissue NaN partition.
- **Result (300 draws, 0 failed):** every tissue's vessel-wall kill carries a **very wide** interval (well-vasc median 0.23 ~[0.00, 0.93]; poorly-vasc 0.08 ~[0.00, 0.83]; CNS/BBB 0.04 ~[0.00, 0.77]) because the bistable RSL3 switch dominates — but the **across-tissue ordering (well > poorly > CNS) is preserved across the prior**, so the penetration-gradient *direction* is parameter-robust, the absolute magnitudes are not.

## Scope / honesty
- The transport params (penetration length, vascular permeability, ECM tortuosity) are held at their **fixed, uncalibrated per-tissue presets** — this interval is the *biochem-rate* uncertainty at each tissue scenario; the transport params have no documented range table so are not swept (flagged in the report). Prior-predictive, not a posterior (blocked on #330).
- Manuscript §5 updated (penetration now reported; only the ABC posterior deferred); v1.tex regenerated, local pdflatex builds clean (129 pp).

## Verify
+3 pure unit tests (`_partition_penetration` NaN/missing-key/failure). Full pytest **169 green**; guard 167→169 + docs 179→181; cargo fmt clean. Regenerate: `python scripts/headline_uncertainty.py --headline penetration` (~2 min).

refs #332


ELares:
## Adversarial review addressed (commit bb6d1f74)

4 confirmed (1 should-fix + 3 nits), all fixed:

1. **(should-fix) The "ordering preserved across the prior" claim was asserted, not shown.** The per-tissue 95% intervals overlap almost completely (the biochem prior dominates the spread), so the marginals + medians alone don't establish well>poorly>CNS. The ordering is a **within-draw paired** property (each draw applies the same biochem override to all three tissues at decreasing fixed vessel-wall concentration 0.8/0.4/0.15). **Fixed**: a pure `_ordering_preserved_fraction` computes the per-draw monotone fraction *before* the marginal partition discards the pairing — **well≥poorly≥CNS held in 300/300 draws**. The report, manuscript §5, CLAUDE.md, and the CLI now state this data-backed paired result + the mechanism. +2 unit tests (one detects a per-draw inversion the marginals would hide).
2. **(nit)** PR-body test count corrected (+4, not +3).
3. **(nit)** module docstring no longer calls penetration a "deferred extension" — implemented, with usage + output path added.
4. **(nit)** `write_tme_report`'s scope sentence (and the committed tme report, hand-mirrored since the numbers are deterministic) no longer mislabels penetration as "pure-Python"/"remaining".

Full pytest **171 green**; guard 169→171 + docs 181→183; local pdflatex builds clean (129 pp). Penetration marginals unchanged (deterministic). Rust hook + byte-identity unchanged from the first push.

PR #417, opened 2026-06-06
feat(#412): non-circular mechanism-tagging recall via MeSH concordance (PR 1)

### LINKED ISSUE #412: Build a non-circular mechanism-tagging gold set (mechanism recall is currently unmeasurable; #346 follow-up)

ELares:
## What & why

The evidence-tagging gold set gives a real evidence-tier recall (55%), but the **mechanism** tagger has had **no non-circular recall number**: the obvious reference label (MeSH descriptors) is folded straight into the matched text by `get_searchable_text`, so any MeSH-vs-tag recall would be partly tautological. This PR makes mechanism recall measurable without that circularity (refs #412).

## How the circularity is removed

1. **Leakage-free re-tag.** `tag_articles.get_searchable_text` gains an `include_metadata` kwarg (default `True`, so production tagging is **byte-identical**). The measurement re-runs the tagger with `include_metadata=False` (title + abstract only; MeSH/disease/gene/drug annotations excluded), so the reference label is never visible to the matcher whose recall we measure.
2. **Independent descriptors only.** Each candidate MeSH leaf is classified at runtime as *independent* or *leaky* via the **live** `text_matches_keyword` matcher (so it can never drift): a descriptor whose own string would trip the mechanism's keywords is *near-tautological* and is reported separately, never in the headline.

## Result (frozen corpus, 4,830 records)

- **9 measurable mechanisms** (independent pool ≥ N_MIN=30): **90.0% volume-weighted / 87.5% macro** recall.
- ADC 97.5%, synthetic-lethality 94.9%, electrochemical 93.7%, metabolic 92.9%, car-t 90.6%, immunotherapy 89.2%, sonodynamic 87.9%, bispecific 86.1%, **epigenetic 54.7%**.
- Epigenetic is the honest low: a mix of a **real keyword-coverage gap** (the field's "epigenetic modifications/regulation", EZH2/BET inhibitor terms are not in the keyword set, an actionable finding) and reference-scope breadth of the lone independent "DNA Methylation" leaf.
- Mechanisms whose only discriminative MeSH leaf is keyword-tautological (crispr/nanoparticle/oncolytic-virus/microbiome/…) or that have no descriptor (ttfields/CAP/cuproptosis/disulfidptosis/radioligand) are reported **not-MeSH-measurable, never 0%**.

This is **distinct** from the evidence-tier 55% recall and does **not** touch the 96% evidence precision.

## Files

- `scripts/mechanism_recall.py` (offline, deterministic; writes `analysis/mechanism-recall-report.md` + `.json`)
- `analysis/mesh-mechanism-map.yaml` (handwritten precision-first descriptor map + unmeasurable block)
- `scripts/tag_articles.py` (the `include_metadata` switch; production byte-identical)
- `tests/test_mechanism_recall.py` (36 tests: load-bearing leak-classifier + non-circularity guards + pure-logic aggregation + structural/self-consistency)
- docs: 183 → 219 Python tests; CLAUDE.md Key Files + repo-state line

## Test plan

- `pytest tests/` green (207 in `tests/` + 12 bindings = 219).
- `test_doc_test_count` enforces the 219 count in lockstep across CLAUDE.md + README.md.

Follow-up (separate PR): integrate the mechanism-recall number into the manuscript's tagging-limitations discussion, then close #412.


ELares:
## Adversarial review pass

Ran a 4-dimension adversarial review (non-circularity/correctness, production byte-identity, test quality, scientific honesty), each finding independently verified before counting.

**34 findings, 32 verified.** The substance is confirmation that the design holds:
- **Non-circularity**: `include_metadata=False` genuinely excludes MeSH/disease/gene/drug from the matched text; a record whose only mechanism signal is in MeSH is NOT tagged by the leakage-free path (proven by `test_mesh_only_signal_not_tagged`). The leak classifier reads the live `text_matches_keyword`/`MECHANISM_KEYWORDS`, so it cannot drift.
- **Production byte-identity**: the default path is unchanged; corpus `INDEX.jsonl` / `by-pmid` / `tags` are untouched by this PR.
- **Math/aggregation**: 9 measurable mechanisms, 1,366 independent-descriptor records, 90.0% volume-weighted / 87.5% macro, self-consistent.
- **Honesty**: the 9-of-~25 scope, the epigenetic 54.7% (keyword-coverage gap + reference breadth), and the not-MeSH-measurable framing (never 0%) are all caveated.

**One actionable finding, fixed** (commit 1c5b4ef2): a YAML folded-scalar line break produced "keyword- coverage" (spurious space) in the epigenetic note, propagating to the generated report/json. Reworded so no compound word straddles a line break; artifacts regenerated.

Follow-up (separate PR): cite this mechanism-recall number in the manuscript's tagging-limitations discussion, then close #412.

Issue #418, opened 2026-06-06
Expand epigenetic mechanism keywords (coverage gap surfaced by #412 recall measurement)


ELares:
**Finding (from #412).** The non-circular mechanism-recall measurement (`scripts/mechanism_recall.py`, `analysis/mechanism-recall-report.md`) puts **epigenetic** at **54.7%** recall, the lowest of the 9 measurable mechanisms (next lowest is bispecific at 86.1%).

**Diagnosis.** Spot-checking the misses (records MeSH-tagged `DNA Methylation` that the leakage-free tagger did not tag `epigenetic`) shows two compounding causes:
1. **Real keyword-coverage gap** — the field's vocabulary includes "epigenetic modifications", "epigenetic regulation", "EZH2 inhibitor", "BET inhibitor", "tazemetostat", etc., none of which are in `MECHANISM_KEYWORDS['epigenetic']` (currently: epigenetic therapy, dna methylation cancer, hdac inhibitor, histone deacetylase, epigenetic reprogramming, dnmt inhibitor, azacitidine, decitabine, vorinostat, romidepsin). Examples missed: PMID 32726977 ("Targeting Epigenetic Modifications in Uveal Melanoma"), 34795289 ("epigenetic regulation"), 37858275 ("EZH2 inhibition").
2. **Reference-scope breadth** — `DNA Methylation` (the lone independent MeSH leaf) also tags basic methylation biology that is not epigenetic *therapy*, so part of the gap is the reference being broader than the mechanism-as-therapy, not a tagger miss.

**Task.** Add the missing epigenetic-therapy terms to `MECHANISM_KEYWORDS['epigenetic']` in `scripts/config.py` (precision-aware: prefer "ezh2 inhibitor"/"bet inhibitor"/"epigenetic modification"/"epigenetic regulation" over bare "ezh2"/"bet" which co-occur broadly), then re-run `scripts/mechanism_recall.py` and confirm the epigenetic recall rises. Note: changing keywords re-tags the corpus, so coordinate with the frozen-19 manuscript results (the frozen INDEX is not regenerated by adding scaffold terms; see CLAUDE.md).

**Provenance.** Surfaced by the post-merge gap analysis of #417. Cause (1) is the actionable half; cause (2) is a measurement caveat already documented in `analysis/mesh-mechanism-map.yaml`.

ELares:
## Read-only precision/recall scan of candidate keywords (no production change yet)

[...]

PR #419, opened 2026-06-06
docs(#412): cite measured non-circular mechanism recall in the manuscript

ELares:
## What & why

Follow-up to #417, which delivered the non-circular mechanism-recall measurement (`scripts/mechanism_recall.py`). This PR delivers the remaining **manuscript** acceptance criterion of #412: the mechanism-coverage caveats now cite the (first-ever) measured mechanism recall.

## Changes (manuscript only)

- **Section 3.6 (Limitations)**: a new paragraph that
  - states the 55% figure is the *evidence-level* recall, and that the 100-article gold set is **circular for mechanism recall** (sampled from the tagger's own positives, so zero mechanism false-negatives by construction);
  - reports the non-circular MeSH-concordance result: **9 mechanisms with an independent MeSH leaf, 90.0% volume-weighted / 87.5% macro recall**;
  - flags **epigenetic (54.7%)** as a genuine keyword-coverage gap (tracked in #418) compounded by descriptor breadth;
  - notes the newest / device / regulated-cell-death modalities are **not MeSH-measurable**, never 0%;
  - draws the consequence: mechanism gap claims rest on mechanism detection, which is high (~90%) where measurable.
- **Chapter 1 overview caveat**: a one-clause cross-reference so the prominent "55% recall, absence is provisional" line no longer reads as if 55% bounds *mechanism* gap claims.

## Verification

- `v1.tex` regenerated via `scripts/generate_latex.py`; **pdflatex 2-pass compiles, RC=0, 130 pages, no fatal errors**.
- `pytest tests/` green (207 in `tests/` + 12 bindings); `test_latex_unicode_safe` and `test_manuscript_inventory` pass.

## Scope note (issue stays open)

`refs #412`, not a close: this satisfies the recall + manuscript-caveat criteria. The issue also lists mechanism **precision**, which MeSH concordance cannot give cleanly (MeSH is non-exhaustive, so true-positives lacking the descriptor would understate it) and which the issue itself scopes as a human-in-the-loop labeling task. #412 remains open to track that optional spot-check.


ELares:
## Adversarial review pass

3-dimension review (factual accuracy vs the data, internal consistency, honesty/style), each finding independently verified. 20 findings, 17 verified, almost all positive confirmations: every cited number (90.0% / 87.5% / 9 mechanisms / epigenetic 54.7% / 55%) matches `analysis/mechanism-recall.json` exactly, the circular-gold-set claim is accurate, and the new text carries no dash punctuation.

Three consistency/precision items addressed in 93c473bf:
- Chapter 10 conclusion: "55% recall" → "55% evidence-level recall" + the mechanism-level ~90% distinction.
- Glossary "Recall" entry: clarified evidence-level, cross-referencing Section 3.6.
- Section 3.6: replaced the loose "newest/device/regulated-cell-death" phrasing with the concrete not-MeSH-measurable set, so it no longer implies all recent mechanisms are unmeasurable.

Manuscript recompiles (pdflatex RC=0, 130 pp). The PDF build runs on push to main, so it re-verifies post-merge.

PR #423, opened 2026-06-07
feat(#332): data-conditioned ABC posterior on the kill switch (in-vitro)

ELares:
## What & why

#332 asks for a **data-conditioned posterior** (the prior-predictive intervals are already shipped). #330's CTRPv2 target makes conditioning possible. `scripts/abc_posterior.py` runs ABC rejection over in-vitro-spanning priors, accepting the draws closest to the CTRPv2 ML162 dose-response, posterior-predictive-checked on held-out ML210.

## The load-bearing finding (why this reframes #332)

The in-vivo PRCC prior ranges used for the prior-predictive intervals are **disjoint** from the in-vitro data:

| parameter | in-vivo PRCC range | in-vitro posterior (2.5% .. median .. 97.5%) | posterior entirely above in-vivo max? |
|---|---|---|---|
| `lp_propagation` | [0.05, 0.2] | 0.38 .. 0.79 .. 0.95 | **YES** |
| `lp_rate` | [0.03, 0.12] | 0.28 .. 0.65 .. 0.95 | **YES** |

The in-vitro GPX4i kill needs a lipid-peroxidation cascade 3-4× stronger than the in-vivo plausible ranges allow. Consequences (stated honestly):
1. The data-conditioned posterior is **in-vitro only** — credible bands on the #330 point calibration (medians track the point fit: lp_propagation 0.79, lp_rate 0.65, K 0.37, gpx4 0.43).
2. The **in-vivo / spatial headline claims cannot be conditioned on in-vitro data** and stay prior-predictive (the existing intervals). Conditioning them needs an **in-vivo** ferroptosis dose-response dataset — the remaining #332 dependency.

## Posterior-predictive check (held-out ML210)

Coverage **7/7**: the 95% posterior-predictive band brackets the held-out GPX4 inhibitor at every dose. The posterior generalizes across GPX4 inhibitors at the ensemble level.

## Scope

`refs #332` (not a close): the data-conditioned posterior is delivered for the conditionable (in-vitro single-cell) quantity, and the reason the in-vivo headlines stay prior-predictive is documented. #332 stays open to track the in-vivo-data dependency.

## Tests

`pytest tests/` green (230 in `tests/` + 12 bindings = 242). 6 new tests (prior-design guards, posterior structure/ordering, the disjunction result, held-out coverage). Deterministic (seeded); committed result re-derives.


ELares:
## Adversarial review pass

3-dimension review (ABC correctness + genuine disjunction, honesty, reproducibility), each finding independently verified by re-running. 24 findings, 22 verified, overwhelmingly positive: ABC rejection valid, the **disjunction is real and not an artifact** (posterior spans [0.38,0.95] well inside priors [0.05,1.0], so not edge-hitting), determinism confirmed across 3 runs, honesty/scoping accurate (in-vivo headlines correctly stay prior-predictive), tests CI-safe.

**One real should-fix, addressed** (9707b3c8): the held-out '7/7' coverage was tolerance-dependent. Now reports **both** strict (3/7 inside the 95% band) and tolerant (7/7 within 0.05 viability), with the report explaining the tolerance is doing real work (CTRPv2 curves are cell-line medians vs a single-cell model; the RMSE summary discards curve shape). The honest reading: the posterior-predictive band is in the right place but not tight enough to bracket every point strictly. Test + CLAUDE.md updated to match.

Issue #333, opened 2026-06-04
Validate the 3D spheroid module against multi-size spheroid structure/kill data


ELares:
**Why.** The `spheroid` zone geometry is already anchored to Browning et al. (*eLife* 2021, PMID 34842141), but the per-zone biochem gradients and the size-dependence of penetration/kill remain uncalibrated, and the `3d_validation_report.md` flags finite-size effects.

**What exists.** Browning 2021 (PMID 34842141) provides quantitative spheroid radial structure (necrotic/quiescent/proliferating zones) vs diameter; multi-size spheroid + organoid kill-curve platforms publish survival-vs-size for cytotoxic agents.

**Task.** Run sim-tme-3d across spheroid sizes (e.g. 100, 300, 500, 1000 um diameter) and compare the predicted zone fractions and depth-dependent kill against the Browning structure and any available size-resolved kill-curve data. Quantify finite-size error and the size range where predictions are trustworthy.

**Acceptance criteria.** A size-sweep validation figure + a documented "valid size range"; spheroid biochem-gradient parameters either calibrated or explicitly bounded by what the structural data can constrain. Follow-up to the spheroid layer (#197/#270, closed).

ELares:
**Update (#424 merged).** The structure half of #333 is delivered:

- **Zone-geometry validation** (`scripts/validate_spheroid_structure.py`, `analysis/calibration/spheroid-structure-validation.{md,json}`): the spheroid module's fixed radial thresholds (rim 0.90, core 0.73 of radius) validated against Browning 2021 size-resolved confocal data (994 spheroids).
- **Valid size range documented**: the fixed thresholds match only for **R >= 300 µm (diameter >= 600 µm)**. Below that, real spheroids have essentially no necrotic core, so the model over-predicts the persister/necrotic core for small spheroids (the core emerges near R ~ 300 µm, the O2 diffusion length). CALIBRATION_STATUS now records the spheroid zones as 'validated with a valid size range'.

**Remaining (why this stays open):**
- **Size-resolved depth-dependent KILL data** — the other half of the acceptance criteria. The Browning dataset is structure-only (no drug kill). A size-resolved kill-curve dataset (organoid/spheroid survival vs size for a cytotoxic/ferroptosis agent) is needed; not yet sourced.
- **Size-aware zone variant** — the model's thresholds are size-independent (the large-spheroid limiting structure). A size-aware variant (necrotic core fraction growing with diameter) is the modeling refinement the validation motivates.

[...]

PR #424, opened 2026-06-07
feat(#333): validate spheroid zone geometry vs Browning 2021 (finite-size + valid range)

ELares:
## What & why

#333 asks to validate the `spheroid` module's radial zones against size-resolved structure data and document a **valid size range**. The module uses **fixed** radial thresholds from the Browning 2021 large-spheroid *limiting* structure (rim at 0.90, core at 0.73 of radius), which are **size-independent**. Real spheroids aren't: the necrotic core appears only above a critical size.

## Data

`scripts/validate_spheroid_structure.py` fetches Browning et al. 2021 (eLife, PMID 34842141; `github.com/ap-browning/Spheroids`, `Data/AllConfocalData.csv`): **994 spheroids**, R 144-407 µm, each with the outer radius R and two internal boundaries as fractions of R (φ = proliferating-rim inner edge, η = necrotic-core outer edge). Raw not committed (offline contract); the size-binned derivative is.

## Finding (the finite-size error + valid range)

| R bin (µm) | n | φ median (model 0.90) | η median (model 0.73) | frac w/ necrotic core |
|---|---:|---:|---:|---:|
| <200 | 124 | 0.14 | 0.00 | 0.8% |
| 200-300 | 398 | 0.59 | 0.00 | 45% |
| 300-400 | 471 | 0.87 | 0.63 | 100% |
| >400 | 1 | 0.90 | 0.76 | 100% |

**The fixed thresholds match the data only for R ≥ 300 µm (diameter ≥ 600 µm).** Below that, real spheroids have essentially **no necrotic core** (η ≈ 0), but the model places a persister/necrotic-like core at 0.73 of radius regardless, so it **over-predicts the resistant core for small spheroids**. The necrotic core emerges near R ≈ 300 µm, consistent with an O₂ diffusion length of ~150-200 µm.

## Scope

- Validates zone **geometry** + a documented valid size range. Does **not** calibrate the per-zone biochem gradient *strengths* (MUFA/GSH/iron, which need metabolomics, not structural boundaries) — unchanged.
- The size-resolved depth-dependent **kill-curve** half of #333 is not in this dataset; `refs #333` (keep open) for that remaining piece.

## Tests

`pytest tests/` green (236 in `tests/` + 12 bindings = 248). 6 new tests incl. a **drift guard** tying the Python model thresholds to the Rust `glycolytic_frac`/`oxphos_frac` defaults. CALIBRATION_STATUS spheroid row → "validated with a valid size range"; PROVENANCE + CLAUDE.md updated.


ELares:
## Adversarial review pass

3-dimension review (data/finding correctness, honesty/scope, reproducibility), each finding independently verified by re-running the validation against the Browning raw data. **26 findings, all verified positive, no actionable issues:**
- **Data**: phi/eta correctly interpreted (zero phi>eta inversions, all in [0,1]); the R~300µm necrotic-core emergence is **real continuous data, not an eta=0 encoding artifact** (R<200: 0.8% have a core; 200-300: 45%; 300-400: 99.8%), consistent with the ~150-200µm O2 diffusion length.
- **Mapping**: model 0.90/0.73 radius boundaries correctly map to the Rust 0.73/0.39 volume fractions (cube roots); drift guard tolerance 1e-6.
- **Valid range**: R≥300µm is the only bin satisfying both boundary errors ≤0.15 AND core present ≥50%.
- **Honesty**: scope correctly limited to zone geometry + size range (not biochem gradient strengths, not size-resolved kill); refs #333 keeps the kill-curve half open. No em/en dashes.
- **Repro**: identical output on dual runs; raw never committed; offline contract respected.

No changes needed.

PR #425, opened 2026-06-07
feat(#411): semantic-retrieval evidence-detection leg (committed embeddings, offline CI)

ELares:
## What & why

The second leg of #346. The MeSH fallback lifted gold-set evidence-detection recall 55.2% → 67.8% but hit a hard floor (empty-MeSH / non-discriminative residue MeSH can't reach). This adds a semantic-retrieval leg: a SPECTER document-embedding **leave-one-out k-NN** that flags a record evidence-bearing when its title+abstract neighbours are predominantly evidence-bearing, combined as `keyword OR MeSH OR embedding`.

## Dependency story (your option B)

- `scripts/embed_gold_set.py` embeds the gold set with `sentence-transformers` (allenai-specter) — run **locally in a venv**, **never** added to `requirements-lock.txt`.
- The committed **vectors** (`analysis/evidence-gold-embeddings.npz`, 816 KB) + `scripts/embed_evidence_leg.py` (numpy-only k-NN) are what the eval and CI use. **CI never downloads or loads a model** — the offline/pinned-deps contract holds.

## Result (balanced v2 gold set)

Evaluated on **v2** (270 records, 32% positive / 183 negatives), not v1 (87% positive — too imbalanced; a majority-vote k-NN degenerates to always-positive there).

| stage | recall | precision |
|---|---:|---:|
| baseline (keyword) | 55.2% | 36.9% |
| + MeSH (#346) | 67.8% | 35.3% |
| **+ MeSH + embedding (k=5)** | **89.7%** | **39.8%** |

The semantic leg recovers **19 of the 28** MeSH-residual false-negatives, lifting recall 67.8% → 89.7% with **no precision drop** (the k-NN's own precision exceeds the keyword baseline's). The v2 recall reproduces the documented v1 55.2%/67.8% (shared 87 positives); v2's low *absolute* precision is from its deliberately hard negatives, not the manuscript's v1 96% — what matters for #411 is the recall lift and that precision doesn't drop.

## Scope / honesty

- Binary **detection** only (does not assign the evidence *level*).
- Off-by-default / gold-set-validated: production recall is unchanged (like the MeSH fallback). It demonstrates the achievable lift; it is not wired into the frozen corpus.
- k is a hyperparameter (sweep in the report); headline k=5. The keyword tagger remains the precision-first reproducible baseline.

## Tests

`pytest tests/` green (243 in `tests/` + 12 bindings = 255). 7 new tests incl. the residue-recovery-without-precision-drop guard. `Closes #411` (completes the second #346 leg).


ELares:
## Adversarial review pass

3-dimension review (ML/eval correctness, honesty/precision-framing, reproducibility/offline-contract), each verified by re-running. **26 findings, 25 confirmed, overwhelmingly positive:**
- **ML**: LOO k-NN self-exclusion correct (no leakage); embedding-only precision **66.2% vs 32% base rate** so the k-NN is genuinely discriminative (not trivial); v2-over-v1 choice is mathematically necessary, not cherry-picked; numbers reproduce bit-exact.
- **Honesty**: 'no precision drop' is correct (it rises +4.5pp); recall ladder matches documented v1.
- **Offline**: vectors committed (755KB), numpy-only eval, no model/sentence-transformers in CI or requirements-lock; embeddings row-normalized.

**One should-fix, addressed** (026bf7b5): the precision framing was honest but the v2-vs-v1 caveat sat *after* the headline numbers, so a skimmer could misread v2's ~37% as a regression. Restructured the report so that caveat **leads** the section ('this is NOT the 96%, and not a regression'), and stated the precision **rise** (+4.5pp) explicitly instead of 'does not drop'.

Issue #408, opened 2026-06-06
Cross-check the reaction-diffusion supply field against PhysiCell/BioFVM (external-simulator validation; #343 follow-up)


ELares:
**Context.** #343 added a steady-state reaction-diffusion supply field (`ferroptosis-core::reaction_diffusion`) as an alternative to the monotonic `exp(-dist/λ)` proxy, wired it into sim-tme-3d (`Overrides.reaction_diffusion`), and benchmarked it against the proxy (`analysis/reaction-diffusion-benchmark.md`). The solver was **verified against the closed-form 1-D analytical reaction-diffusion slab** (the published-model anchor), satisfying #343's "and/or a published tumor reaction-diffusion model" acceptance leg.

**What's still outstanding.** #343's acceptance also named an **independent-simulator** cross-check: "benchmark ... against PhysiCell (via the existing FFI)". That was honestly deferred because it requires building PhysiCell and exporting the same vessel geometry through the existing C-FFI — a heavier, separately-scoped effort. This issue tracks it.

**Task.**
1. Export a fixed 3-D vessel geometry (the benchmark's sparse + dense networks) and the same `D`, `k` (λ) to PhysiCell/BioFVM (or another independent reaction-diffusion solver).
2. Compare the external steady-state O2/drug field to `reaction_diffusion_supply_field` cell-by-cell: agreement (mean/max abs diff, correlation) and any systematic bias.
3. Confirm or correct the benchmark's qualitative conclusions, in particular:
   - the **3-D point-source geometry** dominance (a single isolated vessel's ~0.85→0.31 first-neighbour gap),
   - the **λ-regime sign flip** (depletion when λ ≲ vessel spacing, enrichment when λ ≫ it),
   - the **single-voxel vessel representation / grid-resolution** sensitivity of the magnitude (does a finite-radius vessel source in BioFVM shift it?).

**Acceptance.** A documented external-solver comparison (numbers + interpretation) appended to `analysis/reaction-diffusion-benchmark.md`, replacing the current "PhysiCell cross-check deferred" scope note; CI optionally builds/links the PhysiCell harness via the existing `ferroptosis-ffi`.

[...]

PR #426, opened 2026-06-07
feat(#408): external cross-check of the reaction-diffusion solver vs BioFVM

ELares:
## What & why

The deferred external-simulator leg of #343. The Rust `reaction_diffusion` steady-state supply field had only an analytical 1-D self-consistency check; #408 asks for an independent cross-check against PhysiCell/BioFVM. This delivers it.

The **same matched scenario** (Dirichlet vessel sources c=1, no-flux boundaries, uniform decay, λ=√(D/k)) is solved by the Rust finite-volume SOR solver **and** by **BioFVM** (PhysiCell's independent LOD/Thomas microenvironment solver, run to steady state), then compared.

## Result: the solvers agree

| BioFVM dt | shape log-r | Rust λ | BioFVM λ | ratio Rust/BioFVM |
|---|---:|---:|---:|---:|
| 0.02 | 0.9997 | 55.1 | 55.2 | 2.82 |
| 0.005 | 1.000 | 55.1 | 55.1 | 1.42 |
| 0.001 | 1.000 | 55.1 | 55.1 | **1.08** |

- **Shape: identical** (log-field Pearson r → 1.0, single- and two-source).
- **Decay length: identical** between solvers (55.1µm single / 59.0µm two-source). The Yukawa point-source shortening vs the planar λ=100 is reproduced by **both**, confirming it's real 3D geometry (the benchmark already flagged it), not a Rust-solver artifact.
- **Magnitude: converges to agreement** as BioFVM's dt → 0 (ratio 2.82 → 1.08). The residual is BioFVM's LOD operator-splitting error, a controllable numerical artifact, not a solver disagreement.
- **Two-source (multi-vessel)**: r=1.0, λ match — agreement holds in the regime the nearest-vessel proxy averages away (what #343 cares about).

## Files / offline contract

- `scripts/validate_rd_vs_biofvm.py` — local harness (runs both solvers, writes the committed `analysis/calibration/rd-biofvm-crosscheck.{md,json}`).
- `simulations/calibration/biofvm/{biofvm_rd_check.cpp,README.md}` — the BioFVM driver + build incantation. BioFVM is **not vendored** (own license); the README documents cloning it + the macOS Apple-clang workarounds (empty `omp.h` stub since BioFVM calls no `omp_` functions; explicit `-isysroot`/`-isystem` for libc++).
- `simulations/ferroptosis-core/examples/rd_field_dump.rs` — the Rust field dump (built by `cargo test`).
- **CI reads only the committed result** (BioFVM + cargo are not in CI). 5 tests guard it.

Upgrades the `reaction_diffusion` module from analytical-self-consistency-only to externally cross-checked. `Closes #408`.

## Tests

`pytest tests/` green (248 in `tests/` + 12 bindings = 260); the Rust example compiles.


ELares:
## Adversarial review pass

3-dimension review (cross-check validity, honesty, reproducibility/offline-contract), each verified against the real files (incl. running the Rust example). **25 findings, 23 confirmed, all positive, no changes needed:**
- **Validity**: the scenario is genuinely matched (same 41³ grid, h=20, λ=100 → D=10000/k=1, same source voxels, no-flux both sides, uniform decay); shape-r computed in log space over a sensible region (~10k points, source + noise excluded); the 55.1µm decay length is measured *independently by both* solvers (real 3D Yukawa geometry, not a Rust artifact).
- **Honesty**: 'the solvers agree' is accurate and qualified — shape + physics agree, the magnitude ratio converges to 1 as dt→0, and the ~8% residual at the finest dt is honestly disclosed as LOD operator-splitting error.
- **Offline**: BioFVM not vendored (driver + build README only); committed result is the sole CI input; requirements-lock unchanged; Rust example compiles; 5 regression-safe tests.

The one nit (could emphasize the multi-axis agreement) is explicitly aesthetic, not a fix. No changes needed.

(Note: a few verifier sub-agents in the workflow didn't emit structured output and were dropped; the confirmed findings stand on their own verification.)

PR #577, opened 2026-06-29
sim-tme-3d: guard + invariant test for large-grid RNG seed aliasing (#539)

### LINKED ISSUE #539: sim-tme-3d large-grid RNG seed aliasing (biochem/immune seed strides collide above ~dim 100); add a collision guard + invariant test
In `simulations/sim-tme-3d/src/main.rs` (~1589-1594, 2032-2099) the per-cell biochem seed uses stride 1e6 and the immune seed 2e6; once `n_cells` exceeds the per-step stride (onset around dim > ~100/127) seed streams can collide/alias. Determinism is preserved, so the 60³ byte-identity SHA gate (#253) is blind to it, and only `--bench` (BENCH_GRID_DIM ≥ 101, writes no output) currently exceeds the threshold — so it's a **latent foot-gun, not a live scientific defect**.

A real seed-scheme fix collides with the byte-identity lock and forces a re-baseline (maintainer call). **AI-safe interim:** add a `debug_assert!`/warn when the configured dim exceeds the collision-free stride, plus a seed-uniqueness invariant test, so a future large-grid production run can't silently alias.

---
*Filed from a 2026-06-26 outside-in contribution review (7-lens adversarial fan-out of the repo). Evidence verified against the current tree. Part of the post-#511 hardening backlog.*

ELares:
Closes #539.

### The latent foot-gun
The per-cell RNG seed within a stream is `cond_seed + offset + idx + step*STRIDE` (biochem stride `1e6`, immune `2e6`). It's injective across steps only while `idx` (0..`n_cells`) stays below the stride — cell `STRIDE` at step *s* otherwise reuses cell 0 at step *s+1*'s seed. So above ~1e6 cells (**dim > 100** biochem / ~127 immune) the streams **alias across steps**. Determinism is preserved, so the #253 byte-identity SHA gate is blind to it, and only `--bench` (`BENCH_GRID_DIM ≥ 101`) currently exceeds the threshold. **Latent foot-gun, not a live defect** — the production matrix is dim 60 (216k cells).

### AI-safe interim (this PR)
The real fix — a position+step-independent seed scheme — changes the matrix bytes and needs a **maintainer-approved #253 re-baseline**, so it's deliberately *not* done here. This PR ships the guard the issue asks for:
- Named strides `BIOCHEM_SEED_STRIDE`/`IMMUNE_SEED_STRIDE` + `SEED_COLLISION_FREE_MAX_CELLS` (1e6), replacing the 3 magic stride literals — **byte-identical** (the constants equal the former literals).
- `warn_if_seed_aliasing(n_cells)`, called once when the grid is built: warns loudly (and `debug_assert!` in debug) above the collision-free max, so a future large-grid run can't silently reuse per-cell draws.
- An **invariant test** pinning the collision-free boundary, the aliasing onset (dim 100 safe / dim 101 aliases), the guard threshold, and dim-60 safety.

### Byte-identity
The golden kill-count regression (`constant_path_golden_kill_counts`) passes **unchanged** locally; the guard is a no-op at the dim-60 matrix and every shipped condition. The deeper seed-scheme re-baseline is tracked as a maintainer-gated follow-up.


ELares:
## Adversarial self-review (fresh eyes) — verdict: mergeable, findings addressed

Independent pass focused on **not changing byte-identity** (sim-tme-3d is #253-SHA-locked).

**Verified clean:**
- **Byte-identity:** the 3 stride-literal replacements are exactly equal to the originals (`BIOCHEM_SEED_STRIDE=1_000_000` at the biochem site; `IMMUNE_SEED_STRIDE=2_000_000` at both immune sites); the const offsets are unchanged. The diff contains only new consts/fns, the 3 renames, the one guard call, and the test — nothing alters any RNG draw on the dim-60 path. The **golden kill-count regression passes unchanged** (no matrix drift).
- **Threshold math (no off-by-one):** `seed_collision_free(n) = n <= 1e6` is correct — the seed is injective across steps iff max `idx` (n_cells−1) < stride. Test pins it: dim 100 (1e6) safe, dim 101 (1_030_301) aliases — matches #539 exactly.
- **No behavior change on shipped paths:** on `n_cells <= 1e6` the guard executes nothing (no eprintln, no debug_assert, no RNG). `debug_assert!(false)` lives strictly in the aliasing branch; the largest test grid is dim 60 (216k ≪ 1e6), so `cargo test` can't panic from it.
- Guard placed after `n_cells = grid.cells.len()`, correct for slab/spheroid/default grids; biochem (1e6) correctly binds before immune (2e6).

**Findings addressed (59a5e634):**
1. `assert!(IMMUNE_SEED_STRIDE > BIOCHEM_SEED_STRIDE)` tripped `clippy::assertions_on_constants` → replaced with a module-level **compile-time** `const _: () = assert!(...)` (clippy-clean, stronger guarantee). (clippy isn't CI-gated here, but fixed anyway.)
2. "One-time guard" doc wording → "Per-condition guard" (it runs once per condition).

Merging once the full Rust CI suite is green.

PR #597, opened 2026-06-29
CI: partition the slow sim-tme-3d suite into a parallel job (#549)

### LINKED ISSUE #549: Speed up the ~25-min Rust CI by partitioning the slow sim-tme-3d integration tests
Observed while landing PR #547: the `cargo test --workspace` CI job (and local) takes **~25 min**, dominated by one integration suite — locally measured at **960s for 104 tests** (the sim-tme-3d / heavy spatial tests). The other ~Nx test binaries finish in well under a second each.

This slows the contribution loop for every Rust PR. Options:
- Split the heavy sim-tme-3d integration tests into a separate CI job (or behind `#[ignore]` run on a nightly/weekly schedule, alongside the existing `sim-tme-3d-regression.yml`), so the fast unit suite gives sub-2-min feedback per PR.
- Or gate the heavy suite to run only when `simulations/sim-tme-3d/**` changes.
- Cache is already in place (`Swatinem/rust-cache`); the cost is runtime, not compile.

Low priority / DX, not correctness. Keep the full suite as a required gate; just front-load the fast signal.

ELares:
Closes #549.

## Problem
The Rust `test` job ran `cargo test --workspace` serially. The ~50 heavy sim-tme-3d 3D spatial A/B integration tests (full-grid runs, ~16 min) gated the entire suite even though every other crate finishes in ~9 min — a ~25-min wall clock.

## Change
Split the suite so the slow half runs in **parallel** instead of serially:
- **`test`** — `cargo test --workspace --exclude sim-tme-3d` + the unchanged C-FFI ABI verify. Keeps the existing required-check name.
- **`test-sim-tme-3d`** (new) — `cargo test -p sim-tme-3d`, mirroring `test`'s ubuntu-latest + `@stable` toolchain + `Swatinem/rust-cache@v2` workspace cache.

## Coverage is preserved exactly
`--exclude sim-tme-3d` skips only sim-tme-3d's own test targets; `-p sim-tme-3d` runs only those. Together they cover the full workspace with **no double-run** of tests. ferroptosis-core (a sim-tme-3d dependency) is recompiled in both jobs — `rust-cache` keys per-job so the caches don't collide — but its tests run only in `test`. sim-tme-3d's unit + integration + **doc** tests all run in the new job (`cargo test -p`, unlike a nextest split, keeps doctests).

## Effect
Wall clock drops from the serial sum (~25 min) toward `~max(test, test-sim-tme-3d)` ≈ the sim-tme-3d floor (~16 min). This PR itself exercises the new split (it touches `cargo-test.yml`, which is in the path filter). `sim-tme-3d` confirmed as the exact workspace package name; MANIFEST regenerated for the single workflow-file hash change.


ELares:
## Adversarial self-review (fresh eyes) — VERDICT: SAFE (one NIT resolved below)

Reviewed for the load-bearing risk of a test-suite partition: does any test that ran before now run in **neither** job (silent coverage loss)?

**A. No execution-coverage loss — verified.** 14 workspace members; `simulations/Cargo.toml` has **no `default-members` and no `exclude`**, so `--workspace` selects all 14, `--exclude sim-tme-3d` removes exactly one, and `(--workspace --exclude sim-tme-3d) ∪ (-p sim-tme-3d)` = the full set by construction. sim-tme-3d is **binary-only** (no `lib.rs`, no `sim-tme-3d/tests/`), so its 105 `#[test]` are all `#[cfg(test)]` unit tests inside the bin target → all caught by `cargo test -p sim-tme-3d`; it has no integration tests and no doctests to lose. The only `tests/` dir in the workspace is `ferroptosis-ffi/tests/` (just `test_ffi.c`), and ferroptosis-ffi isn't excluded.

**B. No harmful double-run.** `-p sim-tme-3d` runs only sim-tme-3d's own targets, not its deps' — so ferroptosis-core's ~370 tests run only in `test`. sim-tme-3d doesn't depend on ferroptosis-ffi, so the new job never builds the FFI cdylib, never regenerates `ferroptosis.h`, never runs the header git-diff guard — those stay solely in `test`.

**C. rust-cache independent.** Both jobs use `Swatinem/rust-cache@v2` with no shared key; v2 folds `github.job` into the cache key, so `test` and `test-sim-tme-3d` get distinct caches — no race/poisoning.

**E. Gate-side NIT — RESOLVED (not applicable here).** The reviewer correctly flagged that if `main` required the `test` status check, the 105 sim-tme-3d tests (now in the new job) would no longer *block* merge unless `test-sim-tme-3d` were added to the required list. Checked the actual config: `main` has **`required_status_checks: None`** (classic) and **0 rulesets** with no `required_status_checks` rule — status checks are not a hard merge gate on this repo at all. The `BLOCKED` state comes from `required_pull_request_reviews: True` (with `enforce_admins: False`, hence admin merge). So there is no required-checks list to update; all checks are advisory and verified green before merge.

## Observed speedup (this PR's own run exercises the split)
- `test` (workspace minus sim-tme-3d): **~87 s** (cache-warm)
- `test-sim-tme-3d`: the long pole, now parallel
- build-macos ~48 s, fmt ~30 s

Rust-Tests wall clock is now `max(jobs)` ≈ the sim-tme-3d job alone, vs the old serial sum — the non-sim-tme-3d suite (87 s) no longer waits behind it.

Issue #578, opened 2026-06-29
sim-tme-3d: replace additive per-cell RNG seed scheme to eliminate large-grid aliasing (needs #253 re-baseline)


ELares:
## Background
#539 (closed by #577) added the **AI-safe interim**: named seed strides, a `warn_if_seed_aliasing` guard (warn + `debug_assert!` above the collision-free max), and an invariant test. That makes a large-grid run *fail loudly* instead of silently aliasing — but it does **not** fix the underlying scheme.

## The real defect
The per-cell RNG seed in `simulations/sim-tme-3d/src/main.rs` is `cond_seed + offset + idx + step*STRIDE` (biochem stride `1e6`, immune `2e6`). Because `idx` and `step` enter the seed *additively* with a fixed stride, the stream is injective across steps only while `idx` (0..`n_cells`) stays below the stride. Above ~1e6 cells (**dim > 100** biochem / ~127 immune) cell `STRIDE` at step *s* reuses cell 0 at step *s+1*'s seed → per-cell stochastic draws repeat. Only `--bench` (`BENCH_GRID_DIM ≥ 101`) currently exceeds the threshold; the production matrix (dim 60) is safe, so this is latent, not live.

## Proposed real fix
Replace the additive scheme with a **position+step-independent hash mix** so every `(idx, step)` maps to a well-distributed unique seed regardless of grid size, e.g.:
```rust
// SplitMix64-style finalizer; combine cond_seed, idx, step by hashing, not adding.
fn cell_seed(cond_seed: u64, idx: usize, step: usize, salt: u64) -> u64 {
    let mut z = cond_seed ^ salt
        ^ splitmix64(idx as u64).rotate_left(17)
        ^ splitmix64(step as u64).rotate_left(43);
    splitmix64(z)   // final avalanche
}
```
(biochem vs immune distinguished by `salt`, replacing the 500_000 / 900_000_000 offsets). This removes the stride entirely, so there is no large-grid collision regime.

## Why this needs a MAINTAINER decision (the blocker)
Changing the seed mix **changes every per-cell RNG draw**, which **changes the production-matrix `summary.json` bytes** → the #253 byte-identity SHA gate (`sim-tme-3d/expected_summary.sha256`, `sim-tme-3d-regression.yml`) will **fail by design**. Landing this therefore requires a deliberate, maintainer-approved **re-baseline**:
[...]

PR #598, opened 2026-06-29
CI: optimize the test profile so sim-tme-3d stops dominating the ~25-min suite (#549)

ELares:
Reopened #549. Follow-up to #597.

## Why #597 wasn't enough
#597 partitioned sim-tme-3d into its own parallel job. The merged run proved that gives faster *common-case* signal but **does not cut the wall-clock**:
- `test` (workspace minus sim-tme-3d, debug): **87 s**
- `test-sim-tme-3d` (debug): **~25 min**

sim-tme-3d *alone* ≈ the whole old suite.

## The real diagnosis (measured)
The bottleneck is that `cargo test` runs **unoptimized** (opt-level 0). The compute-heavy 60³×180 spatial integration tests are ~190× slower without optimization:

| | one heavy 60³ test | full `cargo test --workspace` (sim-tme-3d) |
|---|---|---|
| debug (opt 0, current) | **83.1 s** | ~25 min |
| **opt-level 3 (this PR)** | **2.7 s** | **7.7 s**, 105/105 pass |

## The fix
```toml
[profile.test]
opt-level = 3
```
- **Keeps the safety checks**: only `opt-level` is overridden, so `debug-assertions` + `overflow-checks` stay ON (test-profile defaults). The 14 sim-tme-3d `debug_assert!` invariant checks still fire — unlike a blanket `--release`.
- **Verified safe**: full `cargo test --workspace` at opt-3 → sim-tme-3d 105/105, every other crate, and all 4 doctests pass. opt-3 breaks no golden (the kill-count goldens are discrete and robust to opt-level FP).
- **Consistency bonus**: the per-PR golden unit tests now run at the same optimization as the release production SHA run (#253/#499), instead of diverging on opt-level-dependent FP.
- **Narrow scope**: only `cargo test` is affected. `cargo build` / `cargo run --release` (the `build-macos` job, the #253 regression) are unchanged. Costs slightly more test-profile compile time, cached by `rust-cache` in CI.

This PR's own `test-sim-tme-3d` job demonstrates the win (should drop from ~25 min to a few compile-bound minutes).


ELares:
## Adversarial self-review (fresh eyes) — VERDICT: SAFE TO MERGE

The one concern worth rigor — does opt-3 change FP results (vectorization/FMA), possibly platform-dependently, and break the kill-count goldens? **Refuted on Rust semantics + repo evidence + this PR's own CI:**

- **rustc never emits fast-math** (`contract`/`reassoc`/`nsz`) at any opt-level. Confirmed no `mul_add` in `sim-tme-3d/src` or `ferroptosis-core/src`, and the `.sum()` reductions are order-preserving left-folds. So opt-0 and opt-3 are **bit-identical f64 on a fixed platform** — the goldens were baselined under opt-0 (`cargo test` default) in #578 and still pass at opt-3 (proof the bits didn't move).
- **Cross-platform (macOS-aarch64 vs ubuntu-x86_64) at opt-3** is already discharged by the repo's own record: `expected_summary.sha256` documents the full 60³×180 **release (opt-3)** `summary.json` byte-identical across both platforms on the pinned toolchain. Discrete integer kill counts are even more robust than that full-precision serialization.
- **Empirical**: this PR's `test-sim-tme-3d` job passed at opt-3 on ubuntu-x86_64 — and ran in **~84 s vs the prior ~25 min**.

Also verified: `debug-assertions` + `overflow-checks` stay ON (only `opt-level` overridden; two `#[cfg(debug_assertions)]` `#[should_panic]` tests would catch an accidental disable); `[profile.test]` doesn't touch `[profile.release]`/`[profile.dev]`, so `build-macos` and the #253 production SHA are byte-for-byte unchanged (production is *already* opt-3 via `--release`); no opt-0-dependent test (overflow-checks on at both levels; the only wrapping arithmetic is explicit SplitMix64 `.wrapping_*`); no `unsafe` in the hot crates so no UB-miscompile surface.

**Nits addressed/noted:** fixed an inaccurate comment ("14" → dropped; the live count is 7 `debug_assert!` invocations). DX tradeoff (workspace-wide opt-3 slows local test-profile compile) accepted as simpler than per-package scoping and cached by rust-cache. Pre-existing latent note (CI `cargo test` uses `@stable` while goldens were baselined on `@1.96.0`) is orthogonal to opt-level and out of scope here.

Issue #564, opened 2026-06-27
cargo-test path filter is too broad: data/doc-only changes under simulations/ trigger the full ~25-min Rust suite


ELares:
Observed repeatedly while landing the post-#511 backlog (PR #560): editing `simulations/calibration/targets.yaml` — a pure data/citation file with no Rust impact — triggers `cargo-test.yml` (`paths: simulations/**`), which runs the full `cargo test --workspace` (cold-build + the ~960s sim-tme-3d suite, ~25-40 min). Same for any `simulations/**/*.md`, `simulations/calibration/*.json`, or `simulations/.cargo/config.toml` edit.

This added ~70 min of avoidable CI to a citation-text PR and slows the contribution loop generally.

**Fix:** tighten the `cargo-test.yml` path filter to the files that actually affect the Rust build — e.g.
```yaml
paths:
  - 'simulations/**/*.rs'
  - 'simulations/**/Cargo.toml'
  - 'simulations/**/Cargo.lock'
  - 'simulations/.cargo/**'
  - 'simulations/rust-toolchain.toml'
  - '.github/workflows/cargo-test.yml'
```
(keep `.cargo/**` so the #521 link-config stays guarded). Data/doc-only changes under `simulations/calibration/` would then be covered by `python-test` (which already re-runs the calibration-regression gate) + the manifest gate, not a Rust rebuild.

Retrospective from the 2026-06-26 outside-in review work (#546).

Issue #521, opened 2026-06-26
macOS build wall: `cargo build --release` fails on Apple Silicon (pyo3 cdylib has no `-undefined dynamic_lookup`)


ELares:
The documented first build step in `README.md` and `CONTRIBUTING.md` — `cd simulations && cargo build --release` — **hard-fails on macOS arm64**:

```
ld: symbol(s) not found for architecture arm64  (__Py_TrueStruct, _PyModule_Create2, ...)
error: could not compile `ferroptosis-python`
```

**Cause:** `ferroptosis-python` is a default workspace member (`simulations/Cargo.toml`) with `crate-type = ["cdylib"]` + `pyo3` `extension-module`. On macOS the extension-module build expects the Python symbols to be resolved at load time, but a plain `cargo build` doesn't pass `-undefined dynamic_lookup`, so the link fails. `maturin` passes it; plain `cargo build` does not, and there is no `.cargo/config.toml`.

**Why it's invisible:** CI is Ubuntu-only (`cargo-test.yml`), where the cdylib links. The `python-test.yml` macOS job is `if: github.event_name == 'schedule'`, so PRs get **zero** macOS coverage. A new contributor on a Mac (the maintainer's own platform) hits a wall on step one.

**Fix:** add `simulations/.cargo/config.toml`:
```toml
[target.aarch64-apple-darwin]
rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
[target.x86_64-apple-darwin]
rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
```
(Verified locally: `cargo build --release --workspace` → exit 0, `libferroptosis_core.dylib` produced.) Also add a per-PR macOS `cargo build --release` leg to `cargo-test.yml` so this can't regress. Do **not** just drop the member from default-members — `cargo test --workspace` still force-links it.

---
*Filed from a 2026-06-26 outside-in contribution review (7-lens adversarial fan-out of the repo). Evidence verified against the current tree. Part of the post-#511 hardening backlog.*

PR #599, opened 2026-06-29
CI: build + test the ferroptosis-python maturin/pyo3 bindings on macOS (#550)

### LINKED ISSUE #550: Extend per-PR macOS CI to the maturin/PyO3 bindings build (the half #521 didn't cover)
Follow-up from PR #547 (#521). That PR added a per-PR `build-macos` job that runs `cargo build --release --workspace` on Apple Silicon, which now covers the bare-`cargo build` link path. But the **maturin / `maturin develop` bindings build** on macOS is still only exercised by `python-test.yml`'s `test-macos` job, which is `if: github.event_name == 'schedule'` (weekly) — so the documented `maturin develop -m ferroptosis-python/Cargo.toml --release` path has no per-PR macOS signal.

Add a per-PR macOS leg that builds the bindings via maturin and runs `python -c "import ferroptosis_core; ..."` (the 12 binding tests counted in the 334 total), so a PyO3/maturin regression on macOS is caught on the PR, not a week later. Keep it build-light to bound CI minutes.

ELares:
Closes #550.

## Gap
#521 fixed `cargo build` on Apple Silicon and `build-macos` guards it. But **the actual Python packaging path had zero per-PR coverage**:
- `build-macos` runs `cargo build --release --workspace` — it links the `ferroptosis-python` cdylib but never runs maturin's wheel build or imports the extension module.
- `python-test.yml` runs only `tests/` (the corpus pipeline) — never `simulations/ferroptosis-python/test_bindings.py`.

So the maturin/pyo3 half of the macOS surface (interpreter link, wheel tags, module naming) and the 12 binding tests were untested on every PR.

## Change
New **`bindings-macos`** job (macos-latest):
```
maturin develop --release    # build wheel + install ferroptosis_core editable into a venv
python test_bindings.py      # 12 self-contained binding tests
```
A macOS-specific maturin/pyo3 regression now fails the PR — mirroring how `test` builds **and runs** the C-FFI harness rather than only compiling it. Verified locally on Apple Silicon: maturin build + **12/12 tests pass**.

Also extends the cargo-test path filter to fire on the bindings' packaging config + tests (`pyproject.toml`, `test_bindings.py`), so a change to either triggers this job — which **also closes the #564-review NIT** (`test_bindings.py` previously triggered neither workflow per-PR).

## Scope
Additive — a new parallel job + two path-filter entries. No existing job, the #253 regression, or the production build changes. MANIFEST regenerated for the single workflow-file hash.


ELares:
## Adversarial self-review (fresh eyes) — VERDICT: SAFE (one SHOULD-FIX applied)

Verified the new `bindings-macos` job works on the CI runner — and it did (`bindings-macos: SUCCESS` on macos-latest):
- **Mechanics correct**: `source` in a single `run:` block keeps the venv active across the maturin/python lines (default macOS shell is `bash -eo pipefail`, so a failing `maturin develop`/`test_bindings.py` fails the step); `working-directory` has `pyproject.toml`+`Cargo.toml`+`README.md`; module name agrees three ways (`pyproject module-name` = `[lib] name` = `#[pymodule] ferroptosis_core`) so `import ferroptosis_core` resolves.
- **No flakiness**: test_bindings.py imports only `ferroptosis_core`+stdlib; the `__main__` runner propagates a failed assert to a non-zero exit; `sim_batch` is bit-reproducible (position-independent per-cell RNG, fixed-order reduction, integer count ratio) so the seeded `0.3<death_rate<0.5` / `>0.95` assertions are deterministic with wide margins.
- **Purely additive**: #549's `[profile.test] opt-level=3` doesn't touch maturin's `--release`; no existing job, the #253 regression, or the production build changes. `.cargo/config.toml`'s `-undefined dynamic_lookup` composes with maturin's link args (a duplicate is benign — already proven by the green run).
- **Path filter** correct and comprehensive (the job also fires on the crate's `*.rs`/`Cargo.toml`/`Cargo.lock`/`.cargo/**` via existing globs).

**SHOULD-FIX applied:** maturin was installed unpinned (`--upgrade maturin`), and `maturin develop` (CLI) does NOT read pyproject's `build-system.requires`, so a future maturin 2.x could silently break the job — against the repo's pinned-deps convention. Pinned to `"maturin>=1.7,<2.0"` to match the pyproject constraint. Re-running CI on the amended commit.

Issue #585, opened 2026-06-29
Port the #578 SplitMix64 hash-mix seed to the other sim binaries that still carry the additive aliasing scheme (incl. sim-tme, the 2D manuscript engine)


ELares:
Found by a post-#578 multi-agent audit (adversarially verified, high confidence).

## Root cause
The exact additive per-cell RNG seed scheme `seed + offset + idx + step*stride` that #578 (PR #582) just removed from `sim-tme-3d` **still survives in five other binaries**. It aliases per-cell RNG streams across steps/conditions once `n_cells` exceeds the stride. The fix is a mechanical port of the `cell_seed(cond_seed, idx, step, salt)` SplitMix64 hash-mix (with BIOCHEM/IMMUNE salts) + the collision-free invariant test from `sim-tme-3d`.

## Affected sites (ranked by exposure)
| Binary | Sites | Scheme | Status today |
|---|---|---|---|
| **`sim-tme` ⭐** | `src/main.rs:218-222,341-345,543-547` (biochem stride 1e6), `640-644` (immune stride 2e6) | `seed+500_000+idx+step*1e6` / `seed+900_000_000+idx+step*2e6`, `idx=r*cols+c` | **NOT corrupted** at `GRID_SIZE=500` (250k cells, first onset 500k); `GRID_SIZE` is a compile-time const so latent-not-live. **This is the manuscript's quantitative 2D engine** — the highest-value fix. |
| **`sim-spatial`** | `src/main.rs:140-144` (biochem stride 1e6) | same additive scheme | grid is a **CLI arg** (`--grid-size`, default 500). `--grid-size 708`+ trips init-vs-step0 collision; `--grid-size 1000`+ trips within-biochem cross-step aliasing → **user-reachable**. |
| `sim-combo` | `src/main.rs:80-89` | `+ i*2 + delay_days*1e6` (condition-stride) | nit — cross-condition alias only when n_cells>500k (default 50k, 10× margin) |
| `sim-window` | `src/main.rs:77-86` | `+ i*2 + hours*1e6` | nit — default 100k, safe |
| `sim-tumor-pk` | `src/main.rs:47-56` | gen vs sim streams 500k apart | enhancement — `const N_CELLS=10_000` is 50× below threshold; cheapest alt is `const _: () = assert!(N_CELLS < 500_000)` |

## ⚠️ Manuscript re-baseline implication (the maintainer decision)
[...]

PR #600, opened 2026-06-30
#585: port SplitMix64 seed to the two no-cascade binaries (sim-combo, sim-tumor-pk)

ELares:
Partial #585 (scope confirmed with the maintainer).

## Scope decision
A full blast-radius scan showed the 5 additive-seed binaries split into tiers:
- **sim-combo, sim-tumor-pk** — feed **no** manuscript figure, fixture, or guard; their outputs are gitignored. Porting them is pure robustness with **zero** re-baseline cost. ← **this PR**
- **sim-tme** (flagship Figure 24), **sim-spatial** (Fig 8), **sim-window** (Fig 23) — re-seeding cascades into figures / fixtures / exact-integer CI guards / manuscript prose / the draft PREREGISTRATION (~30 locations for sim-tme alone), for **zero correctness gain** (grids are below the aliasing onset, so the numbers aren't corrupted). The maintainer chose to keep these on the guarded additive scheme.

## Change
Mirrors the #578 sim-tme-3d implementation verbatim — `splitmix64` + `cell_seed(cond_seed, idx, step, salt)` — which fully decorrelates the (condition, cell, stream) triple, so there is no aliasing regime at any `n_cells`:
- **sim-combo**: `cond_seed` encodes only `delay_days` (not `tx`/`with_pd1`), preserving the deliberate paired design; the two interleaved streams are separated by **salt** instead of a `+1` offset.
- **sim-tumor-pk**: `cond_seed = SEED` per scenario (as before); gen vs sim streams separated by salt, not the fixed `+500k` offset. The shadowing local `cell_seed` variable is removed.

Both now-obsolete #585 aliasing guards are removed (the hash mix eliminates the regime). Each binary gains a no-additive-aliasing invariant test probing the old 500k danger zone. Builds + tests + fmt clean; both run end-to-end. No committed artifact changes (MANIFEST updated for the two source hashes only).

#585 stays open after this for the documented sim-tme/sim-spatial/sim-window guarded-scheme decision.


ELares:
## Adversarial self-review (fresh eyes) — VERDICT: SAFE TO MERGE

- **Faithful port**: `splitmix64` + `cell_seed` are `diff`-confirmed **byte-identical** to the #578 sim-tme-3d reference (all three magic constants, the 30/27/31 xorshifts, `rotate_left(17)`/`(43)`, the XOR-mix).
- **cond_seed mappings correct**: sim-combo's `cond_seed` depends only on `delay_days` (computed outside the par_iter), so the deliberate paired design is preserved (a fixed delay reuses the same cells + noise across treatments); streams split by salt, not `i*2`/`i*2+1`. sim-tumor-pk uses `cond_seed = SEED` per scenario; gen/sim split by salt, not the old `+500k`; the shadowing local `cell_seed` is fully removed.
- **Aliasing eliminated as a hard guarantee**: splitmix64 is a bijection on u64, so for fixed `(cond_seed, step, salt)` the map `idx → cell_seed` is injective — **zero** within-stream collisions at any n_cells (the old `2*(i-i')=stride` foot-gun is gone). Both salt pairs distinct + compile-time asserted.
- **Determinism preserved**: position-independent per-cell seeding; parallel output == serial.
- **Scope claim independently verified**: `rg` across tests/analysis/article/FIGURES.yaml/scripts/.github found **zero** dependency on either binary's output (Fig 22 uses sim-combo-**mech**, a different crate; the PK calibration regression loads the Python script's `pk-calibration.json`, not the binary; both output dirs gitignored). So zero re-baseline cost, exactly as claimed.
- Guards cleanly removed, invariant tests probe the exact old 500k danger zone, MANIFEST hashes recomputed and match.

CI all green (test / test-sim-tme-3d / build-macos / bindings-macos / fmt / manifest / CodeQL). Merging.

PR #601, opened 2026-06-30
[prototyping] experimenting with adding nix, would love some opinions on this

Vaarij:
## What this changes

Not in response to an issue, but something I noticed while contributing was the separate package managers for Python and Rust, and I thought we might be able to use Nix to consolidate packages and ensure reproducibility. Nix was selected because it seems like the easiest way to combine and build with, and its package management is crucial to ensuring that one environment is similar to another. Since this codebase has a rich test suite, if tests can run in a Nix-managed environment on one local machine, it would be reasonable to assume they can run on another.

Some design considerations. As of right now, adding Nix as the default package would hinder further contributions by removing the lightweight contributions that are necessary. That is why I added Nix as a side, so it's still possible to build without Nix. I am open to changing or removing Nix as well. 

I have also been experimenting with UV, which, in my opinion, makes this PR a bit bloated and thus could be split into multiple  PRs. From my experience, I have seen that uv is a better option for packages than pip because of its faster installs and syncs. Since this repo is split into multiple groups, UV is also preferable for its ability to handle different groups within a single codebase. Some cons that stand out: UV is not as universal as pip, so this would be unfamiliar and would require some bigger changes to `Contributing.md`. Furthermore, lock file generation can be overly strict, hindering development speed.

- [ X] Tests pass locally (`python3 -m pytest tests/ -q` and, for Rust changes,
      `cd simulations && cargo test --workspace`).
- [X ] For Rust changes, `cargo fmt --all --check` is clean.

Anything that needs context: this PR is still a WIP, and I welcome any comments, concerns, or changes. This is based purely on my experience, so I would love to learn if I have overlooked a perspective or angle with my changes.

### DIFF
--- a/.github/workflows/analysis-freshness.yml
+++ b/.github/workflows/analysis-freshness.yml
@@ -46,14 +46,20 @@ jobs:
         with:
           python-version: '3.12'
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+
+      - name: Resolve lockfile
+        run: uv lock
+
       - name: Install dependencies
-        run: pip install -r requirements-lock.txt
+        run: uv sync --frozen --no-install-project
 
       - name: Regenerate derived analysis artifacts and verify they are current
         run: |
-          python3 scripts/analyze_corpus.py
-          python3 scripts/oa_bias_analysis.py
+          uv run python scripts/analyze_corpus.py
+          uv run python scripts/oa_bias_analysis.py
           if ! git diff --exit-code -- analysis/; then
-            echo "::error::Committed analysis/ outputs are stale. Run: python3 scripts/analyze_corpus.py && python3 scripts/oa_bias_analysis.py && git add analysis/"
+            echo "::error::Committed analysis/ outputs are stale. Run: uv run python scripts/analyze_corpus.py && uv run python scripts/oa_bias_analysis.py && git add analysis/"
             exit 1
           fi
--- a/.github/workflows/living-review.yml
+++ b/.github/workflows/living-review.yml
@@ -37,8 +37,14 @@ jobs:
         with:
           python-version: '3.12'
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+
+      - name: Resolve lockfile
+        run: uv lock
+
       - name: Install dependencies
-        run: pip install -r requirements-lock.txt
+        run: uv sync --frozen --no-install-project
 
       - name: Resolve trailing window
         id: window
@@ -56,7 +62,7 @@ jobs:
           # limit; the script works without it at the lower anonymous rate.
           NCBI_API_KEY: ${{ secrets.NCBI_API_KEY }}
         run: |
-          python scripts/living_review_update.py \
+          uv run python scripts/living_review_update.py \
             --since "${{ steps.window.outputs.since }}" \
             --max "${{ github.event.inputs.max || '200' }}"
 
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -15,6 +15,8 @@ on:
       # the calibration suite still gets a per-PR signal (closes the coverage gap
       # the #564 issue assumed python-test already covered):
       - 'simulations/calibration/**'
+      - 'pyproject.toml'
+      - 'uv.lock'
       - 'requirements*.txt'
       - '.github/workflows/python-test.yml'
   pull_request:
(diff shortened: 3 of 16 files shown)


ELares:
Thanks for this, and for framing it as a WIP with an open invitation for feedback — that's exactly the right way to prototype something this cross-cutting. I did a deep review across the CI, packaging, container, and Nix surfaces. Sharing findings below rather than approving, since this needs a few changes and (per your own note) probably wants to be split up.

First, the reassuring part: I looked hard at the supply-chain surface, and nothing here looks adversarial. `uv.lock` is fully hash-pinned (every artifact resolves to PyPI, no custom index, no git/editable/path deps, no install-time hooks), there are no typosquatted or surprise packages, no workflow uses `pull_request_target`/`workflow_run` or adds elevated permissions, and there's no injected content anywhere. The issues below are robustness/reproducibility, not security incidents.

## Priority

**1. `uv lock` runs right before `uv sync --frozen` in all 4 CI jobs + the Dockerfile — so the committed `uv.lock` is never actually enforced.**
Re-resolving the lock on every run means `--frozen` installs whatever `uv lock` just wrote, not the reviewed lockfile. (Note `--frozen` doesn't error on drift by itself — that's `--locked`; `--frozen` just installs the lock as-is.) Two concrete consequences:
- If someone edits `pyproject.toml` without regenerating `uv.lock`, CI silently reconciles instead of flagging it. There's no `uv lock --locked`/`--check` or `git diff uv.lock` gate anywhere.
- If a locked version is yanked from PyPI, `uv lock` re-resolves that sub-graph to a different in-range version whose hash was never reviewed, and the sync installs it.

Suggested shape: drop the `uv lock` step, use `uv sync --frozen` (or `--locked`) against the committed lock, and add `uv lock --check` as a standalone drift lint. This is the main thing — right now it quietly undoes the reproducibility the lockfile exists to provide.

**2. The Nix shell isn't reproducible as written.**
`flake.nix` pins none of its inputs (`nixpkgs` → `nixos-unstable`, `flake-utils`/`rust-overlay` → branch HEAD) and there's no committed `flake.lock`; `shell.nix` resolves via ambient `import <nixpkgs> {}`. Two `nix develop` runs days apart give different Python/Rust/LaTeX/graphviz toolchains, yet README/CONTRIBUTING advertise it as "reproducible." Either commit `flake.lock` (and point `shell.nix` at it) or soften the wording.

**3. The Dockerfile installs the resolver itself unpinned:** `pip install --no-cache-dir uv maturin`. The tool that resolves the whole environment is then whatever PyPI serves at build time. Pin `uv==` and `maturin==`.

## Worth fixing

**4. `astral-sh/setup-uv@v6` is a mutable tag** in `living-review.yml`, which is the one workflow that exposes `NCBI_API_KEY`. It matches the repo's existing convention (`actions/checkout@v7` etc. are also floating) and that workflow only runs on schedule/dispatch, so there's no fork vector — but SHA-pinning it would be good hardening for the secret-bearing job.

**5. The `requirements*.txt` exports can silently drift.** `make export-python` is manual and unverified by CI, unlike `analysis-freshness.yml`/`manifest` which regenerate-then-`git diff --exit-code`. And `export-python` only regenerates `requirements-lock.txt`/`requirements-dashboard.txt` — `requirements.txt` is still hand-edited, so four files now describe one environment with no automated sync. Add a `git diff` gate, or drop the exports if nothing consumes them.

**6. Dead dependency:** `tomli>=2.0.1; python_version < '3.11'` under `requires-python = ">=3.12"` is unsatisfiable (3.11+ ships `tomllib`), duplicated across `pyproject.toml`, `requirements.txt`, and `uv.lock`. Remove it from all three.

## Minor

**7.** The 3-step uv bootstrap block (`setup-uv` / `uv lock` / `uv sync`) is copy-pasted across 4 jobs in 3 workflows — a composite action under `.github/actions/` would keep them from drifting when #1 is fixed.

**8.** `--max "${{ github.event.inputs.max }}"` is interpolated straight into a `run:` block in `living-review.yml`. It pre-exists this PR (only the `uv run` prefix changed nearby) and is reachable only by write-access dispatchers, so it's low-risk, but since the rewrite touched this block it'd be a good moment to move the input into an `env:` var.

## Overall

The uv migration has real merit — a cross-platform, hash-pinned lock plus faster CI is a genuine step up from `pip freeze` on one machine — but as wired, finding #1 switches that reproducibility guarantee back off, so that's the blocker to sort out first. I'd also lean toward splitting uv and Nix into separate PRs (you flagged this too); the uv half is close once #1/#3/#5/#6 are addressed, and Nix can land on its own once `flake.lock` is committed. Happy to help iterate on any of these. Thanks again for digging into the packaging story here.

Vaarij:
New changes here in response to discussions here. The new PR is at: #602

PR #602, opened 2026-07-01
[prototyping] Prototyping with UV, adds a lot of changes, would love opinions

Vaarij:
## What this changes

A short description of the change and why.

In response to comments on PR #601, I thought of separating uv and Nix, implementing uv first. This is a large change to the codebase, and it comes with some pros and cons. Using UV as a single source of truth synchronizes packages and defines intent in advance. `uv.lock` is generated and carried over instead of using `requirements*.txt`. In addition, using UV adds a layer of consistency in CI and with other packages like Docker. 

The biggest drawbacks are portability and the speed of contributions. Using uv requires another package manager that some contributors may not be familiar with. It also requires some setup before running `uv sync`. Another downside is dual use of `requirements*.txt` and uv. There is a likelihood of drift if `uv lock` and `make export-python` are not used together. As of now, I assume another contributor would decide to run both, but if they don't, it would create drift. 

Most of the design choices stem from trying to transition without explicitly switching to uv in a single PR. 

Closes PR: #601 

## Checklist

- [ X] Tests pass locally (`python3 -m pytest tests/ -q` and, for Rust changes,
      `cd simulations && cargo test --workspace`).

## Notes for reviewers

I welcome any opinions, comments, or concerns if I am missing something. I would like to note that I envision using a general package manager such as Nix. I would also consider using Bazel to accelerate builds and test both Python and Rust together. These will add more friction to contributions down the line; I welcome any discussion on that as well. 

Thank you to the reviewers.


ELares:
any reason you deleted my comment / review?

ELares:
If you delete it again. you will be banned from the repo, closing and starting the PR over makes it seem like you have malicious intent and will permanently ban you from the repo.

Vaarij:
> any reason you deleted my comment / review?

I didn't delete any comments or reviews, and I referenced them in this PR. I wanted to split prototyping into two separate PRs, which I referenced in the other PR when I closed it, so it wouldn't be mistaken as "in progress". The idea was to remove confusion about the branches, since they had different purposes. 

> If you delete it again. you will be banned from the repo, closing and starting the PR over makes it seem like you have malicious intent and will permanently ban you from the repo.

I apologize; I was not familiar with that, but I didn't have malicious intent here, since I referenced this PR in the closed PR to signal that it was a continuation, but that branch was stale.

Issue #462, opened 2026-06-24
Calibrate ACSL4-status prevalence from cBioPortal TCGA (anchor the #444 data-gated layer)


ELares:
> Surfaced by a 2026-H2 internet + literature innovation scan (9-angle web search, anchors adversarially verified against PubMed/DOI/dataset hosts). Proposal for discussion; keep the repo standards (off-by-default byte-identical sim layers, direction-over-magnitude with honest uncalibrated labeling, verified citations).

## What
The ACSL4-status biomarker layer (#444, v0.55.0) maps a per-tumor ACSL4 expression status to a PUFA-incorporation boost, but the per-cancer-type prevalence of ACSL4-negative/low (ferroptosis-refractory) tumors is flagged DATA-GATED. cBioPortal exposes TCGA PanCancer Atlas ACSL4/GPX4/SLC7A11 mRNA z-scores via a **login-free public REST API**.

## Tie to reality
Pull the per-cancer-type ACSL4 (and GPX4/SLC7A11) mRNA distributions, derive the fraction of ACSL4-low/negative tumors per cancer type, and commit the derived per-cancer-type prevalence as a reproducible calibration artifact (offline contract: fetch LOCAL, commit the summary, CI never downloads — mirrors scripts/fetch_calibration_data.py #330). Turns the ACSL4_NEGATIVE/LOW placeholders into real per-cancer-type priors and grounds the falsifiable stratification prediction.

## Anchor (verified)
cBioPortal public REST API (docs.cbioportal.org/web-api-and-clients), TCGA PanCancer Atlas mRNA z-scores. Layer: acsl4 module / Params.acsl4_status_boost.

## Standards
Derived summary committed; production defaults unchanged (byte-identical); honest about z-score-threshold choices.

PR #623, opened 2026-08-04
#616: no calibration target exists for the four unmodelled genes, and #462's prevalence prior is not gene-specific

### LINKED ISSUE #616: Calibration targets for the top unmodelled ferroptosis mechanisms (HMOX1, TP53, TFRC, KEAP1)
## Why

`analysis/atlas-model-gaps.md` asked the literature which ferroptosis mechanisms it
emphasises, against what `ferroptosis-core` has a parameter for. The top unmodelled
mechanisms, by ferroptosis-indexed article count:

| gene | articles | engine handle |
|---|---|---|
| HO-1 / HMOX1 | 1,583 | none |
| TP53 | 1,497 | none |
| transferrin receptor (TFRC) | 1,482 | none |
| KEAP1 | 768 | none |

All four are genuine ferroptosis regulators. The engine models NRF2
(`nrf2_gsh_rate`) but not its KEAP1 regulator; it models ferritinophagy
(`ferritinophagy_release`) but not iron *import* via TFRC.

## The gate this must pass

`CONTRIBUTING.md` "Simulation layer policy (calibrate-or-cut, #501)" requires a **named
calibration target and a CALIBRATION_STATUS row** before any new off-by-default axis
lands. This issue is therefore *not* "add four layers" — it is:

**For each of the four, determine whether public data exists that could calibrate it.**

Precedent: #444 added the ACSL4 layer flagged DATA-GATED, and #462 later partially
anchored it against cBioPortal TCGA. The honest outcome for a gene with no usable
dataset is to record that it is data-blocked rather than to add an uncalibrated axis.

## Acceptance

- [ ] for each gene, a search for quantitative data mapping onto a model observable
      (knockdown/overexpression vs ferroptosis sensitivity, dose-response, or a
      measurable rate constant)
- [ ] a `CALIBRATION_STATUS.md` row per gene: anchored, partially anchored, or
      data-blocked with the reason
- [ ] layers added **only** where a target exists
- [ ] `atlas_model_gaps.py` re-run afterwards so the coverage map reflects reality

## Caveat carried from the analysis

This ranks **attention, not importance**. A gene may be heavily studied because it is
easy to assay, and a gap is not a defect — several of these are upstream regulators
whose effect the engine may already absorb into an existing parameter.

ELares:
Closes #616.

## The question

`analysis/atlas-model-gaps.md` asked the cancer literature which ferroptosis mechanisms it emphasises and found four with no parameter in `ferroptosis-core`: **HMOX1** (1,583 ferroptosis-indexed articles), **TP53** (1,497), **TFRC** (1,482), **KEAP1** (768).

The layer-freeze policy (`CONTRIBUTING.md`) requires a named calibration target before any new off-by-default axis lands. So the question is whether public data could anchor them, not whether to write four layers.

## The answer, and why it reaches backwards

It cannot. `scripts/calibration_feasibility.py` extends the #462 cBioPortal pull to all seven genes across the 32 TCGA PanCancer Atlas studies:

| gene | below z = -1 | deviation from normal | spread (IQR) | below z = -2 |
|---|---|---|---|---|
| ACSL4 | 14.4% | -1.5 | 3.0 | 3.0% |
| GPX4 | 14.5% | -1.4 | 2.5 | 1.6% |
| SLC7A11 | 16.0% | +0.2 | 2.9 | 1.7% |
| HMOX1 *(gap)* | 14.9% | -1.0 | 1.8 | 1.9% |
| TP53 *(gap)* | 14.1% | -1.8 | 4.7 | 4.3% |
| TFRC *(gap)* | 15.2% | -0.7 | 2.0 | 2.1% |
| KEAP1 *(gap)* | 13.8% | -2.0 | 3.1 | 1.5% |

Within-cohort mRNA z-scores are standardised per study by construction, so the fraction below z = -1 is **15.87%** for any gene under approximate normality. Every gene lands within two points of it. That is the standardisation being recovered, not biology, so the route cannot separate a gene whose low-expression tail matters from one with no ferroptosis role.

## This corrects how #462's number may be read

`analysis/calibration/acsl4-prevalence-calibration.md` called the same figure "the real, committed population prior" for the refractory-leaning tail. It already carried the by-construction caveat, but it had **no control**, so the number still read as evidence about ACSL4. The six other genes are that control: a gene with no ferroptosis role returns the same figure.

The `status_from_zscore` bridge is unaffected, since it maps a scale and is separately justified. What changes is that the prevalence figure describes how the z-score partitions a cohort, not how many tumours are ACSL4-refractory, and that distinction matters because the refractory phenotype is exactly what the layer exists to stratify.

## What survives

TP53's **deep** tail: 4.3% below z = -2 against an expected 2.28% and the other genes' 1.5-2.1%, with much the widest inter-cancer spread. That is consistent with real TP53 deletion rather than standardisation, and is the one gene here where an expression-derived prior would carry information. Not enough to parameterise a layer; enough to record as worth returning to.

## What shipped

**No layer.** Four "Layers proposed and NOT built" rows in `CALIBRATION_STATUS.md` with the reason per gene, which is #616's stated acceptance criterion. Two of the four are also already absorbed by an existing parameter (TFRC by `ferritinophagy_release`, KEAP1 by `nrf2_gsh_rate`, which it regulates).

- `scripts/calibration_feasibility.py` + committed `analysis/calibration/calibration-feasibility.{md,json}`
- `scripts/fetch_acsl4_prevalence.py` extended to the four gap genes (Entrez IDs verified against NCBI: 3162 / 7157 / 7037 / 9817)
- the #462 doc, `atlas-model-gaps.md`, and `census-findings.md` all carry the correction, regenerated by their scripts rather than hand-edited
- 7 guards in `tests/test_calibration_feasibility.py`

## Verification

468 passed, 4 skipped. Every new guard was mutation-checked rather than assumed: perturbing KEAP1's median and TP53's deep tail in the committed JSON fails 3 tests, and stripping the correction paragraph out of the #462 doc fails a 4th. Restoring both returns to green.

## What would unblock this

A cell-line expression matrix joinable to the committed CTRPv2 curves (828 lines x 5 ferroptosis inducers) would make all four testable at once by correlating expression against EC50. The DepMap download catalogue served that until recently and now returns HTML rather than the documented CSV, so it is blocked on access, not on method.


ELares:
## Fresh-eyes adversarial review found a real error in this PR's analysis. Fixed in c833a8bf.

The review's highest-value attack — that the z-scores might be computed against a diploid/normal-adjacent reference rather than the tumor cohort, which would invalidate the whole inference — was tested and **the premise survived**: the profile selected is `..._all_sample_Zscores`, and independently, raw RSEM medians vary up to 218x across cancer types while `median_z` never leaves ±0.37 of zero, which is arithmetically impossible under a shared reference.

But it found something else, and it is real.

### The error

The PR compared TP53's deep tail against *"the others' ~1.5–2.1%"*. That range **silently omitted ACSL4's own 2.98%**, the highest of the six controls. With the comparator restored, the two thresholds disagree, and that split is the actual finding:

| cut | result |
|---|---|
| `z < -1` | **Nothing separates.** All seven genes within ~2 points of 15.87%. Not gene-specific. *(This half stands as written.)* |
| `z < -2` | **Two genes separate.** TP53 4.27%, above the 2.28% expectation in **31/32** cancer types (sign test p=1.5e−8). ACSL4 2.98%, above it in **25/32** (p=0.002). The other five sit at or below expectation. |

A left tail heavier than a normal is the signature of a discrete low-expression subpopulation, and standardisation cannot manufacture one. So **#462's `z < -2` row survives the very control that washes out its `z < -1` row** — and it is the row the layer actually needs, since `ACSL4_NEGATIVE` sits at z = −2. The correction is now scoped to the shallow cut.

The verdict on the four unmodelled genes is unchanged and better supported, since both cuts were checked instead of one.

### The other blocker: the correction wasn't applied where it mattered

The retracted framing survived in five places, including **`ferroptosis-core/src/acsl4.rs`'s public doc comment on `status_from_zscore`** — a reader landing on the shipped library API got the pre-#616 story with no signal it had been retracted. Also the section heading directly above the correction block, two later restatements, the `CALIBRATION_STATUS` status field, and a test comment. All fixed.

### Tests: 4 of 6 mutations had survived

The commit claimed the guards were mutation-checked. They were, but too narrowly — nothing constrained `z < -2` for any gene except TP53's rank, and the prose guard pinned one literal string, so reinstating the old framing in different words passed. Rewritten to recompute every statistic for all seven genes at both cuts and to pin the separated set rather than a rank. **Nine mutations now run — the four that survived, plus two more found while fixing them — and all nine fail.**

### Also fixed

- *"The top four"* sat directly beneath a table where **KEAP1 ranks ninth**. The selection is #616's, by plausible mechanism rather than rank; it now says so.
- *"TFRC is already absorbed into `ferritinophagy_release`"* was **wrong**, and #616 says so explicitly: the engine models NCOA4 release from intracellular ferritin stores, a different source from transferrin-receptor import across the membrane. TFRC is a genuinely uncovered axis.
- *"No public dataset maps onto a model observable"* overstated a two-route search. Narrowed, with the un-run search named.
- The **"spread (IQR)" column was not an IQR** — it indexed the 25th minus the 9th order statistic, overstating TP53's spread by 19%. Now interpolated quartiles.
- The **"mild right-skew" explanation is wrong** for ACSL4 and TP53, which sit below expectation at z=−1 and above at z=−2. That combination is the subpopulation signature, and it is precisely why the two cuts had to be measured separately.
- `census-findings.md` said the route *"cannot anchor anything"* while the feasibility doc granted TP53 a weak anchor. Two generated docs, opposite conclusions.

472 Python passed, 4 skipped; 369 ferroptosis-core passed.

PR #624, opened 2026-08-04
Census figure (fig28) + thesis-leg trajectory: nothing is gaining ground

ELares:
Two census-layer additions. Retargeted to `main` now that #623 has merged.

---

## 1. fig28 — the sampling-bias finding finally has a figure

The census established that the frozen corpus is a **213-fold non-uniform sample** of the cancer literature. That bounds the widest set of manuscript claims, since every relative prevalence computed over the 4,830-article corpus inherits it. It had only a sentence.

**Panel A** — per-mechanism capture on a log axis, because a linear one collapses everything below 10% into the axis and the range is the point: epigenetic 0.20% to mrna-vaccine 41.86%.

**Panel B** — pharmacological to physical is **9.1:1** by the manuscript's method and **17.6:1** on the census, because the corpus over-samples physical modalities 3.3x. The claim survives, in the direction that costs this project something: the manuscript understates its own case.

### Caught in flight

Computing "pharmacological" as *everything not physical* reports 12.5:1 → 25.0:1, contradicting the committed analysis, because it sweeps in delivery platforms and genetic tools (nanoparticle, crispr, oncolytic-virus, mrna-vaccine) that are neither a drug class nor a physical modality. Rather than copy the curated list into the figure, `PHARMACOLOGICAL` is hoisted to module level in `atlas_landscape.py` and imported — a hand-written list beside the real one is exactly how the discrepancy arose. The hoist is behaviour-preserving and the regenerated ratios match the committed doc exactly.

A mechanism with no census articles is **excluded rather than plotted at 0%**; 0/0 drawn as zero would read as "never captured", a different claim.

---

## 2. Thesis legs: trajectory, not just size

`atlas_thesis_position.py` counted each leg and reported the field's growth, but never asked whether any leg's position *within* that field changed. A snapshot cannot separate "small because it is new and rising" from "small because nobody went there" — and those imply opposite things about a 32-article leg.

Measured as each leg's **share** of the ferroptosis field, pooled into three-year windows because the single-year cells are too small to compare (the 2019 SDT and PDT cells hold 0 and 4 articles). Wilson 95% intervals; a leg counts as moved only when the intervals are **disjoint**, stricter than a two-proportion test.

| leg | 2019-2021 | 2023-2025 | verdict |
|---|---|---|---|
| lipid peroxidation | 20.85% | 11.43% | **falls** |
| drug resistance | 3.59% | 3.84% | flat |
| photodynamic therapy | 1.80% | 1.32% | flat |
| immunotherapy | 0.65% | 0.51% | flat |
| sonodynamic therapy | 0.14% | 0.24% | flat |

**Nothing moved.** The composition of the ferroptosis literature has been stable through a 25-fold expansion.

The one exception is vocabulary, not science: NLM introduced the `Ferroptosis` descriptor on 2020-01-01, so the early window sits on the transition when papers were still co-indexed with the general `Lipid Peroxidation` term. The row is kept and explained rather than dropped.

### What it changes

It removes the most favourable reading the earlier snapshot allowed. A 32-article leg could have been an under-explored direction just starting to attract people. It was not — the sonodynamic share has sat at roughly a quarter of one percent of its own field throughout the boom. **"Persistently unexplored" rather than "newly emerging."** Not a refutation, since the share is not falling either, but the manuscript cannot argue the field is moving toward its thesis.

The same correction cuts in the project's favour on the resistance leg: it is stable rather than growing, so it is a dependable place to stand and should not be called a rapidly growing area on the strength of the field's overall growth.

One claim here was asserted from memory and was wrong: the Ferroptosis descriptor entered MeSH in **2020**, not 2018. Checked against NLM's record for D000079403 and corroborated by the census, which holds no ferroptosis-indexed article before 2019.

Adds `--render-only` so the report rebuilds from committed JSON without rescanning 1,334 census shards; the re-render reproduces every committed total exactly (13,346 / 32 / 177 / 479).

---

## Verification

482 passed, 4 skipped. Ten new guards. The figure's five recompute its numbers from committed JSON and assert they still match `atlas-landscape.md`, so figure and prose cannot drift; mutation-checked by inflating sonodynamic's keyword count and epigenetic's frozen count.

The rebase onto squash-merged #623 conflicted on the four files every branch touches, and resolving with `--ours` silently dropped fig28's CLAUDE.md entry — the shared-append-file hazard. Caught by grepping for each addition after the rebase rather than trusting a clean result; restored in the final commit.


ELares:
## Fresh-eyes adversarial review found two blockers. Both real, both fixed in 2c859325.

### BLOCKER 1 — evidence of absence from a test with no power

The disjoint-Wilson test is conservative, and conservative means **blind over a range**. For the sonodynamic leg, with 2 early articles, it could not have called a rise below **4.65x**. The report turned "flat" into *"the field is measurably not moving toward the thesis"* — a claim that test cannot support.

Every verdict now ships with the change it could actually have detected:

| leg | flat verdict, but would have needed |
|---|---|
| sonodynamic | a **4.7x** change |
| immunotherapy | a 2.2x change |
| photodynamic | a 1.6x change |
| drug resistance | a 1.4x change |

The sonodynamic conclusion is now **"no detectable acceleration"**, not "no acceleration". The verdicts themselves are unchanged and independently sound — Fisher exact agrees with every one, so the conservatism did not manufacture the result. The defect was the strength of the conclusion drawn from it.

### BLOCKER 2 — the narrative was hardcoded while the verdict was computed

The mover list was assigned to a variable and **never read**. A different window choice produced a document asserting "nothing moved except lipid peroxidation" three lines under a table showing two movers. The prose is now conditional on the computed verdicts.

Worse, the test that supposedly guarded this pinned the exact literal the generator always writes, so it **could only ever pass**.

### Also fixed

- **The figure showed two of three arms.** It does not reconcile — 9.1 × 3.3 = 30, not 17.6. The missing factor is the method effect, and it runs the *other* way: holding the corpus fixed and switching keyword labels for MeSH **cuts** the ratio to 5.3:1, so by the more independent labelling the manuscript **overstates**. All three arms are now plotted with the decomposition stated, and panel B is retitled "Method and corpus pull in opposite directions".
- **Panel A was bars on a log axis.** A log axis has no zero, so bar *length* encoded matplotlib's chosen left limit, flattening a true 213x to ~21x visually. Now a dot plot, where position is the encoding.
- **The lipid-peroxidation "vocabulary artifact" was asserted as fact.** The series keeps falling three years past the descriptor's introduction and then *reverses*: 30.1, 22.3, 18.2, 10.1, 8.3, 11.5, 12.8, 13.0. Indexers dropping a general term cannot produce a rise. Now stated as an open question.
- **The window split was undisclosed and post-hoc.** A sweep of every contiguous 2- and 3-year pair now runs per leg. The sonodynamic verdict holds in all of them; **drug resistance moves in 2 of 6**, and the report says so rather than reporting only the split chosen first.
- **A bogus corroboration.** "The census contains no ferroptosis article before 2019" proves nothing about a 2020 descriptor date — 2019 precedes 2020, and PubMed retro-indexes. Removed.
- **Two growth multiples in one file** (11x computed, 25x hardcoded). The hardcoded one is gone.
- **`--render-only` could fabricate a row.** Adding a leg and re-rendering silently emitted `0.00%`, indistinguishable from a measurement. It now records `LEGS`/`CORE` provenance and refuses on a mismatch.
- **`atlas_landscape`'s maturity table labelled a column "pharmacological" while computing the complement.** The comparator is deliberate for a maturity question; the label was not. Regenerated — JSON byte-identical, only the label and its explanation changed.

### Tests

The figure guards **never ran the figure** — they recomputed in a parallel re-implementation, so they passed while panel B plotted the opposite finding, while the log axis was removed, and while the PNG was replaced entirely. They now call the figure code into a temp directory and assert on what it reports plotting.

Five mutations verified after the fix: wrong arm plotted, no log axis, dropped method arm, removed detection floors, falsified mover line. **All five fail.**

487 passed, 4 skipped.

### Checked and clean

The review also independently confirmed the parts that hold: all Part 1 numbers recompute exactly from the committed JSON; the `PHARMACOLOGICAL` hoist is behaviour-preserving so `atlas-landscape.md` was never stale; excluding `frequency-therapy` hides nothing; the MeSH descriptor ID and 2020-01-01 date are both correct; `--render-only` is a clean fixed point; and the manifest is right for every changed file.

Issue #617, opened 2026-08-04
Manual precision check on the 29.1% of co-mention disagreements that are abstract-visible


ELares:
## Why

`analysis/atlas-comention-audit.md` gave the full-text co-mention layer its first
precision measurement, and it is a **bound, not a number**: 44.3% to 83.8%.

The bound cannot be narrowed from that data. PubTator reads abstracts; the co-mention
layer reads full text, so an entity discussed only in Methods is genuinely present and
genuinely absent from PubTator's annotation.

Splitting the disagreements localises where the remaining uncertainty lives:

| | share of disagreements |
|---|---|
| body-only — PubTator could not have seen it | 70.9% |
| **alias visible in the abstract PubTator read** | **29.1%** |

The body-only share is the layer doing its job. **The 29.1% is where false positives
would be**, because PubTator read the same text and did not assign that entity.

## Why it matters

`atlas-module-support.md` uses co-mention counts to argue that a zero in the relation
column is an abstract-level extraction failure rather than absence of evidence. That
argument is only as good as this layer's precision.

## What to do

`corpus/atlas/comention/audit-sample.jsonl.gz` holds 400 uniformly sampled sentences
with resolved entity names. Filter to the abstract-visible disagreements and read them.

## Acceptance

- [ ] the abstract-visible disagreement subset extracted from the audit sample
- [ ] each judged by hand: is the matched alias genuinely referring to that entity in
      that sentence?
- [ ] a precision point estimate reported, replacing the 44.3%-83.8% bound
- [ ] any systematic false-positive pattern fed back into `usable_alias` or the
      sense filter in `scripts/atlas_comention.py`

## Note

The mechanical check is already clean: 1,112 of 1,112 sampled entity mentions have
their alias present in the sentence, so there is no tokenizer bug to chase.

PR #627, opened 2026-08-04
#617 did not work: co-mention precision fell 7 points, and retuning will not fix it

ELares:
#617 measured this layer at 55.5% precision, traced it to one gap, fixed it, and **asked for a fresh sample after the next rebuild to confirm the repair**. The rebuild has run. The change made things worse.

## Population-weighted precision fell ~7 points, 56.8% → 49.9%

| per 400 sampled sentences | before | after |
|---|---|---|
| entity mentions | 1,112 | 1,484 (1.33x) |
| PubTator corroborates | 44.3% | 32.5% |
| **abstract-visible disagreement** | 8.2% | **23.7%** |
| body-only disagreement | 47.5% | 43.9% |

## The mechanism is not the obvious one

The abstract-visible stratum — the only one that *can* hold false positives, since PubTator read that text — actually got **cleaner**: 14.6% → **26.7%** precise on 60 hand-judged mentions (95% CI 17.1–39.0%).

But it nearly **tripled** as a share of all mentions. Tripling a stratum at 27% precision costs more than cleaning it gains.

## Retuning the thresholds will not recover it

Over the judged sample, the two filters that #617 introduced as replacements **do not separate true matches from false ones at all**:

| | census support | share of identifier |
|---|---|---|
| false positives | 77 – 355,572 | 11.5% – 274% |
| true positives | 297 – 100,758 | 20.2% – 68.9% |

The distributions overlap almost entirely. No threshold recovers the loss without taking the true matches with it. **Design failure, not a tuning problem.**

## A concrete bug, biased the wrong way

`alias_support` sums a form's mentions across **every sense it carries**. `ident_mentions` sums an identifier's across **every form**. Their ratio is not a share and can exceed 1 — **274% for `as`**, 133% for `gp`, 104% for `tss`.

The direction of the bias is the damaging part: an ambiguous generic word collects a *larger* numerator from its other senses, so a **minimum**-share filter admits it more readily than a specific name. The filter written to exclude `as` and `treatment` was structurally disposed to keep them.

Fixed at the source — `atlas_graph.build_index` now records `alias_ident_support`, the count of a form for the identifier it actually resolves to. The index rebuild is running; the co-mention rebuild that confirms the repair is the follow-up.

## The transferable lesson

The #617 justification — "kills the same 141 of 152 false positives and retains strictly more true ones" — was **true about the sample it was measured on and wrong about the population it was applied to**. The false positives were multi-word *because the multi-word channel was the unfiltered one*. Closing it moved the pressure to the channel that had just been opened, and the top offenders are now bare English words: `treatment` (16), `effects` (15), `as` (13), `advanced` (10), `function` (7), `other` (6).

**A filter justified by an error distribution it then changes has to be re-measured.**

## What this does not change

- The multi-word repair was real. `tumor cells`, `overall survival` and `et al` appear nowhere in the fresh sample's disagreements.
- The layer's purpose is unaffected — body-only matches are still entities an abstract-level extractor structurally cannot reach.
- **No manuscript number depends on this layer.** Its consumer `atlas-module-support.md` should now be read against 50% precision rather than 55%.

## On the evidence

The 60 judgements are **committed as a CSV**, not asserted, so the number is auditable. The write-up states plainly that they are mine and unblinded — which is exactly the bias that would have flattered a fix I wrote, and the result runs against it.

496 passed, 4 skipped. Six guards pin the finding to recomputed quantities, including one that fails if the share bug returns.


ELares:
## Correction: fixing the share bug does **not** recover the precision

Two things checked after opening this, both of which change what it claims.

**1. The source fix is insufficient.** The PR said the bug was "fixed at the source" and a rebuild would confirm. I rebuilt the index with `alias_ident_support` and re-scored the judged forms against the corrected share:

- `as` still reads **124.5%** — down from 274%, but still above 1. The co-mention layer can redirect a blocklisted form to a curated sense whose identifier is not the one the majority vote counted, so numerator and denominator are still not guaranteed commensurable.
- More decisively: at the shipped 5% threshold the corrected share cuts **nothing** — 0 of 37 judged false positives and 0 of 13 true positives. **The filter is inert on this stratum whether it is computed correctly or not.**

So the bug is real and worth fixing on its own terms, but it does not rescue the filter. That is the same conclusion the support/share overlap reaches, arrived at independently — which strengthens the "design failure, not a tuning problem" reading rather than weakening it.

Claiming a repair the measurement does not support is exactly the failure mode this document is about, so it would have been a poor place to commit one.

**2. The comparison baseline was read from `git HEAD`**, which broke the moment the post-rebuild audit was committed — the script then compared the run against itself and silently refused to produce output. The historical run cannot change, so it is now pinned to the commit it came from and verified against git on every run, failing loudly if the pinned numbers stop matching.

497 passed, 4 skipped. Two further guards added: one that fails if the baseline goes back to being read from HEAD, one that fails if the document starts claiming the repair.

Issue #628, opened 2026-08-04
Co-mention alias filtering needs a redesign, not a threshold


ELares:
Follow-up to the measurement in `analysis/comention-regression.md` (PR #627). Recording the open problem rather than leaving it implicit in a merged analysis.

## State

The full-text co-mention layer sits at ~50% population-weighted precision, down from ~56% before the #617 filter change. The two filters introduced there — census support and share of an identifier's mentions — were measured and **do not discriminate**:

| | census support | share of identifier |
|---|---|---|
| false positives | 77 – 355,572 | 11.5% – 274% |
| true positives | 297 – 100,758 | 20.2% – 68.9% |

At the shipped 5% share threshold, the corrected share cuts 0 of 37 judged false positives and 0 of 13 true positives. It is inert.

## Why a threshold cannot fix it

The distributions overlap almost entirely, so any threshold that removes false positives removes true ones at a similar rate. The measured failure mode is bare English words matching in their ordinary sense — `treatment` (16 of the sampled disagreements), `effects` (15), `as` (13), `advanced` (10), `function` (7), `other` (6) — while genuine biomedical concepts that are *also* common (`apoptosis`, `proliferation`, `resistance`, `migration`) must survive. Frequency alone does not separate those two groups, which is why the support filter fails.

## What is known about the shape of a fix

- The deleted single-token shape proxy (require a digit or hyphen) did work on this channel, but it also removed every purely alphabetic drug name — `erastin`, `cisplatin`, `pembrolizumab` — so restoring it as-is is not acceptable.
- The discriminator needed is *specificity of the form as a name*, not frequency. Candidate signals not yet measured: document frequency across the full-text corpus (a form in most documents cannot discriminate), whether the form is the identifier's preferred term rather than any alias, and the number of distinct identifiers a form maps to.
- A residual incommensurability remains in the share itself: `alias_ident_support` uses the majority-vote identifier's count, while the layer may assign a redirected curated sense, so `as` still scores 124.5%. A correct share needs the count for the identifier actually assigned.

## Acceptance criteria

1. A candidate discriminator measured against the committed 60 judged mentions (`analysis/comention/abstract-visible-judgements.csv`) **before** any rebuild, showing separation the current filters lack.
2. A held-out judged sample after the rebuild, since the existing 60 will have been used for selection.
3. Population-weighted precision reported against the 49.9% baseline, with the body-only and corroborated strata re-measured rather than carried over — they are currently assumptions.
[...]

PR #629, opened 2026-08-04
Correct the judging criterion (47.5%, not 49.9%), and find a discriminator that works (#628)

ELares:
Two things. The first corrects my own measurement from #627.

## 1. The judging criterion was wrong, and it flattered the layer

The first pass asked: *does the sentence contain the string that matched?* That is the wrong question. The **identifier** is what the co-mention pair is recorded against, so what matters is whether the sentence discusses the entity that identifier *denotes*.

Checked against NLM's own descriptor labels:

| matched form | what the identifier actually denotes |
|---|---|
| `apoptosis` | Malformations of Cortical Development, Group I |
| `node` | Sick Sinus Syndrome |
| `chemotherapy` | Chemotherapy-Related Cognitive Impairment |
| `resistance` | Disease Resistance (host resistance, not drug) |
| `Migration` | Tooth Migration |
| `Aging` | Aging, Premature |
| `artery` | Renal Artery Obstruction |

**Every one of those was scored correct on the first pass. None is.**

Abstract-visible precision is **10/60 = 16.7%** (95% CI 9.3–28.0%), not 26.7%. The weighted total falls to **47.5%**, so the drop from the pre-#617 baseline is **9.3 points**, not 7.

Three consequences are now stated in the document rather than left implicit:
- The comparison against the earlier 14.6% is **not known to be methodologically comparable** — that figure may have used the same surface-form criterion.
- The same caveat attaches to the carried-over 92.5% and 30.8% strata. If they share the flaw, they are upper bounds and the total is optimistic **on both sides** of the comparison.
- Both verdict columns are committed side by side, with each identifier's NLM label attached, so the correction can be audited rather than taken on trust.

## 2. A discriminator that works — #628's first acceptance criterion

The support and share filters fail because they measure how *often* a form appears, and the offenders are common words. The signal that separates them is whether the form is a **name of the entity at all**.

The index cannot answer that: its `canon` field is the most frequent surface form, derived from the same annotations, so testing against it is circular. NLM's labels are external.

Requiring the matched form to be the entity's authority name, compared as a word bag so `squamous cell carcinoma` matches `Carcinoma, Squamous Cell`:

| | true | false |
|---|---|---|
| kept | 4 | 2 |
| cut | 6 | 48 |

**Removes 96% of false positives** against 60% of true ones, lifting precision on what it keeps from 16.7% to 66.7%. The filters it would replace cut **nothing at all**, so this is a discriminator rather than a threshold.

### It is not a recommendation yet

Stated in the document and guarded by a test: it was **selected on this sample**, so its numbers are optimistic by construction; only 6 mentions survive it, so the kept-precision CI is very wide (30–90%); and it has been tested only on the abstract-visible stratum. #628 already records what a real evaluation needs — a held-out judged sample, and a MeSH/NCBI label dump so the check does not depend on a live API.

## How this was found

An experiment looking for something else. The authority lookup returned "Malformations of Cortical Development" for `apoptosis`, which looked like a broken lookup — I checked it against a known descriptor instead of assuming, and the lookup was right. The layer really had assigned that identifier.

## Verification

499 passed, 4 skipped. Three new guards: one that fails if the reported precision stops using the identifier-level verdicts, one that spot-checks the case that exposed the flaw, and one that fails if the discriminator stops beating the filters it replaces **or** if the document starts presenting it as validated.


ELares:
## Addressed the weakest point in this PR before it was raised

The comparison's soft spot is that the carried-over precisions (92.5% corroborated, 30.8% body-only) were measured under #617 and may share the surface-form criterion this PR abandons. If they do, the `before` total is overstated too — and the whole regression could be an artifact of correcting only one side.

Worked out rather than hedged. Scale both sides' impure strata by the same factor `k`, leaving the corroborated stratum alone since PubTator's agreement is already an identifier-level check:

| k | before | after | change |
|---|---|---|---|
| 1.0 | 56.8% | 47.5% | −9.3 |
| 0.8 | 53.7% | 44.0% | −9.7 |
| 0.6 | 50.5% | 40.5% | −10.0 |
| 0.4 | 47.3% | 37.0% | −10.3 |
| 0.2 | 44.2% | 33.5% | −10.6 |
| 0.0 | 41.0% | 30.0% | −11.0 |

**The change is negative for every `k`, and gets more negative as `k` falls.** Correcting both sides *widens* the gap rather than closing it, because the `after` run carries more of its weight in the impure strata.

So the regression does not depend on the carried-over numbers being right — only on their being wrong by a similar factor on both sides, which is precisely the assumption the objection needs.

Guarded by a test that sweeps `k` across [0,1] and fails if the sign ever flips or if the gap stops widening under correction.

500 passed, 4 skipped.

ELares:
## Review found a blocker in how this PR described its own result. Fixed.

It is the same class of error the document is about, which is worth stating plainly.

### BLOCKER — the discriminator is an identifier-level check, not a per-mention one

The section claimed to require *"the matched form"* to be the entity's authority name. **The audit sample never records the matched span.** `atlas_comention` writes `entity_names` as `canon[identifier]` — the identifier's most frequent surface form *across the whole corpus*.

Verified: the column is constant for every identifier appearing more than once (0 of 10 have two values), and it is **absent from the sentence in 20 of the 60 rows**. `left ventricular dysfunction` sits against *"She was diagnosed with left breast cancer"* — the span that actually fired was presumably `left`.

So what was measured is: *does this identifier's dominant corpus name coincide with its authority label?* It cannot distinguish two mentions of the same identifier, and it would be applied when **building the alias map**, not when matching text. Cheaper than the filter described, and blunter.

The 2×2 stands as a real association. The mechanism named for it did not. The section was also self-refuting — it argued the index "cannot answer this because its `canon` field is circular", then computed the answer from that field.

### The filter removes every gene

NCBI Gene identifiers have no MeSH descriptor, so the rule cuts them unconditionally. This sample is 58 of 60 MeSH, so the 96% figure **says nothing about the gene half** of a layer built over gene, chemical and disease annotations — in a repository whose subject is GPX4 and ACSL4. A label source covering NCBI Gene is a precondition, not an optimisation.

### The word bag is not better than plain equality here

Equality fires on 4 rows, the word bag on 6. The two extra are one true match (`squamous cell carcinoma`) and one **false** one (`left ventricular dysfunction` vs `Ventricular Dysfunction, Left`). It adds a TP and an FP and **lowers** kept precision from 75% to 66.7%. Only the favourable example was quoted; both are now.

### The headline mixed two criteria

It paired a lenient before-abstract with a strict after. All three are now tabulated:

| comparison | change |
|---|---|
| both lenient | −7.0 |
| both strict | −8.9 |
| **as reported (mixed)** | **−9.7** |

The reported figure is the most pessimistic. The direction holds in all three; the magnitude is a 7–10 point band, not a single number.

### Tests

Three real gaps, all demonstrated by the review with passing mutations:
- Nothing derived `is_authority_name` — marking the flagship counterexample (`apoptosis` vs *Malformations of Cortical Development*) as a name match passed all ten tests. Now re-derived in the guard.
- No recall guard — a filter keeping exactly one mention scored 100% precision and passed. Bounded on both sides now.
- A live test still asserted the stratum "got cleaner", which the corrected prose denies, sitting 2 verdicts from failing.

**All four mutations fail now.**

### One verdict flipped

`viral infection` → *Virus Diseases*, for a sentence about viral mRNA synthesis, is virology rather than a viral disease. Letting it stand was inconsistent with the strictness that made `radiation` and `chemotherapy` false positives. Precision is **15.0%**, weighted total **47.1%**.

### Independently confirmed by the review

All 60 verdicts re-judged (one flip, everything else inside the committed CI); all 50 authority labels re-verified against NLM (**48/48 resolvable match byte-for-byte**, both empties correct); the arithmetic reproduces exactly; the document regenerates byte-identically; and the regression's direction survives correcting both sides.

502 passed, 4 skipped.

PR #633, opened 2026-08-04
#628: measure all three strata, and record what actually matched

ELares:
Two assumptions were doing the work in the co-mention precision figure. One of them was wrong. Both are now measured.

## What the strata actually measure

40 hand-judged mentions each, at the identifier level, with the matched span and the NLM label in view:

| stratum | measured | 95% CI | assumed | verdict |
|---|---|---|---|---|
| corroborated | **90.0%** (36/40) | 76.9–96.0 | 92.5% | holds |
| body-only | **20.0%** (8/40) | 10.5–34.8 | 30.8% | **optimistic by a third** — and it carries the most volume |

Weighted precision with nothing carried over: **50.6% → 41.6%**. The regression is **9.0 points**, consistent with every earlier estimate, but both endpoints move down because the assumptions flattered them.

## The failure modes have different shapes

**Body-only** is the same generic-form problem seen everywhere else, with no abstract-level extractor available to disagree: `evaluation` → *Phobia, Social*; `world` → *Cutaneous Leishmaniasis*; and `mPTP` — the mitochondrial permeability transition pore — → the neurotoxin *MPTP*, a textbook sense collision.

**Corroborated** failures are milder and systematic: a generic span landing on an over-specific descriptor. `hypoxic` → *Hypoxia, **Brain***; `metabolic` → *Metabolic **Diseases***.

## This retracts a framing the repo has repeated

`atlas-comention-audit.md` said the body-only share was *"the layer doing the job it exists for"*. **It is the worst stratum of the three.** Being unreachable by PubTator is what the layer exists to find; it is not evidence that what it found is right. The report now says so where it used to say the opposite.

## Recording what actually matched

The audit sample stored `canon[identifier]` — the identifier's most frequent corpus-wide surface form — and **never the span that fired**. That is why the first hand-judging pass went wrong, and it made every verdict unverifiable.

The build now records `matched_spans`. Spans for the existing sample were recovered by re-running the matcher over the stored sentences (58 of 60; the miss is explained by the alias having since been filtered). They are worse than the canonical names suggested:

| stored as | actually matched |
|---|---|
| `left ventricular dysfunction` | `left` |
| `peritonitis` | `primary` |
| `hypersensitivity reactions` | `reaction` |
| `AS` | `activity` |
| `GP` | `glycoprotein` |

This also resolved a verdict flagged as unverifiable in the last review: `lactate` was judged correct but the word sat past the 300-character truncation in the committed CSV. It is there in the full sentence.

## Verification

513 passed, 4 skipped. Both judged samples are committed with spans and labels so every verdict is checkable. Three mutations verified: marking body-only all-correct fails, shrinking a sample below a usable size fails, dropping the matched-span column fails.

Closes acceptance criterion 3 of #628. Criterion 2 (held-out sample) remains open.


ELares:
## Smoke-tested the new build path (it would otherwise fail after hours)

`matched_spans` is computed in the audit-writing loop at the *end* of a multi-hour build, so a bug there costs a full rebuild. Replayed that exact loop against the real 400-row sample:

```
400 rows, 1484 entity mentions, 0.9s (0.6 ms per mention)
spans recovered for 1471/1484 mentions (99.1%)
alias map: 44,287 forms
serialised OK -> 170 KB
```

`alias` is in scope (assigned at `atlas_comention.py:341`, inside `main()`), the cost is under a second, and the output serialises.

Worth recording how I nearly got this wrong: an AST scan I wrote to check scope reported `alias` as *not assigned in main()* — because `alias, amb = build_alias_map(idx)` is a tuple assignment and my scan only looked at `Name` targets. I was one step from "fixing" a bug that did not exist. Running the code settled it; reading a script about the code did not.

ELares:
## Review found two blockers, both places where a shipped artifact contradicted its own committed data. Fixed.

### BLOCKER 1 — "the worst of the three" is false

The audit said body-only was *"the worst of the three"* at 20.0%. **Abstract-visible is 15.0%**, so body-only is the *second* worst — and the two are not separable at these sample sizes (Fisher p ≈ 0.59, intervals overlapping across most of their range).

Worse: both the figure and the superlative were **hardcoded strings** that would not track the next rebuild. The ordering is now read from the measurement and printed with intervals:

| stratum | precision | 95% CI | n |
|---|---|---|---|
| abstract-visible | 15.0% | [8.1%, 26.1%] | 60 |
| body-only | 20.0% | [10.5%, 34.8%] | 40 |
| corroborated | 90.0% | [76.9%, 96.0%] | 40 |

The retraction itself was right. The superlative was not.

### BLOCKER 2 — the report claimed all three strata were measured *and* that two were not

Both live in the shipped file: *"All three rows are now hand-judged"* in the body, and *"Body-only and corroborated precision are carried over unmeasured"* in its own Limits — a verbatim denial of the headline. A source comment also asserted an assumption was "stated in Limits" when it appeared nowhere a reader sees. Both fixed.

### Also fixed

- **#617's 30.8% was a measurement, not an assumption** (39 body-only mentions of the *before* run), and substituting the after-run 20.0% into that side was undisclosed. Both bases now tabulated: **−9.0** as reported, **−14.2** on each run's own measurement. Not separable (Fisher p ≈ 0.31), so it's a choice between defensible options — and the reported row is the conservative one. The delta is robust to the choice; the *level* is not.
- **`census-findings` attributed 51% to #617**, which measured 55.5%. Two files in one change disagreeing about a cited prior measurement.
- **`matched_forms` did not reproduce the build's matcher.** It generated n-gram candidates; the build does longest-match *with consumption*. In "breast cancer", `cancer` and `breast` are both aliases but only the longer one fires. That would have put non-firing spans in front of the next judging round — the exact failure the field was added to prevent. Rewritten and guarded; two mutations (unigrams-only, substring search) previously passed the whole suite. Re-deriving the committed spans changed **3 of 140**, all correctly, and no verdict moves.
- **"Every verdict is checkable" was false and its guard vacuous.** `csv.DictReader` supplies every header key, so the test checked the *header* — blanking every span passed. Three rows carried a span past the CSV truncation, including the `lactate` row this PR claimed to have resolved. Full sentences now committed for those rows; the guard checks values.
- **The intervals were computed, stored, and never propagated.** After total spans **31.5–52.6%**, before **40.3–60.3%**. `atlas-module-support` now quotes the interval beside the point estimate it was repeating five times.
- **An undisclosed confound**: 11 of 36 corroborated true positives sit in reference lists, table rows or PMC markup rather than the paper's own prose, against **0 of 8** body-only. A co-mention harvested from a cited title is recorded against the *citing* paper. Asymmetric, in the highest-precision stratum with a third of the volume, now in Limits and computed rather than asserted.
- Stale module docstring (56.8% → 47.1%). And the PR body's "58 of 60" should read **59 of 60** — exactly one span is unrecovered, consistent with "the miss", singular.

### Independently confirmed by the review

All 80 verdicts re-judged with **zero cross-file inconsistencies** (every `(span, label)` pair judged in two files agrees); one contestable verdict that stays inside its CI; sampling verified sound by pmid-cluster bootstrap giving intervals essentially identical to the committed Wilson ones; the build path confirmed non-crashing at ~9s; and `comention-regression` regenerating byte-for-byte.

516 passed, 4 skipped.

PR #677, opened 2026-08-06
close the staleness audit: the last generator was clean

ELares:
`atlas_entity_audit.py` was the one graph-derived generator #675 left un-run — it queries NCBI, and I'd rather flag a network dependency than run it carelessly at the end of a long change.

Run now: it regenerates **byte-identical**. The FSP1 finding is intact after the corrections grew — still 1 mismatch of 53 symbols, still `FSP1` → gene 51062 = ATL1 against the real AIFM2.

## The audit's tally is worth more than its result

A "last commit predates the input" heuristic flagged **7 artifacts**. Two had a real problem:

- `atlas-emergence-error` — genuinely stale
- `atlas-contradictions` — a total that had drifted ~50% while living only in stdout

`atlas-model-gaps` and `atlas-entity-audit` were fine; their content simply doesn't depend on what changed.

So the heuristic ran at **under a third precision**, and every one of the four verdicts required actually regenerating and diffing.

**A timestamp is a prompt to check, not a finding.** Recorded in CLAUDE.md so the next person auditing this doesn't read 7 flags as 7 problems.

608 passed, 4 skipped.


ELares:

PR #678, opened 2026-08-06
no seed-specific signal adds anything over a candidate-only prior

ELares:
The discovery leaderboard compares rankers **in isolation**, which can't answer the question anyone actually has. Losing alone isn't the same as adding nothing — a signal can be weaker than a baseline and still be complementary to it.

The structural fact that makes this the right question: `popularity` is the **only** ranking whose score is a function of the candidate alone (`deg_c` doesn't vary with the seed). Every other method varies through the bridge set. So the published table already reads as a *seed-blind* score beating every seed-aware one — and what it leaves open is whether the seed-aware ones carry anything the seed-blind one doesn't.

## Method

Blend each method's within-seed percentile into a degree-only benchmark — which reproduces the published popularity precision **exactly** (15.5%, and is asserted to) — and test the per-seed change in hits through the same paired bootstrap the evaluation uses. Percentiles rather than raw scores, so seed-level scale differences can't leak into the blend. **Decision rule fixed in the docstring before the result**, Holm-corrected across five methods.

## Result: none adds measurable precision

| method | mean Δ hits/20 | 95% CI | headroom? |
|---|---|---|---|
| adamic_adar | −0.095 | [−0.215, +0.025] | no |
| resource_alloc | −0.095 | [−0.220, +0.025] | no |
| bridges | −0.125 | [−0.265, +0.005] | no |
| abc | **−0.485** | [−0.725, −0.255] | no (hurts) |
| jaccard | **−1.795** | [−2.180, −1.425] | no (hurts) |

On this metric the candidate prior **exhausts what is measurable** — which makes "the shipped ranker is bad" and "the metric cannot see discovery" indistinguishable on this evidence. That's a statement about the evaluation, not a defence of any ranker.

A five-weight sweep confirms it isn't the mixture: 2 of 25 cells are positive at all, both at the weight nearest the benchmark, and neither survives the decision rule.

## The sweep's first version read signs instead of testing them

It saw those two positive cells — **+0.03 and +0.06 hits out of 20, with no interval attached** — and wrote that they were "the ones worth pursuing." Every positive cell now goes through the same bootstrap as the headline, the verdict is conditional on that outcome, and a guard fails if the retracted framing returns.

The prose also promised a weight sweep before one existed — the defect this repo keeps finding in its own documents. It's computed now rather than referred to.

614 passed, 4 skipped.


ELares:

PR #679, opened 2026-08-06
carry the discovery correction into MISSION.md, the last place it was asserted

ELares:
`MISSION.md` states the literature-based-discovery result and ended on the same summary the findings page carried: **"A good generator, a bad ranker."** The first half stands. The second doesn't follow, and the two measurements showing why are now in the repo.

It also already said *"the methods line up by how hard each corrects for degree: the harder the correction, the worse it does"* — an observation drawn from what each **formula does**, not from what each method **selects**. That's now measured: ordering by realised hub-selection reproduces the precision leaderboard exactly (rank correlation **1.00** over seven methods), and blending any seed-specific signal into a degree-only prior adds nothing measurable at any weight tested.

So the sentence was right, and is now a quantity.

## Why here

A mission statement's wording outlives the analysis it came from and travels further than it. The correction belongs where the claim is made, not only in the analysis that produced it.

## Guarded conditionally

Same shape as the findings page: if a method ever shows headroom, the original summary becomes defensible again and the guard steps aside — rather than freezing a conclusion the data could overturn.

The guard also **fails loudly if the anchor text is reworded**, instead of returning silently. That silent-return hole was found in this file's sibling guard yesterday, and it's not worth reintroducing.

616 passed, 4 skipped.


ELares:

Issue #531, opened 2026-06-26
ferroptosis-python `sim_batch` mean_lp/gsh/gpx4 are not bit-reproducible (order-dependent rayon float sum) despite the "seed for reproducibility" docstring


ELares:
`simulations/ferroptosis-python/src/lib.rs` (~lines 275-293) sums per-cell `lp/gsh/gpx4` floats via rayon `fold`/`reduce`, whose combination order depends on thread scheduling. Floating-point addition isn't associative, so `mean_lp/mean_gsh/mean_gpx4` are **not bitwise reproducible across runs** even with a fixed seed — contradicting the public `seed=...` "for reproducibility" contract. `n_dead`/`death_rate` are integer-summed and safe.

**Fix:** mirror `ferroptosis-core/src/stats.rs:74-76` — collect per-cell results then sum in a fixed (index) order. No production figure consumes these means today, so it's low-impact but a stated-guarantee violation.

---
*Filed from a 2026-06-26 outside-in contribution review (7-lens adversarial fan-out of the repo). Evidence verified against the current tree. Part of the post-#511 hardening backlog.*

PR #682, opened 2026-08-11
the per-cell buffer, not the arithmetic, was the ceiling on n

ELares:
`run_condition` collected every per-cell outcome into a `Vec` so the means could be summed in a fixed index order — float addition isn't associative, so a rayon fold made them depend on thread scheduling (#531). That bought determinism at **32 bytes a cell**: 32 MB at the n=1e6 the figures use, **32 GB at 1e9, 3.2 TB at 1e11**.

The buffer was the limit, not the compute. Measured here: a cell costs **5.3 µs of CPU**, so 1e11 cells is 147 CPU-hours — an overnight run — against a terabyte allocation no single machine will do.

## Chunked deterministic fold

| | before | after |
|---|---|---|
| peak RSS (16M sims) | 38.1 MB | **2.6 MB** |
| wall | 6.66 s | 7.33 s |
| CPU | 84.7 s | 84.2 s |
| partials at n=1e11 | 3.2 TB | **~49 MB** |

**The chunk size is a function of `n` alone, deliberately.** Tying it to thread count would make the summation order — and the reported means — differ between machines, which is exactly what the collect existed to prevent. A first attempt used a fixed 65536, gave only 16 blocks at n=1e6, and cost 2.3× in wall time on 14 cores.

## Counts bit-identical; means now *correct*

`n_dead` and `death_rate` match exactly across all 16 conditions. The means change in the last four digits because the summation order changed — and the new order isn't merely different, it's right: against a 50-digit exact sum of a million values at the same magnitude, the chunked result has **zero error** where naive sequential is off by 1.3e-8.

Naive error grows like `n·ε` — ~2e-5 relative at 1e11, which had started to eat the mean it reports. One level of pairwise summation makes it grow like `(chunk + n/chunk)·ε`.

Nothing asserts those digits, `sim-tme-3d`'s byte-identity gate doesn't use this path, and both suites pass (**484 Rust, 626 Python**).


ELares:

**Verified, not assumed:**

* **The byte-identity gate cannot see this.** `sim-tme-3d` never calls `run_condition` (`rg` over the whole crate returns nothing). The only consumers are `sim-original` and the new `sim-scale`; `sim-combo-mech` defines its own local function of the same name, which is unrelated.
* **Determinism holds.** `chunk` is a function of `n` alone, never of thread count or core count, so the summation order is identical on every machine — the property the old `collect()` existed to protect. `into_par_iter().collect()` over a `Range` is order-preserving, so the partials are folded in ascending block order.
* **Per-cell streams are untouched.** The seeding is still `i*2` / `i*2+1` on the global index, so every cell gets the same two RNG streams it got before. This is what makes the counts bit-identical rather than merely close.
* **Live evidence on this build:** `sim-scale --verify` reproduces all six committed Figure 7 conditions at n=1e6 and passed at the start of the running sweep.
* **`n = 0` is unchanged.** Old path: empty vec, `0/0` = NaN. New path: `n_chunks = 0`, empty partials, `0/0` = NaN. Same, including the NaN.

**One consequence worth stating, which is not a blocker:**

`simulations/simulation_results.json` is produced by `sim-original`, which does use this path, so its `mean_*` fields are now stale in roughly the last four digits. Nothing depends on them — no test reads the file, no figure generator consumes those keys, and its `MANIFEST.sha256` entry is consistent because the file itself is unchanged. The counts and death rates in it are bit-identical. It will refresh on its next regeneration; regenerating it here would only add noise to a performance change.

The direction of the numerical difference is worth recording: the chunked fold is not merely *different*, it is **more accurate**. Checked against a 50-digit `Decimal` reference, the chunked sum has zero error on the test case where the old naive sequential sum was off by 1.26e-8. The last digits moved toward the true value, not away from it.

PR #683, opened 2026-08-11
the per-cell draw is unbounded Gaussian, not bounded uniform — and that decides whether 0% means anything

ELares:
Found during pre-flight for a large rare-event run, which is the only reason it mattered enough to check.

§5.3 said each cell's parameters are *"drawn from a ±20% uniform random variation around the population mean"*. `gen_cell` draws each from `Normal::new(mean, sd)`, per-parameter CV **10%–40%**, floored at a physiological minimum.

| | manuscript | code |
|---|---|---|
| family | uniform | **Gaussian** |
| width | ±20% | 95% spans **±20% to ±78%** |
| **support** | **bounded** | **unbounded** |

## Why the third row is the one that matters

Under a **bounded** draw, a cell outside the box is impossible — so a death rate reported as 0% would be *exactly* zero, and no sample size could ever find an event.

Under the **Gaussian** draw, every threshold keeps a positive-probability tail — so a reported 0% is an **upper bound set by n**.

Several conditions in Figure 7 report exactly 0%. Which of those two things they mean is the difference between a measurement and a detection limit.

It also decides whether a large-n rare-event experiment is worth running. Taking the manuscript at its word, it isn't — the events would be *impossible*, not rare. Taking the code at its word, it is. **The code is right.**

## The guard caught my own first correction

I derived the CV range from the Glycolytic block alone and wrote "12% to 25%". Across all phenotypes it's **10% to 40%** — which makes the original "±20%" wronger still, since the widest parameter has a 95% span of ±78%. The test compares the stated range against *every* `norm(rng, mean, sd)` in `gen_cell`, so a phenotype-local reading fails.

Guarded four ways: the code still draws from normals (the premise), the prose doesn't call it uniform, the stated CV range matches every draw, and the operational consequence — a reported zero is an upper bound — is stated rather than left for the reader to derive.

630 passed, 4 skipped.


ELares:

**The substantive claim checks out.** `gen_cell` draws each parameter through `norm(rng, mean, sd)`, i.e. `Normal::new`, so the support is unbounded. The manuscript's "±20% uniform" was wrong in all three ways the paragraph now names, and the third one is the one that matters: under a bounded draw a reported 0% death rate would be *exactly* zero and no sample size could ever find an event. The entire rare-event arm depends on which of those two is true.

**One defect found and folded in.** The corrected CV range reached the prose and the assertion but not the docstring three lines above them, which still read "12% to 25%" with 95% spans of "±24% to ±49%". Recomputed from `cell.rs`: the distinct CVs are 10, 11, 12, 13, 17, 20, 21, 24, 25, 27, 30, 40 — so **10% to 40%**, spans ±20% to ±78%. The stale numbers were the Glycolytic block alone, which is the first phenotype in the file and so the first one read; the widest draw is nearly twice what that narrow reading suggested.

Worth noting *why* the assertion stayed correct while the sentence beside it rotted: `test_the_stated_cv_range_matches_the_code` computes the range from `cell.rs` rather than restating it. A guard that derives its expectation cannot drift with the thing it guards. The docstring, being prose, had to be fixed by hand — and now carries the correction and its cause, so the same misreading does not get made twice.

Manifest regenerated for the edited file (one line).

Issue #500, opened 2026-06-25
[calibration] Joint multi-inducer fit of the single-cell switch (a posterior, not more knobs)


ELares:
> From the 2026 fresh-eyes honesty review.

## Context
The one real calibration anchors a SINGLE inducer's in-vitro switch (#330). The headline in-vivo/spatial layers stay prior-predictive, and ABC shows the in-vivo priors and in-vitro data are disjoint. The fix is a joint, multi-inducer posterior so the core switch carries intervals derived from data, not point guesses.

## What to do
- [ ] Jointly fit `lp_propagation`, `lp_rate`, `gpx4_rate`, `gsh_scav_efficiency` (and the cystine axis from the System Xc- issue) to a multi-inducer panel: RSL3 / ML162 / ML210 / erastin across DepMap/CTRPv2 lines already fetched (`analysis/calibration/ctrpv2_ferroptosis_curves.csv`).
- [ ] Fold in erastin (System Xc-), which is currently used only as a cross-mechanism contrast.
- [ ] Produce a POSTERIOR with credible intervals (extend `scripts/abc_posterior.py`), plus posterior-predictive checks on held-out lines/inducers.
- [ ] State explicitly what remains unconstrained.
- [ ] Propagate the intervals into every manuscript switch number.

## Acceptance criteria
A committed posterior + posterior-predictive plots + a calibration report; the manuscript switch magnitudes appear with intervals.

## Effort / who
I can implement (needs the compiled extension / `sim_batch`).

Standards: report a posterior, not another tuned knob; show held-out generalization, not just training fit.

PR #684, opened 2026-08-11
a harness for large-n rare-event sweeps, and the reason a 0% death rate needed one

ELares:
## What this is for

Several conditions in Figure 7 report a death rate of exactly **0%** at the manuscript's standard n = 1,000,000. That is not a measurement of zero. `gen_cell` draws each cell's parameters from **normal** distributions, which have unbounded support, so every threshold keeps a positive-probability tail. A reported 0% is an upper bound set by the sample size — about 3e-6 at a million cells — and nothing more.

This adds what is needed to resolve those bounds, and the discipline to read the result correctly.

## The distinction the whole thing turns on

With zero events the 95% upper bound is the rule of three, `3/n`. It falls by exactly one decade per decade of n: a straight line of slope -1 on log-log axes. **That curve is the sample size, not the biology.** Plotting it without saying so would manufacture the appearance of a measured decline, which is the single easiest way for this exercise to produce a wrong finding.

So every condition gets classified:

| label | meaning |
|---|---|
| `resolution-limited` | zero at every n; the bound tracks 3/n exactly. We learned a **limit**, not a value. |
| `emergent` | zero at small n, non-zero at large n. Locates the order of magnitude where the tail begins. |
| `resolved` | events throughout; the rate is measured and large n buys precision. |

`emergent` is the outcome worth the money. No amount of running at n = 1e6 can produce it.

## The four pieces

**`simulations/sim-scale`** — the harness. `--verify` reruns the six committed Figure 7 conditions at n = 1e6 and compares against what the figure reports; a large run refuses to start unless that passes, because an engine that cannot reproduce a known result should not be trusted with an expensive one. `--params` takes the same JSON override map as `FERRO_PARAM_OVERRIDES` (#331), so a run can be driven from the CTRPv2 in-vitro posterior instead of the defaults — and every output row records which, since the two parameterisations are provably disjoint (#332, #500).

**`scripts/rare_event_sweep.py`** — the driver. Appends each result the moment it lands and never recomputes a completed cell, so an interrupted run loses at most one condition. On a metered instance with an auto-terminate timer that is not optional.

**`scripts/rare_event_analysis.py`** — the reading. Draws the 3/n floor as an explicit reference line, plots zero-event runs as downward carets *at their upper bound* rather than as points at zero (0 has no place on a log axis, and drawing it as a small number would be a lie), and writes the classification plus what it does not show.

**`scripts/aws/launch_scale_run.sh`** — the box for the 1e11 tier. Four independent kill switches, because the failure that costs money is the one where the job dies and the instance does not:

1. `--instance-initiated-shutdown-behavior terminate` (a *stopped* instance still bills for its EBS)
2. a dead-man `shutdown -h +N` scheduled as the **first** action of user-data, before the build — so a failed build still dies on schedule
3. an explicit shutdown when the job ends, success or failure
4. a CloudWatch idle-CPU alarm, for when the job exits without running its trap

The root volume is deliberately `DeleteOnTermination=false`. This account has no S3, so there is nowhere to stream results; without that flag an auto-terminate would destroy exactly the thing the run existed to produce. Nothing account-specific is in the file — region, key, subnet and AMI all resolve at run time.

## Verification

The exact Poisson interval is hand-rolled on stdlib (a gamma quantile by bisection), and it is the number a reader takes entirely on trust. At 42% death a normal approximation would be fine; at **one event in a million** the interval spans two orders of magnitude, the approximation is meaningless, and a bug would produce a plausible-looking number with nothing to eyeball it against.

So it is checked against the textbook chi-square definition rather than against itself:

| k | expected | source |
|---:|---|---|
| 0 | upper 3.689e-6 | `chi2(0.975, 2)/2` |
| 1 | [2.53e-8, 5.57e-6] | `chi2(0.025,2)/2`, `chi2(0.975,4)/2` |
| 451 | [4.10e-4, 4.95e-4] | same, and near-symmetric as a large count should be |

The one-decade-per-decade property the figure rests on is asserted directly, and the classifier is tested at its boundary — a *single* event at the top scale must read `emergent`, not get rounded back into `resolution-limited`.

## First result, already in hand

**Glycolytic + RSL3 at n = 1e9: zero deaths in a billion cells** (628 s). The bound on the manuscript's own selectivity constraint tightens from 3.0e-6 to **3.69e-9** — in a 1-gram lesion, fewer than about 4 glycolytic cells would be expected to die from RSL3. Still resolution-limited, still tracking 3/n, so this is a statement about how far down we can now see.

Results, figure, and the findings write-up land in a follow-up once the sweep completes; this PR is the machinery.

## What it does not claim

1e11 **independent** cells is not a 100-gram tumor. These cells never interact — no diffusion, no crowding, no vasculature, no clonal competition. The defensible phrase, used throughout, is *a population the size of an advanced burden, sampled as independent cells*. The spatial model is the one with interaction, and it does not reach these counts.


ELares:
## Review before merge

A 35-agent adversarial review over five lenses (Rust, driver, cloud launcher, scientific framing, repo conventions) raised **30 findings; 23 were refuted, 7 survived** independent refutation. All 7 are fixed in this PR.

The important one corrected my own diagnosis of a failure I had already reported. **The sweep was running a binary compiled before #682** — built 22:48, against a `stats.rs` last written 23:15 and a #682 commit at 23:11. It still had the per-cell buffer: **32.30 bytes/cell measured, against 0.29 on the current build.** At n=1e10 that is ~320 GB on a 48 GB machine, so the 27-minute death with empty stderr was an OOM SIGKILL, not the external kill I had concluded. My "RSS is 1 MB" reading came from `pgrep -f` matching a wrapper process rather than the simulation; the live process showed VSZ 713 GB.

What makes it worth writing down is why nothing caught it. `--verify` runs at n=1e6, where even the old path costs ~32 MB, so a memory regression is structurally invisible to it — and the arithmetic it checks was never wrong, because per-cell RNG seeding is by global index in both paths. **A stale binary passes its own self-check perfectly.** The gate proves the engine reproduces the figure; it says nothing about *which* engine. So the driver now builds every time rather than checking more cleverly, and refuses outright if cargo is absent.

Scope of the damage, stated precisely: `n_dead`, `death_rate` and the Poisson intervals were unaffected (identical seeding; #682 verified counts bit-identical), and the rebuilt binary reproduces 0/0/1 at 1e6 and 0/0/760 at 1e9 exactly. The `mean_*` and timing fields were the old path's, and the claim that the chunked fold had been demonstrated at n≥1e9 was not supported by any committed artifact. The sweep has been fully regenerated on the correct build — **and Glycolytic+RSL3 at n=1e10 has now completed, 0 deaths, bound 3.69e-10, 6542 s**, which is the first real demonstration of the fold at that scale.

The other six:

| finding | fix |
|---|---|
| resume key ignored `--params`, so a perturbed sweep was silently skipped, or worse appended beside default rows and plotted as one curve across two provably disjoint parameterisations (#332, #500) | parameter set is part of the key; mixed provenance in one file is now reported as an error |
| the findings document asserted "in-vivo defaults, zero calibration targets" as hardcoded prose, while `--params` exists precisely so a run can use the in-vitro posterior | derived from the per-row `params_source` the harness already stamps |
| derived artifacts had no regeneration path, so the next tier would silently desync them | the sweep regenerates them on completion |
| `--cells 0` parsed as an exact usize and skipped the guard, emitting a record of nulls with exit 0 | both parse branches guarded; boundary tested at `0`, `0.0`, `-1`, `0e9` |
| the launcher's `REPO_REF=main` default contains none of this code until it lands | preflights that the ref actually contains `sim-scale`, with the command it used |
| binary count 11 → 12, stale in three places including the `CLAUDE.md` line this PR was already editing for a different counter | corrected, and now derived by a test from the crates that exist |

Separately, an accidental local execution of the launcher's user-data (I used `source` where `bash -n` was meant) proved the self-check gate printed "aborting before the expensive run" and then **did not abort** — no `exit` after `shutdown`, which is asynchronous even as root, so the ten-hour run would have started inside the halt window on an engine that had just failed to reproduce the figure. Fixed, with a check that no `shutdown` meant to stop the script lacks a following `exit`.

645 passed, 4 skipped. fmt clean. All checks green.

PR #686, opened 2026-08-11
cargo was never on PATH, and the instance had no way to tell us

ELares:
Two 1e11 launches died within 80 seconds. The self-check gate worked correctly both times — it refused to spend ten hours on a tree that had not compiled — but the instance shut itself down with `/var/log` on a detached volume, writes to `/dev/console` never surfaced in `get-console-output`, this account has no S3, and there was no SSH key to hand. Three plausible causes were checked and **all were wrong**: the subnet does assign public IPs, the route table has an internet gateway, and the default security group allows all egress.

**So the first fix is diagnosability.** Everything written to `/dev/kmsg` lands in console output, because it is the kernel's own buffer. Each phase now announces itself and its exit status. The next launch said:

```
FERRO: BEGIN dnf install    END rc=0
FERRO: BEGIN rustup         END rc=0
FERRO: cargo version: ...: cargo: command not found
FERRO: BEGIN git clone      END rc=0
FERRO: BEGIN cargo build    END rc=127
```

which is the whole answer. **cloud-init runs user-data as root but without a sensible `HOME`** — unset or `/` — and rustup's env file says `export PATH="$HOME/.cargo/bin:$PATH"`. Sourcing it under the wrong HOME puts a nonexistent directory on PATH and *succeeds while doing it*, so nothing reports an error; cargo is simply absent three commands later.

Fixes: `HOME`/`CARGO_HOME`/`RUSTUP_HOME` set explicitly, the absolute path on PATH so the env file isn't depended on at all, and a `command -v cargo` check that aborts immediately rather than discovering it at build time. `tmux` is dropped from the `dnf` line for the same class of reason — dnf is transactional, so one unavailable package installs *nothing*, including git.

Verified by diffing the generated user-data against the user-data of the instance now running: identical on every executable line.

That instance is up — cargo 1.97.1, **self-check passed against the committed Figure 7**, sweep started, 96 vCPUs at 99.9%.


ELares:

Verified rather than assumed: the generated user-data was diffed against the user-data of the instance **currently running the 1e11 sweep**, recovered with `describe-instance-attribute --attribute userData`. Identical on every executable line (the only differences are a line continuation and one message's wording). So this template is known to produce a boot sequence that reaches a passing self-check and a running sweep, rather than one that merely looks right.

The diagnostic value is the part worth keeping. Three plausible causes were checked before the markers existed — public IP assignment, the route table's internet gateway, the default security group's egress — and **all three were wrong**. The markers answered it on the first launch after they were added. A one-shot instance that shuts itself down and takes its filesystem with it needs a channel that survives it; `/dev/kmsg` is that channel because it is the kernel's own buffer, and `/dev/console` measurably is not.

Live state at time of merge: 1.3 h elapsed of an 18 h dead-man, 96 vCPUs at 99.9%, on track against a ~7 h estimate.

PR #704, opened 2026-08-13
dependency batch four: pyo3 0.29.2 and streamlit >=1.61.1

ELares:
Two dependabot PRs (#700, #701), each failing on `manifest` **alone** and neither able to fix it: the gate hashes every in-scope tracked file, so any lockfile bump stales `MANIFEST.sha256` and a bot cannot regenerate it. Batching is the documented answer — take the bumps, regenerate once, run the gate.

**Taking only the file each PR intends to change, not merging their branches.** Both were cut before the ABC tolerance work landed, so their view of the tree would revert it: `git diff main..dep-pyo3` shows 17 files and 500 deletions, of which exactly one line is the actual bump.

## pyo3 is exercised, not assumed

It's used by the *bindings* crate, not by the simulation binaries — so building those proves nothing, and the first build here finished in 0.51s having rebuilt precisely nothing. Rebuilt `ferroptosis-python` with maturin and checked:

- `sim_batch` still returns (Persister/RSL3 at n=1000 → 0.4000)
- bindings suite: 12 passed
- full Rust workspace: 111 passed
- `cargo fmt --all --check`: clean

streamlit is dashboard-only and absent from `requirements-lock.txt`, so it cannot affect the offline contract.

688 Python tests passed, 2 skipped.


ELares:
**Closing the gap I flagged on this PR.** The convention when pyo3 moves is to re-check the sim-tme-3d byte-identity contract rather than argue that it is only the binding layer. I verified the bindings and the Rust suite here but did not run the production matrix, so:

```
$ rustc --version
rustc 1.96.0 (ac68faa20 2026-05-25)          # the pinned channel the hash requires
$ cargo run --release -p sim-tme-3d          # full 24-condition 60^3 x 180 matrix
$ sha256sum -c sim-tme-3d/expected_summary.sha256
output/tme-3d/summary.json: OK
```

So pyo3 0.29.2 does not move the production output. That is the expected answer — `expected_summary.sha256` names the toolchain and `rand`/`rand_distr` as the things that can move it, and sim-tme-3d does not link pyo3 at all — but the file's own instruction is to measure rather than reason, and the reasoning is what would have been wrong if pyo3 had reached further than I thought.

